PR #1, opened 2019-10-24
Bugfix for MacOS installs from pip

mfherbst:
There was an issue properly replacing the linux binaries shipped with with the downloaded MacOS binaries. This should be fixed by this PR. I have made a test upload with the proposed fix at <https://get.adc-connect.org/adcc-0.13.1.tar.gz>.

@maxscheurer Could you test that?

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -63,16 +63,17 @@ macOS 10.13 (High Sierra) and 10.14 (Mojave)
 .. note::
    Supported from adcc 0.13.2.
 
-.. note::
-   TODO This section needs more details!
-
+0. **Homebrew:**
+   Support for macOS currently requires the `Homebrew <https://brew.sh>`_ package manager
+   and a recent version of ``gcc`` (e.g. ``gcc@9``). Hopefully, we will support ``clang`` in the future.
+   
 1. **adcc:**
    Install from `PyPi <https://pypi.org>`_, using ``pip``:
 
    .. code-block:: shell
 
       pip install pybind11     # Install pybind11 first to suppress some error messages
-      pip install adcc
+      CXX=g++-9 CC=gcc-9 pip install adcc   # Install adcc using the correct compiler for Python bindings
 
 .. _install-hostprogram:
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -60,9 +60,8 @@ macOS 10.13 (High Sierra) and 10.14 (Mojave)
    only covers High Sierra and Mojave.
    We would love to hear your feedback in case things fail.
 
-   In adcc 0.13.1 there is a known bug for the pip installation
-   on macOS, causing it to produce a corrupt installation,
-   for a workaround, see https://github.com/adc-connect/adcc/pull/1.
+.. note::
+   Supported from adcc 0.13.2.
 
 .. note::
    TODO This section needs more details!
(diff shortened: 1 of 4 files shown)


maxscheurer:
The problem still exists when installing via
```
pip3 install --no-cache-dir https://get.adc-connect.org/adcc-0.13.1.tar.gz
```

I have no idea what's going wrong here...

mfherbst:
But you can checkout the github repository and successfully build and test adcc, right? So it's only an issue with the pip installation?

maxscheurer:
Let me check this quickly...

maxscheurer:
Okay I found the bug, but I don't know how to fix it...

The difference between the two installations is that in the `adcc/lib` directory installed by pip, I find `libadccore.so`, whereas in the case of my locally, working installation `libadccore.0.13.1.dylib` and `libadccore.dylib`.

mfherbst:
Ok ... that's the error! Because `libadccore.so` is a linux elf binary. That really surprises me as it should be removed with the fix I introduced here. Also I just checked ... the fix is definitely deployed in the tarball I have linked above.

Could you double-check that you do not use some weird cache from pip? I'll try to cook up a way to enable a more "verbose" version of the installation, such that we can check what happens on your machine.

maxscheurer:
I did triple-check the installation in three new virtualenvs.
What surprises me is that there is no `dylib` at all...

mfherbst:
Oh ... that's weird, too ... and definitely not the way it's supposed to be.

mfherbst:
I think I'm starting to understand what the problem is. When pip installs a package, then it does the download and the compilation of the extension in a temporary directory first and only later copies the resulting files over to the site-packages folder. It could be, that the download and replacement is only done inside this temporary folder, but (apart from the extension `so` file) the copying is done from the unchanged tarball anew. That way the installation does as it should, it just does not operate on the "right" folder.

This might be different for installations from a directory. We'll have to investigate this further.

maxscheurer:
Yes, because the site-packages folder is created at the last instance of the installation...
So annoying...

maxscheurer:
Probably this information helps:

In the temporary directory that `pip` builds `adcc`, the `*.so` files live in `adcc/lib` and `build`, whereas the `dylib` files live in `extension/adccore/lib`.

See the following output run in the temporary `pip` build directory for clarification:

```
find . -name "*.so"
./adcc/lib/libadccore.so
./build/lib.macosx-10.14-x86_64-3.7/adcc/lib/libadccore.so
```

vs.

```
find . -name "*.dylib"
./extension/adccore/lib/libadccore.dylib
./extension/adccore/lib/libadccore.0.13.1.dylib
```

mfherbst:
Yes, that's it. This does not matter for source builds, because there `adcc/lib` is symlinked to `extension/adccore/lib`, so it's the same thing.

I'll think about something for that.

mfherbst:
Ok, I now made the paths relative to the top of the repository and use the adcc/lib path explicitly in the globbing expressions. This makes the tarballs incompatible with 0.13.1, so we will have to version-bump.

I'll build the new adccore tarballs using our Travis setup now and then we can first verify that everthing is working on Mac.

mfherbst:
@maxscheurer Also, could you extend the documentation on the MacOS installation if neccessary? If not, just remove the TODO note :smile:.

PR #9, opened 2019-10-26
Canonically declaring dependencies in setup.py

maxscheurer:
Dependencies are now declared without parentheses around the version as specified [here](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies).

This will keep us out of trouble when going for conda packaging.


mfherbst:
By the way ... for this you can use `[no ci]` in the commit message to suppress the Travis run. IMHO this is not needed for such minors if you have done a bit of checking locally :smile:.

PR #12, opened 2019-10-26
Add zenodo badges

mfherbst:


mfherbst:
Turns out you cannot "[skip ci]" all commits of a PR unfortunately ... otherwise you cannot merge it.

PR #15, opened 2019-10-27
[WIP] Clang support

maxscheurer:
This PR will add support for `clang`, which will be beneficial for `conda` deployment.

## ToDo:
- [x] Apple Clang (10.0.1)
- [x] clang 4.0.1 (conda)
- [x] clang 9.0.0 (conda-forge)
- [x] test clang compilers on Linux (clang 9 from conda-forge)
- [x] test gcc compilers on Linux (gcc 7 from conda-forge)
- [x] revert formatting changes (@mfherbst's clang-format has been run)
- [x] deploy corresponding `adccore` (**done**)
- [x] let Travis run with new `adccore` here (**Change MacOS to use clang**)
- [x] Adapt documentation for MacOS

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -61,22 +61,20 @@ macOS 10.13 (High Sierra) and 10.14 (Mojave)
    macOS support is still experimental and so far
    only covers High Sierra and Mojave.
    We would love to hear your feedback in case things fail.
+   We are currently working on deploying adcc to `Anaconda <https://anaconda.org>`_.
 
 .. note::
-   Supported from adcc 0.13.2.
-
-0. **Homebrew:**
-   Support for macOS currently requires the `Homebrew <https://brew.sh>`_ package manager
-   and a recent version of ``gcc`` (e.g. ``brew install gcc@9``).
-   Hopefully, we will support ``clang`` in the future.
-   
+   The installation on macOS requires a ``clang`` compiler.
+   Make sure to have XCode and the command line tools installed.
+   ``clang`` is supported from adcc 0.13.2.
+ 
 1. **adcc:**
    Install from `PyPi <https://pypi.org>`_, using ``pip``:
 
    .. code-block:: shell
 
       pip install pybind11     # Install pybind11 first to suppress some error messages
-      CXX=g++-9 CC=gcc-9 pip install adcc   # Install adcc using the correct compiler for Python bindings
+      pip install adcc   # Install adcc
 
 .. _install-hostprogram:
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -63,10 +63,9 @@ macOS 10.13 (High Sierra) and 10.14 (Mojave)
    We would love to hear your feedback in case things fail.
    We are currently working on deploying adcc to `Anaconda <https://anaconda.org>`_.
 
-.. note::
-   The installation on macOS requires a ``clang`` compiler.
-   Make sure to have XCode and the command line tools installed.
-   ``clang`` is supported from adcc 0.13.2.
+
+The installation on macOS requires a ``clang`` compiler.
+Make sure to have XCode and the command line tools installed.
  
 1. **adcc:**
    Install from `PyPi <https://pypi.org>`_, using ``pip``:
--- a/extension/export_HartreeFockProvider.cc
+++ b/extension/export_HartreeFockProvider.cc
@@ -67,7 +67,7 @@ class HartreeFockProvider : public HartreeFockSolution_i {
                                ").");
     }
     py::memoryview memview(py::buffer_info(buffer, n_orbs));
-    fill_occupation_f(py::array({n_orbs}, buffer, memview));
+    fill_occupation_f(py::array({{n_orbs}}, buffer, memview));
   }
 
   void orben_f(scalar_type* buffer, size_t size) const override {
@@ -79,7 +79,7 @@ class HartreeFockProvider : public HartreeFockSolution_i {
                                ").");
     }
     py::memoryview memview(py::buffer_info(buffer, n_orbs));
-    fill_orben_f(py::array({n_orbs}, buffer, memview));
+    fill_orben_f(py::array({{n_orbs}}, buffer, memview));
   }
 
   void orbcoeff_fb(scalar_type* buffer, size_t size) const override {
@@ -303,10 +303,10 @@ class PyHartreeFockProvider : public HartreeFockProvider {
   void fill_fock_ff(py::tuple slices, py::array out) const override {
     PYBIND11_OVERLOAD_PURE(void, HartreeFockProvider, fill_fock_ff, slices, out);
   }
-  void fill_eri_ffff(py::tuple slices, py::array out) const {
+  void fill_eri_ffff(py::tuple slices, py::array out) const override {
     PYBIND11_OVERLOAD_PURE(void, HartreeFockProvider, fill_eri_ffff, slices, out);
   }
-  void fill_eri_phys_asym_ffff(py::tuple slices, py::array out) const {
+  void fill_eri_phys_asym_ffff(py::tuple slices, py::array out) const override {
     PYBIND11_OVERLOAD_PURE(void, HartreeFockProvider, fill_eri_phys_asym_ffff, slices,
                            out);
   }
@@ -381,11 +381,13 @@ void export_HartreeFockProvider(py::module& m) {
         //      and maybe should be removed for this reason
         .def_property_readonly("n_alpha",
                                [](const adcc::HartreeFockSolution_i& self) {
-                                 return count_electrons(self, /* count_beta = */ false);
+                                 return count_electrons(self,
+                                                        /* count_beta = */ false);
                                })
         .def_property_readonly("n_beta",
                                [](const adcc::HartreeFockSolution_i& self) {
-                                 return count_electrons(self, /* count_beta = */ true);
+                                 return count_electrons(self,
+                                                        /* count_beta = */ true);
                                })
         .def_property_readonly("conv_tol", &HartreeFockSolution_i::conv_tol)
         .def_property_readonly("restricted", &HartreeFockSolution_i::restricted)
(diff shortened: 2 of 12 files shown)


maxscheurer:
## Construction sites


- clang 9.0.0 currently complains about `index<N>` used in `TensorImpl.cc`
```
/tmp/adcc_clang/adccore/src/adcc/TensorImpl.cc:130:11: error: reference to 'index' is ambiguous
std::pair<index<N>, index<N>> assert_convert_tensor_index(
          ^
/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/strings.h:73:7: note: candidate found by name lookup is 'index'
char    *index(const char *, int) __POSIX_C_DEPRECATED(200112L);
         ^
/tmp/adcc_clang/adccore/external/libtensor/libtensor/core/index.h:111:7: note: candidate found by name lookup is 'libtensor::index'
class index : public sequence<N, size_t> {
```

- For better compatibility with `conda`, I tried to use a `conda`-provided MKL as BLAS library, but the `FindMKL` script provided by `libtensor` fails to properly recognize it.

extension/export_Tensor.cc
```diff
@@ -122,7 +122,7 @@ static py::array Tensor_dot_list(const Tensor& self, py::list tensors) {
   py::array_t<scalar_type> ret(dots.size());
   std::copy(dots.begin(), dots.end(), ret.mutable_data());
 
-  return ret;
+  return std::move(ret);
```
mfherbst: Are you sure this is correct? That looks very weird to me and breaks a couple of patterns I am aware of.

extension/export_OneParticleOperator.cc
```diff
@@ -57,7 +57,7 @@ static py::array OneParticleOperator__to_ndarray(const OneParticleOperator& self
   // Get an empty array of the required shape and export the data into it.
   py::array_t<scalar_type> res(self.shape());
   self.export_to(res.mutable_data(), self.size());
-  return res;
+  return std::move(res);
```
mfherbst: Same here.

extension/export_HartreeFockProvider.cc
```diff
@@ -408,79 +410,110 @@ void export_HartreeFockProvider(py::module& m) {
...
+             "terms.")
         .def("get_spin_multiplicity", &HartreeFockProvider::get_spin_multiplicity,
-             "Returns the spin multiplicity of the HF ground state. A value of 0* (for "
+             "Returns the spin multiplicity of the HF ground state. A value of "
+             "0* (for "
              "unknown) should be supplied for unrestricted calculations.")
         .def("get_n_orbs_alpha", &HartreeFockProvider::get_n_orbs_alpha,
-             "Returns the number of HF *spin* orbitals of alpha spin. It is assumed the "
-             "same number of beta spin orbitals are used. This value is abbreviated by "
+             "Returns the number of HF *spin* orbitals of alpha spin. It is "
```
mfherbst: The line breaks here (and below) need to be cleaned up.

extension/export_HartreeFockProvider.cc
```diff
@@ -303,10 +303,10 @@ class PyHartreeFockProvider : public HartreeFockProvider {
   void fill_fock_ff(py::tuple slices, py::array out) const override {
     PYBIND11_OVERLOAD_PURE(void, HartreeFockProvider, fill_fock_ff, slices, out);
   }
-  void fill_eri_ffff(py::tuple slices, py::array out) const {
+  void fill_eri_ffff(py::tuple slices, py::array out) const override {
```
mfherbst: Nice catch!

mfherbst:
Regarding the index thing. Just rename it to `lt::index`. Usually I use the namespace alias `lt` as `libtensor` somewhere at the top of the file. If not you can add it if you want. Make sure it does not pollute the header files, which are needed for compiling the adcc extensions, however :smile:.

mfherbst:
I'll look at the MKL issue. I assume one needs to write a few lines of cmake for that. With the other BLAS-Issues it was the same.

extension/export_HartreeFockProvider.cc
maxscheurer: Yes... Do you have the appropriate `.clang-format` file for this?

extension/export_HartreeFockProvider.cc
maxscheurer: Not my catch though... Caught by Apple Clang 😄

maxscheurer:
> I'll look at the MKL issue. I assume one needs to write a few lines of cmake for that. With the other BLAS-Issues it was the same.

We can make another PR for this once this PR goes in 👍

extension/export_Tensor.cc
maxscheurer: Apple Clang tells me the following:
```
extension/export_Tensor.cc:125:10: error: local variable 'ret' will be copied despite being returned by name [-Werror,-Wreturn-std-move]
  return ret;
         ^~~
extension/export_Tensor.cc:125:10: note: call 'std::move' explicitly to avoid copying
  return ret;
         ^~~
         std::move(ret)
```

We can stick to the previous version (without `std::move`) and disable the warning.

extension/export_HartreeFockProvider.cc
mfherbst: :+1:

extension/export_HartreeFockProvider.cc
mfherbst: Ah ... true, it's not committed ... let me work on that.

mfherbst: No it is committed :smile: ... see the `extension` subfolder ... and I checked for me it does not make those breaks (using clang-format 8)

mfherbst:
Can you then open an issue somewhere and assign me to not loose track.

extension/export_Tensor.cc
mfherbst: Hmmm ... I think this might be a complicated matter and ignoring it is not the best solution. Understanding is better. 

See for example this [stackoverflow](https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-confusion-return-statement) thread. If I understand correctly, the move-constructor of `array_t` should have been automatically called, so question is why not?

maxscheurer: Oh boy... that sounds complicated.
Do we need to check whether `array_t` is move-constructible? Maybe you can have a [look](https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h)... I have basically no experience with such things...

extension/export_HartreeFockProvider.cc
maxscheurer: My `clang-format` (also version 8) seems to ignore the `.clang-format` file 😦 
I assume running
```
clang-format -style=file -i bla.cc
```
is correct?

mfherbst: Usually it loads the file from the folder automatically, so `-style=file` should not even be needed, but otherwise it's correct as far as I know.

extension/export_Tensor.cc
mfherbst: I found the issue, see https://reviews.llvm.org/D43322 ... the problem is the return type is different! Just replace `std::array` by `std::array<scalar_type>` and that's it.

maxscheurer: Done.

extension/export_OneParticleOperator.cc
maxscheurer: Done.

extension/export_HartreeFockProvider.cc
maxscheurer: I'm not able to fix this, unfortunately... probably _you_ can run `clang-format` before we merge this PR.

setup.py
```diff
@@ -87,6 +88,9 @@ def __init__(self):
         self.library_paths += ["/usr/lib", "/lib"]
 
         for conf in ["/etc/ld.so.conf"] + glob.glob("/etc/ld.so.conf.d/*.conf"):
+            if not os.path.isfile(conf):
+                warnings.warn("File {} is broken.".format(conf))
```
mfherbst: Can you be more specific in the warning. If I saw this during installation I had no clue what was going on.

maxscheurer: Yes.

maxscheurer:
Looking good so far. Works with clang on macOS and Linux.
However, we also need to deploy a new `adccore` version :frowning:

setup.py
mfherbst: Ah ... I see what's going on. Why not be more specific? `os.path.islink(filename) and not os.path.exists(filename)`?

maxscheurer: I guess `exists` is fine here, because the symlink needs to be broken in order for this not to work.

mfherbst: Ok :smile: Good point!

extension/export_HartreeFockProvider.cc
mfherbst: I don't mind doing the formatting here, but can you undo your formatting changes to these files first? Then I just have to run `clang-format -i` and commit.

mfherbst:
I have a couple of questions here, which are in line, with what is discussed in #16:
- Do you think about having one adccore for each compiler (I think this is a bad idea)
- Do you know whether there are some forward / backward compatibility issues one needs to think about for clang (like the one in `gcc`, where I ship the standard library in order to avoid it).

Other points:
- In my experience compiler supports gets lost if one does not test for it automatically. So I think before we "enable" this, we need to make sure we test. So I suggest one Travis job here (probably best to pick MacOS) should do clang. Similarly we should change the `adccore` deployment to run one job with clang and one with gcc.
- So that already suggests to go with `clang` for MacOS in adccore and stick to `gcc` for linux, I think.

extension/export_HartreeFockProvider.cc
maxscheurer: Reverted.

mfherbst:
Just ran clang-format and pushed the changes.

maxscheurer:
Alright, then we just need to wait for `adccore` deployment 😄 🚀

mfherbst:
Easy :smile: :fireworks:

mfherbst:
I just noted that we actually used again gcc to build the macos binaries for adccore. This was probably not planned, right?

mfherbst:
There was an issue building the adccore 10.13 clang version, I'm not quite sure what's going on ... can you help @maxscheurer ?

maxscheurer:
> I just noted that we actually used again gcc to build the macos binaries for adccore. This was probably not planned, right? I'm rebuilding now.

Oh sorry... I totally forgot this in the hurry...

> There was an issue building the adccore 10.13 clang version

Looks like the `brew`-provided `numpy` is broken...

docs/installation.rst
```diff
@@ -61,22 +61,20 @@ macOS 10.13 (High Sierra) and 10.14 (Mojave)
...
-
-0. **Homebrew:**
-   Support for macOS currently requires the `Homebrew <https://brew.sh>`_ package manager
-   and a recent version of ``gcc`` (e.g. ``brew install gcc@9``).
-   Hopefully, we will support ``clang`` in the future.
-   
+   The installation on macOS requires a ``clang`` compiler.
+   Make sure to have XCode and the command line tools installed.
+   ``clang`` is supported from adcc 0.13.2.
+
```
mfherbst: I wouldn't put this full thing into a note. In fact, I think the note can now be dropped.

mfherbst: No, not drop it ... just make it plain text. I think the `clang is supported from adcc 0.13.2` adds no info, because 0.13.1 was not really working anyway :smile:

maxscheurer: Done.

PR #20, opened 2019-10-29
[WIP] UHF support for PySCF and Psi4

maxscheurer:
This PR will add "real" support for unrestricted HF references in PySCF and Psi4.
Until now, UHF was only available for PySCF through the fallback `dict` mechanism, which is slow.

- [x] refactor `EriBuilder` to make UHF ERI import work
- [x] UHF examples working (both programs)
- [x] re-enable caching of ERIs
- [x] cleanup `EriBuilder` (print statements, unused code, commented code, etc.)
- [x] enable unit tests ~~(`TestBackendsImportReferenceData` are broken for CN)~~
- [x] fix tests
- [x] add crossref test for Psi4/PySCF
- [x] run all tests with crossref
- [x] Rebasing and squashing the commits sensibly.


maxscheurer:
I get some weird errors in `TestBackendsImportReferenceData` when enabling unrestricted systems (CN molecule).

- mismatch between `conv_tols` [here](https://github.com/adc-connect/adcc/blob/3004bf74ecf2a739d365ba6d12ebb78e9c09e8c4/adcc/test_ReferenceState_refdata.py#L63)
- if I disable the above `assert`, I get mismatching ERIs [here](https://github.com/adc-connect/adcc/blob/3004bf74ecf2a739d365ba6d12ebb78e9c09e8c4/adcc/test_ReferenceState_refdata.py#L93)

😦

maxscheurer:
I think figuring out the issue with the ERIs is going to take a while... Can we be sure that the ERIs
in unrestricted form are actually **unique** in the MO space for different implementations?
I have no gut feeling about this, but you probably do, @mfherbst...

The calculations seem to work, however, and all the other data (e.g., Fock matrix) are in agreement with the reference data.

mfherbst:
> Can we be sure that the ERIs in unrestricted form are actually unique in the MO space for different implementations?

I'm not completely sure what you mean here, but there are some subtleties with UHF. In the particular case of CN we are testing, I think the problem is that breaking the spin symmetry during the UHF-SCF also breaks the spatial symmetry. In other words the linear CN molecule is no longer rotationally symmetric about its rotation axis. The direction in which the symmetry is broken depends on the numerical details since any direction perpendicular to the linear axis would work. In other words ... it depends on the details of the backend. Since the reference data is done with pyscf , psi4 will almost definitely find a differently spin-broken SCF result. But even with pyscf the answer might be dependent on subtle things such as DIIS space size or tolerance ...

I think the best we can do is test the ERI generation differently. I'm working on that now.

maxscheurer:
> I'm not completely sure what you mean here, but there are some subtleties with UHF

Yes, what you explained is what I meant 😄 

> I think the best we can do is test the ERI generation differently. I'm working on that now.

Very good! Let me know about updates, then I can pick up the work on this PR again and finish it.

mfherbst:
Ok ... I added another test case for UHF, which has less symmetries. There the comparsion of the ERI values works in the problematic test. On my end things are working. Let's see what Travis has to say.

@maxscheurer Could you also run the long tests on your machine as well to confirm?

maxscheurer:
Works on my machine (full tests). Travis seems to be fine as well.
I'll look through the code again for some cleanup.
Thanks for helping out :+1:

maxscheurer:
From my point of view, this can go in (I'd be okay with a squash-merge).

Having RHF and UHF, we could go for some benchmarks in the [adcc-bench](https://github.com/adc-connect/adcc-bench) repository.... Would be nice to have for the website.

adcc/backends/eri_build_helper.py
```diff
@@ -216,36 +205,36 @@ def fill_slice(self, slices, out):
...
-        co = self.coeffs_occ_alpha
-        cv = self.coeffs_virt_alpha
         block = can_block
         asym_block = "".join([block[i] for i in [0, 3, 2, 1]])
-        both_blocks = "{}-{}-{}-{}".format(
-            block, asym_block, str(spin_symm.pref1), str(spin_symm.pref2)
-        )
+        both_blocks = f"{block}-{asym_block}-{spin_symm.pref1}-{spin_symm.pref2}"
+        if not self.restricted:
+            both_blocks += "-" + spin_symm.transposition
```
mfherbst: I made this a format-string ... since we require python 3.6 we can do this now ... helps a lot, as in here :smile:

adcc/backends/psi4.py
```diff
@@ -168,7 +169,7 @@ def import_scf(wfn):
             "calculation."
         )
 
-    if not isinstance(wfn, psi4.core.RHF):
+    if not isinstance(wfn, (psi4.core.RHF, psi4.core.UHF)):
         raise InvalidReference("Right now only restricted references (RHF) are "
                                "supported for Psi4.")
```
mfherbst: This is out of date ...

adcc/backends/test_backends_crossref.py
```diff
@@ -50,6 +50,18 @@ def template_adc2_h2o(self, basis):
             results[b] = adcc.adc2(scfres, n_singlets=5, conv_tol=1e-10)
         compare_adc_results(results, 5e-9)
 
+    def template_adc2_uhf_h2o(self, basis):
+        results = {}
+        # UHF not supported for VeloxChem
+        if "veloxchem" in backends:
+            backends.remove("veloxchem")
+            if not len(backends):
+                pytest.skip("Not enough backends that support UHF available.")
```
mfherbst: The indention level here is wrong, right?

adcc/backends/test_psi4.py
```diff
@@ -116,3 +116,18 @@ def template_rhf_h2o(self, basis):
         mints = psi4.core.MintsHelper(wfn.basisset())
         ao_dip = [np.array(comp) for comp in mints.ao_dipole()]
         operator_import_test(wfn, ao_dip)
+
+    def template_uhf_h2o(self, basis):
```
mfherbst: I think it makes more sense to use `ch2nh2` here now ... same for PySCF, don't you agree?

adcc/backends/test_psi4.py
```diff
@@ -116,3 +116,18 @@ def template_rhf_h2o(self, basis):
...
+        wfn = adcc.backends.run_hf(
+            "psi4", geometry.xyz["h2o"], basis, multiplicity=3
+        )
```
mfherbst: I don't think the line break is needed.

adcc/backends/test_pyscf.py
```diff
@@ -159,7 +152,22 @@ def template_rhf_h2o(self, basis):
         ao_dip = mf.mol.intor_symmetric('int1e_r', comp=3)
         operator_import_test(mf, list(ao_dip))
 
+    def template_uhf_h2o(self, basis):
```
mfherbst: Also use `ch2nh2` here?

adcc/backends/test_pyscf.py
```diff
@@ -159,7 +152,22 @@ def template_rhf_h2o(self, basis):
...
+        # Test ERI
+        eri_asymm_construction_test(mf)
+        eri_asymm_construction_test(mf, core_orbitals=1)
+
+        # Test dipole
+        ao_dip = mf.mol.intor_symmetric('int1e_r', comp=3)
+        operator_import_test(mf, list(ao_dip))
+
     def test_h2o_sto3g_core_hole(self):
+        pytest.xfail("Systems with a core-hole are currently not supported.")
```
mfherbst: I'm not happy about that, because that's a feature we do want and need. Where does it fail?

mfherbst: Actually the example still seems to work and produce the right results.

mfherbst: I think we should add a "proper" core-hole test including an XES calculation and comparison of energies ... I'll do that.

mfherbst: XES test passes now ... the point is that with the new interface and mechanisms, the requirements that the occupied orbitals are sorted first is no longer given, since the MoSpaces now resorts as needed. I'll proceed to add a functionality test for XES.

maxscheurer: I really don't know why this works... but if you do, I'm okay with it 😄

mfherbst: The reordering py-side was a hack because the "old" adcc-MoSpaces-interface did assume the first mos were occupied and the later ones were not. This has been changed a while back. Essentially the fact that core-hole states have an empty core orbital now no longer matters as far as the adcc interface is concerned, so we do not need to assert and test this. That's all there is to it.

adcc/backends/test_psi4.py
mfherbst: removed.

adcc/backends/psi4.py
mfherbst: fixed.

adcc/backends/test_psi4.py
mfherbst: resolved.

adcc/backends/test_pyscf.py
mfherbst: resolved.

adcc/backends/test_backends_crossref.py
mfherbst: @maxscheurer You're call :smile:

mfherbst: Actually ... this could be made `ch2nh2` as well, don't you think?

maxscheurer: The indent level at this point does not matter for the code to work properly, but it's more clear if lines 58/59 were out-dented once.

`ch2nh2` is okay for me here.

maxscheurer: Done.

maxscheurer:
G2G one Travis is done.

PR #21, opened 2019-10-31
Changes for integration with Psi4

### LINKED ISSUE #6: Check host program versions for supported version
When first using a particular backend, adcc should check whether this backend version is “known to work” and if not, it should issue a warning.

mfherbst:
This PR introduces a few changes for simplifying and hardening the integration with Psi4 (see https://github.com/psi4/psi4/pull/1744). I'll add todos and notes as I go along

- [x] Closes #6 
- [x] Check for no density-fitting and choleski in SCF reference


mfherbst:
With psi4/psi4#1744 progressing towards installation and tests, I'd say this can go in as I do not expect more changes to be required in adcc, except #20.

PR #22, opened 2019-11-02
Implementation of an amplitudes printer

mfherbst:
This PR implements a `describe_amplitude` function for `ExcitedStates`, where Tensor indices can be formatted in various ways (using HOMO/LUMO and similar labels, the adcc index' convention or the host programm's index convention). Some more generalisation and cleanup of the code is needed.


mfherbst:
@maxscheurer This can go in now. This PR introduces a couple of classes, which can be used for formatting tensor indices in general. This way one can also get a good overview over other kind of tensors (e.g. ERI elements, density matrix elements, ...) and display the tensor indices in various ways (adcc indexing, HF provider indexing, relative to HOMO / LUMO etc.).

adcc/MoSpaces.py
```diff
@@ -20,79 +20,157 @@
...
 
 import libadcc
 
+from .backends import import_scf_results
+from .memory_pool import memory_pool
+
 __all__ = ["MoSpaces"]
 
 
+def expand_spaceargs(hfdata, **spaceargs):
```
maxscheurer: I guess this was moved here from the C-side?

maxscheurer: Never mind... I saw now that it came from `ReferenceState`...

mfherbst: Yes indeed. It came from `ReferenceState`, where it was a duplicate from `C`-side for reasons I cannot remember, so good it's now only once at an understandable place.

PR #23, opened 2019-11-05
Detect and honour an MKL numpy when building adccore

### LINKED ISSUE #16: MKL support
Enable support for MKL.

- "normal" MKL from an Intel installation
- conda-provided MKL

mfherbst:
I tried it with a locally built numpy, linked to MKL. I did not try it with conda, but I leave that to @maxscheurer  :smile:.

Closes #16.

### DIFF
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -139,6 +139,23 @@ This is done to avoid automatically overwriting some development changes
 you might have made inside ``adccore``.
 
 
+Building adccore with MKL support
+---------------------------------
+
+If you have full source code access
+and you are able to follow the :ref:`adccore-sources`,
+the `Intel Math Kernel Library (R) <https://software.intel.com/en-us/mkl>`_
+can also be integrated into adccore and thus adcc.
+In fact this integration happens automatically during the build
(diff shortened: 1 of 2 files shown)


maxscheurer:
I will try it with `conda`-provided `numpy` later today and approve if everything works fine 🤓

mfherbst:
Well ... in either case there are features in master, which won't work, because the changes introduced in 0.13.4 of adccore only work with the `amplitudes-printer` branch. Since the latter is not yet finished, so I'd wait for that with adccore version bump.

Two different intertwined repos is not that much fun ...

maxscheurer:
As you said, I have to test it with an "older" version of adccore, because otherwise this branch breaks when compiling the extensions requiring changes from #22 ... It's really horrible to have these two repos. :angry:

mfherbst:
Ok ... we have two options ... wait until #22 is merged and then merge this, but I'll definitely need another few days for that, or else merge now and live with the fact that master is kind of unstable (in some sense this is an example chase for how to deal with such situations in general for the future). I'd say we wait, but what's your opinion @maxscheurer ?

maxscheurer:
Let's wait.

PR #24, opened 2019-11-11
Add deployment to pypi

mfherbst:
This PR adds deployment to pypi directly from Travis.

I'd say with this merged we can tag and deploy `0.13.2`.


mfherbst:
This is basically done now since 0.13.2 successfully released.

PR #29, opened 2019-11-16
Implementation of various features for the benchmarking suite

mfherbst:
- [x] Cleaner threading interface (plus its documentation)
- ~[ ] CVS memory estimate~ (now in #32)
- ~[ ] Check for memory estimate in `run_adc`.~ (now in #32)

Once this is merged: Use benchmark suite to determine suitable default value for `n_threads_total` (includes *ready* threads not yet running), such that this is fixed once and for all. My guess is that `n_threads_total = n_threads` will work fine ... the  `n_threads_total = 2*n_threads -1` has always been fishy to me.


mfherbst:
Depends whether this gives the best speeds or not.

PR #30, opened 2019-11-21
Integrate PE energy corrections

maxscheurer:
Integrate computations of perturbative state-specific and linear-response-type energy corrections
for PE solvent models:

<a href="https://www.codecogs.com/eqnedit.php?latex=\fn_jvn&space;\Delta&space;E^{0\rightarrow&space;n}_{\text{ptLR}}&space;=&space;-&space;\vec{F}_\text{el}[\mat{P^{\text{tr}}}]^\dagger&space;\vec{\mu}_\text{ind}[\mat{P^{\text{tr}}}]&space;\\&space;\Delta&space;E^{0\rightarrow&space;n}_{\text{ptSS}}&space;=&space;-&space;0.5&space;\vec{F}_\text{el}[\mat{\Delta&space;P}]^\dagger&space;\vec{\mu}_\text{ind}[\mat{\Delta&space;P}]" target="_blank"></a>

# Questions:
- [x] What is the best strategy to integrate the "hook" back to the host program?
- [x] If we implement a decorator to automatically generate the fields of `StateView`/`Excitation`, what happens to the names? They need to be transformed from "plural" to "singular" form, i.e., `state.excitation_energies` will need to be `view[0].excitation_energy`. I suppose it's best to specify the name explicitly, like `@excitation_property(name="excitation_energy")`. It's a bit lengthy, but transparent.

# ToDo:
- [x] install `cppe` for tests with `pyscf` on Travis
- [x] tests against static data Q-Chem (Formaldehyde, (ADC(1), ADC(2), ADC(3)) for all host programs (if PE + program available)
- [x] `ExcitedStates` ctor can have additional excitation energy correction functions as parameter
- [x] refactor `Excitation` and add decorators to automatically generate the attributes
    - [x] access to AO density matrices 
- [x] test custom excitation energy corrections
- [x] test `Excitation` properties
- [x] run full tests
- [x] Think about something for __excitation_propery in cached_property decoratior
- [x] to_nd_array or not?

# Status:
- [x] ready for review
- [x] ready for merge


mfherbst:
Thanks for getting a go at this! Right now I'm a bit busy, but I'll have a think about the questions you raised. One aspect is: How general is the problem, i.e. does it only occur for solvent models or could it be that we have other corrections implied by the reference, which we will want to add to the states.

What I'm getting at is that it might make sense to have some sort of a addon interface, where a HfProvider can request extra functions to be executed on the computed states and if such addons are found during / after the Davidson run, they are executed. What I have in mind is like a state-specific slim version of ExcitedStates, which lazily only exposes information about one state. Of course this should somehow be linked to ExcitedStates in the sense that we do not want to keep recomputing the densities if this can be avoided. Not sure, however, if link implies a reference, because maybe you want the "StateView" (need a better name) to live independent of ExcitedStates. But either way you could then have a function `pe_ptss_correction(excitation)` working on the StateView inserting things such as energy corrections or similar. These functions would be constructed and placed inside a list by the HfProvider and in this way indicating that they should be employed.

Just my quick thoughts for now. Comment?

maxscheurer:
> How general is the problem, i.e. does it only occur for solvent models or could it be that we have other corrections implied by the reference, which we will want to add to the states.

That's the point... At the moment, I don't know of anything else than solvent energy corrections...

> Not sure, however, if link implies a reference, because maybe you want the "StateView" (need a better name) to live independent of ExcitedStates.

If we implement such a `StateView`, it needs to be linked by reference. Otherwise, we have duplication issue, in my opinion... So for now, I guess a `StateView` is a bit too complicated to
integrate.

In a first try, we can implement such a `pe_ptss_correction` function working on the whole `ExcitedStates` object, computing the correction for each excited state it finds. The only problem left is then how to expose the result to the user, i.e., where to store the data inside `ExcitedStates`.

PS: Sorry for the delayed reply...

adcc/ReferenceState.py
```diff
@@ -155,7 +155,9 @@ def __init__(self, hfdata, core_orbitals=None, frozen_core=None,
             self.orbital_coefficients, self.conv_tol
         )
 
-        self.addons = hfdata.addon_functions
+        if hasattr(hfdata, "excitation_energy_corrections"):
+            self.excitation_energy_corrections =\
+                hfdata.excitation_energy_corrections
```
mfherbst, line 160:
```python
for name in ["excitation_energy_corrections"]:
    if hasattr(hfdata, name):
         setattr(self, name, getattr(hfdata, name))
```

mfherbst: Think towards the future ;)

maxscheurer: Sure, good idea. I was just trying to fix the existing tests...

maxscheurer:
@mfherbst you can give this a first, initial look if you want.

Some inconsistencies that need to be resolved:

- I think the functions in the backend should be called `pe_ptlr_correction` (singular form), because they work on a single `Excitation`. However, the property of `ExcitedStates` should be `pe_ptlr_corrections` (plural form), because it's an array of all corrections. So it's probably best to introduce a `namedtuple` after all, containing both names. `excitation_energy_corrections` in the backend would then return a list of `namedtuple`s. What do you think?
- The energy corrections need to be marked with `@mark_excitation_property` somehow.

mfherbst:
Regarding the names: I think having both singular and plural form will only keep causing us trouble. Think about the hassle we have to go through at the moment where if we just used one version could just use simple `getattr`, `setattr` magic. What would you think about just moving to the singular form for all properties, i.e. even `states.transition_dipole_moment` or `state.excitation_energy`? They would still return arrays of course.

adcc/testdata/qchem/__init__.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
mfherbst, line 1: TODO to @mfherbst: Merge this with the get testdata script and upload the file to the adc-connect.org server.

adcc/testdata/qchem/generate_qchem_data.py
```diff
@@ -0,0 +1,171 @@
...
+            "basis": basis,
+            "pe": pe,
+            "geometry": geom
+        }
+        generate_qchem_input_file(
+            infile, method, basis, geom, **kwargs
+        )
+        # only works with my (ms) fork of cclib
+        # github.com/maxscheurer/cclib, branch dev-qchem
+        # os.system("qchem {} {}".format(infile, outfile))
```
mfherbst, line 145: Comment.

adcc/testdata/qchem/generate_qchem_data.py
```diff
@@ -0,0 +1,171 @@
...
+            "method": method,
+            "basis": basis,
+            "pe": pe,
+            "geometry": geom
+        }
+        generate_qchem_input_file(
+            infile, method, basis, geom, **kwargs
+        )
+        # only works with my (ms) fork of cclib
+        # github.com/maxscheurer/cclib, branch dev-qchem
```
mfherbst, line 144: Why do you not make a PR there?

adcc/testdata/qchem/generate_qchem_data.py
```diff
@@ -0,0 +1,171 @@
...
+    basissets = ["sto3g", "ccpvdz"]
+    methods = ["adc1", "adc2", "adc3"]
+    qchem_results = {}
+    for method in methods:
+        for basis in basissets:
+            key, ret = dump_qchem("formaldehyde", method, basis,
+                                  potfile=pe_potentials["fa_6w"])
+            qchem_results[key] = ret
+            print(f"Dumped {key}.")
+    hdf5io.save(fname="qchem_dump.hdf5", dictionary=qchem_results)
```
mfherbst, line 167: Since your hdf5 file actually only contains a few scalar datapoints (and no arrays like in the other cases) I could be tempted to just check it into the repository as a yaml file. That saves a lot of hassle with the downloading etc.

Also: Is there any reason for this to live in a subfolder? So far all generation scripts have lived in the main `testdata` folder. You could of course dump the qchem input and output into a subdirectory, but that does not need to be checked in.

mfherbst: Also I would dump the data as close as possible to the structure of the other datafiles (i.e. use bohr units, same hierarchy / names of keys for the properties etc. That makes it easier to mix match and extend as needed.

adcc/backends/test_backends_polembed.py
```diff
@@ -0,0 +1,79 @@
...
+
+import pytest
+
+from ..misc import expand_test_templates
+from .testing import cached_backend_hf
+from ..testdata.static_data import pe_potentials
+from ..testdata.qchem import qchem_data
+
+from scipy import constants
+eV = constants.value("Hartree energy in eV")
```
mfherbst, line 38: Move into testdata file ... see below.

adcc/backends/test_backends_polembed.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+backends = [b for b in adcc.backends.available()
+            if b not in ["molsturm", "veloxchem"]]
+basissets = ["sto3g", "ccpvdz"]
+methods = ["adc1", "adc2", "adc3"]
+
+
+@pytest.mark.skipif(len(backends) == 0,
+                    reason="No backend found.")
```
mfherbst, line 48: one line.

adcc/backends/test_backends_polembed.py
```diff
@@ -0,0 +1,79 @@
...
+                                   potfile=pe_potentials["fa_6w"])
+        state = adcc.run_adc(scfres, method=method,
+                             n_singlets=5, conv_tol=1e-10)
+        assert_allclose(
+            qc_result["excitation_energies_ev"],
+            state.excitation_energies_uncorrected * eV,
+            atol=1e-5
+        )
+        assert_allclose(
+            qc_result["excitation_energies_ev"]
```
mfherbst, line 64: In such cases I like it if the pluses line up. I would therefore add a unary plus before the first line.

adcc/ExcitedStates.py
```diff
@@ -40,6 +40,17 @@
 from .solver.SolverStateBase import EigenSolverStateBase
 
 
+def mark_excitation_property(**kwargs):
+    """
+    Decorator to mark properties of excited states which
+    can be transferred to Excitation
```
mfherbst, line 46: to the Excitation class defined below.

adcc/ExcitedStates.py
```diff
@@ -586,3 +654,60 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
         return ret[:-1]
+
+    @property
+    def excitations(self):
+        """
+        Provides a list of Excitations, i.e., a view to all individual
+        excited states and their properties. Still under heavy development.
+        """
+        class Excitation:
+            pass
```
mfherbst, line 665: Define this as a proper class with docs before the ExcitedStates. It is too important to be hidden in there. Actually it's constructor could take care of parsing "self" (just the ExcitedStates instance as an argument `parent` to the constructor). This would also structure this code better. Perhaps even move it to a different file to stress the importance of this class.

adcc/ExcitedStates.py
```diff
@@ -131,7 +142,8 @@ def format(self, vector):
 
 
 class ExcitedStates:
-    def __init__(self, data, method=None, property_method=None):
+    def __init__(self, data, method=None, property_method=None,
+                 excitation_energy_corrections=None):
```
mfherbst, line 146: `excitation_energy_corrections={}`

adcc/ExcitedStates.py
```diff
@@ -194,17 +209,48 @@ def __init__(self, data, method=None, property_method=None):
...
             if hasattr(data, "excitation_vectors"):
                 self.excitation_vectors = data.excitation_vectors
 
+        # Collect all excitation energy corrections
+        self.__excitation_energies_uncorrected = self.excitation_energies.copy()
+
+        if excitation_energy_corrections is None:
+            excitation_energy_corrections = {}
+        elif not isinstance(excitation_energy_corrections, dict):
+            raise TypeError("excitation_energy_corrections must be a dict.")
```
mfherbst, line 229: Remove.

adcc/ExcitedStates.py
```diff
@@ -194,17 +209,48 @@ def __init__(self, data, method=None, property_method=None):
...
+
+        if hasattr(self.reference_state, "excitation_energy_corrections"):
+            excitation_energy_corrections.update(
+                self.reference_state.excitation_energy_corrections)
```
mfherbst, line 233: Now comes the tricky business: I think what the user passes explicitly should take preference, therefore:
```
merged = self.reference_state.excitation_energy_corrections.copy()
merged.update(excitation_energy_corrections)
excitation_energy_corrections = merged
```

adcc/ExcitedStates.py
```diff
@@ -194,17 +209,48 @@ def __init__(self, data, method=None, property_method=None):
...
+
+        for key in excitation_energy_corrections:
+            corr_function = excitation_energy_corrections[key]
+            if not callable(corr_function):
+                raise TypeError("Elements in excitation_energy_corrections "
+                                "must be callable.")
+            # call the function for exc. energy correction
```
mfherbst, line 240: I think this is clearer:
```
correction = np.array([corr_function(exci) for exci in self.excitations])
setattr(self, key, correction)
self.__excitation_energies += correction
```

adcc/ExcitedStates.py
```diff
@@ -194,17 +209,48 @@ def __init__(self, data, method=None, property_method=None):
...
+            # TODO: add energy corrections with reasonable name to Excitation
+            # add value to excitation energies
+            self.__excitation_energies += getattr(self, key)
+
+    def __len__(self):
+        return self.size
+
+    @property
+    def size(self):
+        return self.excitation_energies.size
```
mfherbst, line 252: self.__excitation_energies.size() ??

adcc/ExcitedStates.py
```diff
@@ -586,3 +654,60 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+        return excitations
+
+
+def transform_to_ao(dm):
+    if not isinstance(dm, OneParticleOperator):
+        raise TypeError("Density matrix must be of type"
+                        " OneParticleOperator.")
+    density = dm.to_ao_basis()
+    density_ao = density[0] + density[1]
+    return density_ao.to_ndarray()
```
mfherbst, line 713: Once the lazy stuff is merged you can just do `sum(dm.to_ao_basis())`. Also I am not sure `to_ndarray()` should be done by default. In either way this will only be one line, so I would not even have this function.

adcc/ExcitedStates.py
```diff
@@ -586,3 +654,60 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+        Provides a list of Excitations, i.e., a view to all individual
+        excited states and their properties. Still under heavy development.
+        """
+        class Excitation:
+            pass
+
+        excitations = []
+        for i in range(self.size):
+            e = Excitation()
+            e.idx = i
```
mfherbst, line 670: `e.index`. The two latters do not hurt. Also I would have a `kind` and maybe a `method` attribute.

adcc/ExcitedStates.py
```diff
@@ -586,3 +654,60 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+            for key in dir(self):
+                if key == "excitations":
+                    continue
+                if "__" in key or key.startswith("_"):
+                    continue  # skip "private" fields
+                if hasattr(type(self), key):
+                    if not isinstance(getattr(type(self), key), property):
+                        continue
+                else:
+                    continue
```
mfherbst, line 681: I would branch differently. This I find very hard to read.

adcc/ExcitedStates.py
```diff
@@ -586,3 +654,60 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+
+                fget = getattr(type(self), key).fget
+                if not hasattr(fget, "__excitation_property"):
+                    # print(dir(fget), key, "not found.")
```
mfherbst, line 685: remove

adcc/ExcitedStates.py
```diff
@@ -586,3 +654,60 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+                    continue
+                # Extract the kwargs passed to mark_excitation_property
+                kwargs = getattr(fget, "__excitation_property").copy()
+
+                def payload(self, key=key, kwargs=kwargs):
```
mfherbst, line 690: Payload is not a good name. I just chose it in my trial code.

adcc/testdata/qchem/generate_qchem_data.py
maxscheurer: I'd have to think about it. Some of the stuff could definitely go in there, but I also changed a lot of units that `cclib` stores its data in. Who wants excitation energies in cm^-1 ?! 😄

adcc/ExcitedStates.py
maxscheurer: I'd rather not do this, because it makes the default argument mutable from the outside. But I guess in this specific case, it does not do us any harm 👍

.travis.yml
```diff
@@ -19,6 +19,7 @@ addons:
 
 install:
   - pip3 install pyscf
+  - pip3 install git+https://github.com/maxscheurer/cppe.git
```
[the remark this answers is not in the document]
maxscheurer: For the time being this will stay as is. We can exchange it later on.

adcc/backends/__init__.py
```diff
@@ -155,6 +155,8 @@ def run_hf(backend, xyz, basis, **kwargs):
         conv_tol:       energy convergence tolerance
         conv_tol_grad:  convergence tolerance of the electronic gradient
         max_iter:       maximum number of SCF iterations
+        potfile:        name of the potential file for polarizable embedding
+        TODO: change to option dictionary for PE?
```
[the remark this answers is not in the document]
maxscheurer: Done.

adcc/backends/psi4.py
```diff
@@ -89,6 +89,28 @@ def __init__(self, wfn):
...
+                                                        elec_only=elec_only)
+        return e_pe
+
+    def pe_ptss_correction(self, view):
+        dm = view.state_diffdm_ao
+        return self.pe_energy(dm, elec_only=True)
+
+    def pe_ptlr_correction(self, view):
+        dm = view.transition_dm_ao
+        return 2.0 * self.pe_energy(dm, elec_only=True)
```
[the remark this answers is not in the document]
maxscheurer: Done.

adcc/backends/pyscf.py
```diff
@@ -109,6 +110,28 @@ def __init__(self, scfres):
...
+        e_pe, _ = pe_state.kernel(dm, elec_only=elec_only)
+        return e_pe
+
+    def pe_ptss_correction(self, view):
+        dm = view.state_diffdm_ao
+        return self.pe_energy(dm, elec_only=True)
+
+    def pe_ptlr_correction(self, view):
+        dm = view.transition_dm_ao
+        return 2.0 * self.pe_energy(dm, elec_only=True)
```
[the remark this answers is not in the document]
maxscheurer: Done.

adcc/backends/pyscf.py
```diff
@@ -221,7 +244,7 @@ def import_scf(scfres):
 
 
 def run_hf(xyz, basis, charge=0, multiplicity=1, conv_tol=1e-11,
-           conv_tol_grad=1e-9, max_iter=150):
+           conv_tol_grad=1e-9, max_iter=150, potfile=None):
```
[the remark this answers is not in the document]
maxscheurer: Done.

adcc/backends/psi4.py
```diff
@@ -197,7 +219,7 @@ def import_scf(wfn):
 
 
 def run_hf(xyz, basis, charge=0, multiplicity=1, conv_tol=1e-11,
-           conv_tol_grad=1e-8, max_iter=150):
+           conv_tol_grad=1e-8, max_iter=150, potfile=None):
```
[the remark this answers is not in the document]
maxscheurer: Done.

adcc/backends/testing.py
```diff
@@ -177,7 +177,8 @@ def operator_import_from_ao_test(scfres, ao_dict, operator="electric_dipole"):
             )
 
 
-def cached_backend_hf(backend, molecule, basis, multiplicity=1, conv_tol=1e-12):
+def cached_backend_hf(backend, molecule, basis, multiplicity=1, conv_tol=1e-12,
+                      potfile=None):
```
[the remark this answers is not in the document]
maxscheurer: Done.

adcc/backends/veloxchem.py
```diff
@@ -122,6 +122,26 @@ def __init__(self, scfdrv):
...
+        e_pe, _ = self.scfdrv.pe_drv.get_pe_contribution(dm, elec_only=elec_only)
+        return e_pe
+
+    def pe_ptss_correction(self, view):
+        dm = view.state_diffdm_ao
+        return self.pe_energy(dm, elec_only=True)
+
+    def pe_ptlr_correction(self, view):
+        dm = view.transition_dm_ao
+        return 2.0 * self.pe_energy(dm, elec_only=True)
```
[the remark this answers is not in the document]
maxscheurer: Done.

examples/pna/psi4_adc2_pna_6w_pol_embed.py
```diff
@@ -0,0 +1,66 @@
...
+    """)
+
+# mol = psi4.geometry("""
+#     O 0 0 0
+#     H 0 0 1.795239827225189
+#     H 1.693194615993441 0 -0.599043184453037
+#     symmetry c1
+#     units au
+#     """)
+
```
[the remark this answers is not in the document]
maxscheurer: Done (forgot to commit)

examples/pna/psi4_adc2_pna_6w_pol_embed.py
```diff
@@ -0,0 +1,66 @@
...
+psi4.core.set_num_threads(4)
+# psi4.core.be_quiet()
+psi4.set_options({'basis': "sto-3g",
+                  'scf_type': 'pk',
+                  'pe': 'true',
+                  'e_convergence': 1e-10,
+                  'd_convergence': 1e-10})
+psi4.set_module_options(
+    "pe", {"potfile": "pna_6w.pot"}
+)
```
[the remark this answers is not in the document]
maxscheurer: Done.

adcc/testdata/qchem/generate_qchem_data.py
maxscheurer: Done.

adcc/backends/test_backends_polembed.py
maxscheurer: Done.

adcc/backends/test_backends_polembed.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: True. Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Renamed.

adcc/ExcitedStates.py
maxscheurer: Done. Left a comment for the lazy stuff. About the `to_ndarray()`, I'm also not sure at the moment.

maxscheurer:
> What would you think about just moving to the singular form for all properties, i.e. even states.transition_dipole_moment or state.excitation_energy? They would still return arrays of course.

I think that's a good idea. Should I go for this, and do we need to first mark the properties deprecated? I guess we don't have so many users whose code we break with this change.

mfherbst:
Yes please go for it in this PR. About deprecation: I think we have a growing user base at the moment, which I would not want to make uncomfortable. I think I would indeed add deprecation warnings. Main reason is, however, that we likely forget to update something in the docs and then at least our examples will still work.

adcc/ExcitedStates.py
```diff
@@ -216,40 +216,37 @@ def __init__(self, data, method=None, property_method=None,
             if hasattr(data, "eigenvectors"):
                 self.excitation_vectors = data.eigenvectors
             if hasattr(data, "excitation_energies"):
-                self.__excitation_energies = data.excitation_energies
+                # TODO: what if data has energy corrections?
+                # make better ctor if data is other ExcitedStates object
```
mfherbst, line 220: That's a good point and definitely happens for CIS for example. I think this should be handled properly.

adcc/ExcitedStates.py
```diff
@@ -216,40 +216,37 @@ def __init__(self, data, method=None, property_method=None,
...
                 raise TypeError("Elements in excitation_energy_corrections "
                                 "must be callable.")
             # call the function for exc. energy correction
-            setattr(self, key, np.array([
-                corr_function(exci) for exci in self.excitations]))
-            # TODO: add energy corrections with reasonable name to Excitation
-            # add value to excitation energies
-            self.__excitation_energies += getattr(self, key)
+            correction = np.array([
+                corr_function(exci) for exci in self.excitations])
```
mfherbst, line 240:
```
np.array([corr_function(exci)
          for exci in self.excitations])
```
maybe?

adcc/ExcitedStates.py
```diff
@@ -667,47 +664,40 @@ class Excitation:
...
 
                 transform = kwargs.pop("transform_to_ao", False)
                 if transform:
-                    def payload_transform(self, key=key):
-                        return transform_to_ao(
-                            getattr(self.__parent_state, key)[self.idx])
+                    def get_parent_property_transform(self, key=key):
+                        matrix = getattr(self.__parent_state, key)[self.index]
+                        # TODO: make it a one-liner once the lazy tensor stuff
+                        # is merged
```
mfherbst, line 695: By the way: This just waits for your approval ;)

maxscheurer: Yes, I know. I think we should merge the lazy stuff first, because refactoring `ExcitedStates` will take some time and thinking (with deprecations, "copy"-ctor, etc.).

examples/water/pyscf_ccpvdz_adc2_spectrum.py
```diff
@@ -10,7 +10,7 @@
     atom='O 0 0 0;'
          'H 0 0 1.795239827225189;'
          'H 1.693194615993441 0 -0.599043184453037',
-    basis='cc-pvdz',
+    basis='sto-3g',
```
mfherbst, line 13: This probably unintended.

adcc/backends/test_backends_polembed.py
```diff
@@ -0,0 +1,79 @@
...
+                             n_singlets=5, conv_tol=1e-10)
+        assert_allclose(
+            qc_result["excitation_energies_ev"],
+            state.excitation_energy_uncorrected * eV,
+            atol=1e-5
+        )
+        assert_allclose(
+            + qc_result["excitation_energies_ev"]
+            + qc_result["pe_ptss_corrections_ev"]
+            + qc_result["pe_ptlr_corrections_ev"],
```
mfherbst, line 66: These should probably now be singular, too.

adcc/ExcitedStates.py
```diff
@@ -193,17 +208,75 @@ def __init__(self, data, method=None, property_method=None):
...
+            "oscillator_stenths_velocity": "oscillator_strength_velocity",
+            "rotatory_strengths": "rotatory_strength",
+            "excitation_vectors": "excitation_vector",
+        }
+
+        for dep_property in deprecated:
+            new_key = deprecated[dep_property]
+
+            def deprecated_property(self, key=new_key, old_key=dep_property):
+                # TODO: not sure about the version?
```
mfherbst, line 267: That's fine.

adcc/ExcitedStates.py
```diff
@@ -193,17 +208,75 @@ def __init__(self, data, method=None, property_method=None):
...
+            if not callable(corr_function):
+                raise TypeError("Elements in excitation_energy_corrections "
+                                "must be callable.")
+            # call the function for exc. energy correction
+            correction = np.array([
+                corr_function(exci) for exci in self.excitations])
+            setattr(self, key, correction)
+            self.__excitation_energy += correction
+
+        # FIXME: poor man's deprecation warnings
```
mfherbst, line 244: I would do it no different, however I would put it below the class and not into the constructor.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

examples/water/pyscf_ccpvdz_adc2_spectrum.py
maxscheurer: Done.

adcc/testdata/static_data/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+        return fn
+    else:
+        return ""
+
+
+_potential_files = {
+    "fa_6w": "fa_6w.pot",  # Formaldehyde + 6 Water molecules
+}
+
+pe_potentials = {key: fullfile(_potential_files[key]) for key in _potential_files}
```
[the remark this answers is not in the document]
maxscheurer: Done. `xyz` and `pe_potentials` are now in a single `static_data.py` file.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: `kind` is not always set, so I left it out. Otherwise done.

adcc/Excitation.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    def inner(f, kwargs=kwargs):
+        f.__excitation_property = kwargs
+        return f
+    return inner
+
+
+class Excitation:
+    def __init__(self, parent_state, index, method):
+        """Construct an Excitation from an ExcitedStates parent object
```
mfherbst, line 47: ... from an :class:`adcc.ExcitedStates` parent object

adcc/testdata/generate_hfdata_ch2nh2_sto3g.py
```diff
@@ -23,7 +23,7 @@
 import sys
 
 from pyscf import gto, scf
-from geometry import xyz
+from .static_data import xyz
```
mfherbst, line 26: I think this won't work. It has to be an absolute import in these cases.

adcc/testdata/generate_hfdata_cn_ccpvdz.py
```diff
@@ -24,7 +24,7 @@
 import numpy as np
 
 from pyscf import gto, scf
-from geometry import xyz
+from .static_data import xyz
```
mfherbst, line 27: same here ... etc ...

adcc/testdata/static_data.py
```diff
@@ -67,4 +78,18 @@
...
+    "formaldehyde": """
+    C 2.0092420208996 3.8300915804899 0.8199294419789
+    O 2.1078857690998 2.0406638776593 2.1812021228452
+    H 2.0682421748693 5.7438044586615 1.5798996515014
+    H 1.8588483602149 3.6361694243085 -1.2192956060942
+    """
 }
+
+_potential_files_relative = {
+    "fa_6w": "potentials/fa_6w.pot",  # Formaldehyde + 6 Water molecules
```
mfherbst, line 91: why not just `os.path.join(thisdir, "potentials/fa_6w.pot")` and skip the fullfile function completely?

adcc/ExcitedStates.py
maxscheurer: Handled.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/backends/test_backends_polembed.py
maxscheurer: Yes, so I guess the test data from Q-Chem are the last major thing to fix now...

adcc/ExcitedStates.py
```diff
@@ -582,7 +643,78 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+
+                transform = kwargs.pop("transform_to_ao", False)
+                if transform:
+                    def get_parent_property_transform(self, key=key):
+                        matrix = getattr(self.parent_state, key)[self.index]
+                        density_ao = sum(matrix.to_ao_basis())
+                        return density_ao.to_ndarray()
+                    setattr(type(e), key + "_ao",
+                            property(get_parent_property_transform))
+            excitations.append(e)
```
mfherbst, line 689: I would actually put this into the Excitation constructor as well. What do you think?

adcc/testdata/generate_hfdata_ch2nh2_sto3g.py
maxscheurer: Done.

adcc/testdata/generate_hfdata_cn_ccpvdz.py
maxscheurer: Done for all.

adcc/ExcitedStates.py
maxscheurer: Let me try quickly if it works.

maxscheurer: Well... I guess then we have a cyclic import? `ExcitedStates.py` imports `Excitation` and the other way around?

mfherbst: Why? you never need the class definition, you only need the instance variable and `dir` and `getattr` and such magic. Maybe I miss something, but I don't see right now why `Excitation`  needs to know anything from `ExcitedStates.py`.

maxscheurer: You're right, I made a wrong assumption about the code 😉 
I like the split version much better now 👍

adcc/testdata/static_data.py
maxscheurer: Done.

adcc/Excitation.py
maxscheurer: Done.

maxscheurer:
I just squashed some commits together, so we don't have too many commits when merging.

adcc/testdata/qchem/generate_qchem_data.py
maxscheurer: I'm for the yaml solution. How would we deal with the `np` arrays?
I guess `tolist`, otherwise we get binary stuff in there as well.

mfherbst:
Completely fine. Wow this starts to look really good. Great work @maxscheurer!

adcc/testdata/qchem/generate_qchem_data.py
mfherbst: Yes I think that is the best option. There is also the "proper" way to use representers, see [molsturm](https://github.com/molsturm/molsturm/blob/master/src/interface/python/molsturm/yaml_utils.py), but I'm not sure that is appropriate here. If we only need it once it's a bit over the top.

maxscheurer: I'd say we add the representers... Can probably be used if we want to serialized `ExcitedStates` to yaml? We could create a `yamlio.py`
(with the user-exposed functions `save` and `load`)
analogous to `hdf5io.py`? I think I'd enjoy having that 😄

mfherbst: Well I'm no longer a fan of hdf5io, actually. It's really not such a well-written piece of code and especially it is not perfectly compatible to the rest of the world. That same problem we have if we use our custom representers for yaml, but ok. If you think it helps, than go for it.

maxscheurer: The "outside" world argument is a good point. I guess I'll convert the arrays to lists, and convert back to numpy array upon read-in. Then we add less code, which is kind of good.

adcc/testdata/cache.py
```diff
@@ -234,3 +236,27 @@ def adc_states(self):
...
+
+
+class QChemData:
+    @property
+    def data(self):
+        thisdir = os.path.dirname(__file__)
+        yaml_file = os.path.join(thisdir, "qchem_dump.yml")
+        with open(yaml_file, "r") as f:
+            data_raw = yaml.safe_load(f)
+        return lists_to_ndarray(data_raw)
```
mfherbst, line 259: why do you even need this class if all it does is being constructed and then the data property called? Just make it a function.

maxscheurer: Done.

adcc/testdata/generate_qchem_data.py
```diff
@@ -0,0 +1,177 @@
...
+        ret = {
+            "molecule": molecule,
+            "method": method,
+            "basis": basis,
+            "pe": pe,
+            "geometry": geom
+        }
+        generate_qchem_input_file(
+            infile, method, basis, geom, **kwargs
+        )
```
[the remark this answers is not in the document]
maxscheurer: Yes.

adcc/backends/test_backends_polembed.py
maxscheurer: Done.

adcc/testdata/qchem/generate_qchem_data.py
maxscheurer: Done.

adcc/backends/test_backends_polembed.py
maxscheurer: Done.

mfherbst:
As far as I see it the two open questions are:
- What to do with `__excitation_property` in `cached_property` decorator
- Keep the `to_ndarray` in the to AO transformation.

As far as I am concerned I could imagine inventing a special prefix to solve the first problem (e.g. `__fwd_excitation_property`) and just blindly forward all properties starting with `__fwd`.

For the second thing I am still undecided. If we do it like now the advantage is that the copy to the numpy array is only done once. The disadvantage is that it breaks the pattern that our interface never returns numpy arrays of non-trivial quantities (such as matrices). I kind of value that consistency as important and would be willing to call `to_ndarray` explicitly in the backend.

@maxscheurer Your thoughts?

maxscheurer:
- I have no problem with the current solution for forwarding `__excitation_property`. If we need to forward more of these guys, we can always add the `__fwd` prefix. It's not too much code at the moment.
- I agree. Let's call `to_ndarray` in the backend then.

mfherbst:
I fully agree.

maxscheurer:
Full tests also seem to be fine 👍

mfherbst:
So there is no more blocker for the merge left? I'm a bit surprised how fast this now went ... good job!

maxscheurer:
Not from my point of view. I'm going to play around with the conda stuff in a separate PR.

mfherbst:
Ok then ...

mfherbst:
So you rather wait a bit more for a release, because I would've made one first thing tomorrow (0.15.0)

maxscheurer:
Let's wait until Friday? Then I can do some test runs with conda.

mfherbst:
Sounds good. Maybe I can get a first set of more advanced opt_einsum features working until then.

PR #31, opened 2019-12-05
Correctly print transition dipole moments

maxscheurer:
Print the transition dipole moments in `describe()` instead of state dipole moments 😄


mfherbst:
Oops ... copy and paste error ... my bad.

PR #32, opened 2019-12-18
Implement crude estimates for memory requirements

### LINKED ISSUE #27: Add memory estimation function
It was brought up in psi4/psi4#1744 that a memory estimate for ADC calculations could be a very helpful feature. What could be helpful is to allow the estimation of the memory needed for a Tensor before allocating it, based on its symmetry and generalise the notion from there to the whole of adcc.

mfherbst:
- [x] Full ADC memory estimate
- [ ] CVS memory estimate
- [ ] Check for memory estimate in `run_adc`.

Fixes #27.


adcc/estimate_memory.py
```diff
@@ -0,0 +1,170 @@
...
+    if method.name in ["adc2", "adc2x"]:
+        count["oo"] += 1  # I1 intermediate
+        count["vv"] += 1  # I2 intermediate
+        eriimport = "vvvv"
+
+    if method.level >= 3:
+        count["ovov"] += 1  # M11 intermediate
+        count["ovvv"] += 3  # t2eri + Pia intermediate
+        count["ooov"] += 3  # t2eri + Pib intermediate
+        eriimport = "vvvv"
```
maxscheurer, line 109: Probably we should also count the oooo and vvvv block if properties are requested?

mfherbst: Good point, but this PR is far away from such micro-optimisation at the moment, unfortunately ... the results are still incredibly unreliable.

mfherbst:
Closing this now ... #71 will changes so much of the workflow, that this work is most likely outdated anyway.

PR #36, opened 2020-01-05
Add binder support

mfherbst:
This PR adds a very basic support for running adcc in binder. It's available simply via https://try.adc-connect.org (once this PR is merged), so basically trivial to get others to try adcc without installation. Better examples should be added later.

Current issues are that adcc is actually built from source ... which is pretty slow. Also the example is all but good for now.


maxscheurer:
> Current issues are that adcc is actually built from source ...

Why can't we pull it from conda?

mfherbst:
Because than it's not the version from master.

PR #37, opened 2020-01-17
ADC matrix blockview

mfherbst:
Datastructure to transparently view only the singles or doubles block of a bigger ADC matrix. Works with the Jacobi-Davidson. Others have not been tested, but should work as well.

Does not work for the doubles part because right now AmplitudeVector assumes that the first block is the singles block (and if we zoom into doubles only than the first block is the doubles block). I'll not attempt to cure that for now, but just to keep that in mind.

Tests are coming ...

### DIFF
--- a/adcc/AdcBlockView.py
+++ b/adcc/AdcBlockView.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+## vi: tabstop=4 shiftwidth=4 softtabstop=4 expandtab
+## ---------------------------------------------------------------------
+##
+## Copyright (C) 2020 by the adcc authors
+##
+## This file is part of adcc.
+##
+## adcc is free software: you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published
+## by the Free Software Foundation, either version 3 of the License, or
+## (at your option) any later version.
+##
+## adcc is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with adcc. If not, see <http://www.gnu.org/licenses/>.
+##
+## ---------------------------------------------------------------------
+import numpy as np
+
+from .AdcMatrix import AdcMatrixlike
+from .functions import empty_like
+from .AmplitudeVector import AmplitudeVector
+
+import libadcc
+
+
+class AdcBlockView(AdcMatrixlike):
+    def __init__(self, fullmatrix, block):
+        """
+        Object which acts as a single block of the ADC matrix.
+
+        Parameters
+        ----------
+        fullmatrix : AdcMatrixlike
+            The full matrix of which one wants to view a block
+        block : str
+            The block to view. Valid choices include `s` (view the
+            singles-singles block)or `d` (view the doubles-doubles block).
+        """
+        if block not in fullmatrix.blocks:
+            raise ValueError(f"Block {block} not a valid block of passed "
+                             "matrix {self.__fullmatrix}")
+        self.__block = block
+        self.__fullmatrix = fullmatrix
+        super().__init__(fullmatrix)
+
+    def __repr__(self):
+        return f"AdcBlockView({self.__fullmatrix},{self.__block})"
+
+    @property
+    def blocks(self):
+        return [self.__block]
+
+    def has_block(self, block):
+        return self.__block == block
+
+    def __len__(self):
+        return np.prod([
+            self.mospaces.n_orbs(sp) for sp in self.block_spaces(self.__block)
+        ])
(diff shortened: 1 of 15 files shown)


adcc/AdcBlockView.py
```diff
@@ -0,0 +1,120 @@
...
+from .functions import empty_like
+from .AmplitudeVector import AmplitudeVector
+
+import libadcc
+
+
+class AdcBlockView(AdcMatrixlike):
+    def __init__(self, fullmatrix, block):
+        """
+        Object which acts as a single block of the ADC matrix.
```
maxscheurer: Please provide an example for `block`. Is it a string like `ss`, `sd`, `ds`, and `dd`?

adcc/AdcBlockView.py
```diff
@@ -0,0 +1,120 @@
...
+            if not out_ampl.blocks == [self.__block]:
+                raise ValueError("out_ampl does not consist of the "
+                                 "correct blocks")
+            out_ampl = out_ampl[self.__block]
+        elif not isinstance(out_ampl, libadcc.Tensor):
+            raise TypeError("out_ampl has to be of type AmplitudeVector "
+                            "or Tensor.")
+
+        # Compute the matrix-vector product
+        super().compute_apply(2 * self.__block, in_ampl, out_ampl)
```
maxscheurer: This looks as if one can only view a diagonal block (also fits the description of the PR)?
Then the class should be called `AdcDiagonalBlockView`.

adcc/workflow.py
```diff
@@ -202,68 +234,70 @@ def run_adc(data_or_matrix, n_states=None, kind="any", conv_tol=None,
...
-    if conv_tol is None:
-        conv_tol = max(refstate.conv_tol / 100, 1e-6)
-    if refstate.conv_tol > conv_tol:
-        raise ValueError("Convergence tolerance of SCF results (== {}) needs to"
-                         " be lower than ADC convergence tolerance parameter "
-                         "conv_tol (== {}).".format(refstate.conv_tol,
-                                                    conv_tol))
-
-    # Normalise guess parameters
+def normalise_state_parameters(reference_state, n_states=None, n_singlets=None,
```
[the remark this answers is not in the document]
mfherbst: Agreed.

adcc/solver/preconditioner.py
```diff
@@ -51,7 +51,7 @@ class JacobiPreconditioner:
     D is the diagonal of the adcmatrix.
     """
     def __init__(self, adcmatrix, shifts=0.0):
-        if not isinstance(adcmatrix, AdcMatrix):
+        if not isinstance(adcmatrix, AdcMatrixlike):
             raise TypeError("Only an AdcMatrix may be used with this "
```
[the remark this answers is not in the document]
mfherbst: Good catch.

adcc/workflow.py
```diff
@@ -377,13 +411,12 @@ def estimate_n_guesses(matrix, n_states, singles_only=True):
     return max(n_states, n_guesses)
 
 
-def find_guesses(guessfctn, matrix, n_guesses, n_guesses_doubles=None):
+def obtain_guesses_by_inspection(matrix, n_guesses, kind, n_guesses_doubles=None):
```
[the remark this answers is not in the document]
mfherbst: Well that's kind of deliberate. The idea is that unlike solving an approximation or so, one just looks at the elements ... in this case only the diagonals, but that might not necessarily be the case. Also there is already `guesses_from_diagonal`, which is the core building block to the current function, so I wanted to use some other name to avoid confusion.

If you think that's being too general in the name, I'm happy to change it.

adcc/AdcBlockView.py
mfherbst: I think than I'd scratch the `View` to make it a more handy name. How about `AdcDiagonalBlock`?

adcc/AdcBlockView.py
mfherbst: No it's just "s" or "d", because that's the easiest to code. I mean for example for "ds" there is no expression in libadc afaik, so implementing that gets kind of annoying. We could leave that for future work, however. I do not see for what we would need that at the moment, however. Do you have a use case in mind.

mfherbst:
Right now it's used to be able to solve only the singles part of ADC(2) and beyond, so nothing fancy, but that's of course a building block for other algorithms (DIIS for ADC(2) for example), so it's good to have, I think.

mfherbst:
@maxscheurer ping

maxscheurer:
I‘m on holiday, expect sth over the weekend.

mfherbst:
Enjoy.

adcc/AdcBlockView.py
maxscheurer: I think it's okay like that, I was just a little bit confused.

adcc/workflow.py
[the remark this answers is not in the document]
maxscheurer: Okay, let's leave it as is.

PR #49, opened 2020-03-13
Conda suggestions

mfherbst:
Some suggestions to hopefully solve https://github.com/psi4/psi4/pull/1848.

- Output of failing run on Azure: https://dev.azure.com/psi4/psi4/_build/results?buildId=1779&view=logs&j=96451287-da82-57a8-2c11-8da7db9ab71a&t=41b51342-9bfb-52c2-7efb-0050141a3f7e

My rationale for the change:
- We need HDF5 only via h5py. Any version of h5py above the specified boundary should be good. No need to pin.
- Same for tiqn, matplotlib, scipy.
- HDF5 now pinned to a version compatible with chemps2

Any other suggestions?

### DIFF
--- a/conda/meta.yaml.in
+++ b/conda/meta.yaml.in
@@ -32,11 +32,11 @@ requirements:
   run:
     - {{ pin_compatible('python', max_pin='x.x') }}
     - {{ pin_compatible('mkl', max_pin='x') }}
-    - {{ pin_compatible('h5py', max_pin='x.x.x') }}
     - {{ pin_compatible('numpy', max_pin='x.x') }}
-    - {{ pin_compatible('scipy', max_pin='x') }}
-    - {{ pin_compatible('matplotlib', max_pin='x') }}
-    - {{ pin_compatible('tqdm', max_pin='x') }}
+    - h5py >=2.9
+    - scipy >=1.2
+    - matplotlib >=3.0
+    - tqdm >=4.30
 
 test:
   imports:


mfherbst:
Installing psi4 from `psi4/label/dev` in the CI fails with the very strange ResolvePackageNotFound exception (after the Package plan has gone through and has been resoved!):
```
conda.exceptions.ResolvePackageNotFound: 
  - psi4[version='>=1.4.a2.0.DEV,<1.5a']
```
I'm not sure what to make from this. Maybe we should wait until 1.4 is stable before installing it here ...

loriab:
For background, hdf5 pinning is at odds between [defaults](https://github.com/AnacondaRecipes/aggregate/blob/master/conda_build_config.yaml#L97-L98) and [c-f](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml#L426-L427). To help out xtensor integration in Nov, I built the Linux psi stack (which is usually defaults-based) against 1.10.5 and then copied that package over to `psi4/label/dev` https://anaconda.org/psi4/hdf5/files so that plain `-c psi4/label/dev` continued to work.

mfherbst:
Thanks for the quick reply. So this means if we use hdf5 1.10.5 on linux for our builds, then we should be back in business for the Psi4 tests, right?

loriab:
psi4 isn't needed to actually build adcc, right? it should probably be in the `test/requires` section, not in `requirements/host`. it's kind of you to take us into consideration at package-building stage, but it's also not unreasonable for you to leave that testing to downstream. (smooth who-tests-what upon delta-code is something I keep in mind but haven't got a master plan for.)

mfherbst:
Well you're right in theory, but for $reasons we use a hack where we test during the build stage, so unfortunately if we want to test against psi4, then it needs to be done as it is for now.

What we test is mainly our interfaces into Psi4 and pyscf.

Since adcc/psi4 are kind of separate modules but kind of depend on another (we need Psi4's SCF, Psi4 uses our ADC), I'm not that sure who is upstream and who is downstream :smile:. And too much testing never hurts, was our rationale.

loriab:
> So this means if we use hdf5 1.10.5 on linux for our builds, then we should be back in business for the Psi4 tests, right?

Yes, but it also means my questionable 1.10.5 choice is now an unnecessary constraint for you. You could consider building separate pkgs for .4 and .5. Or, stick with the `h5py >=2.9` bit and leave the hdf5 decision to the channel (looks like psi4/label/dev will drag in .5 via https://anaconda.org/psi4/h5py/files). (Neither of those is a fully-baked plan, so consider them hints, not guidance.)

> So this means if we use hdf5 1.10.5 on linux for our builds, then we should be back in business for the Psi4 tests, right?

There's also the `GG` error over at psi4, but unless there's a weird packages constraint forbidding gau2grid >=2 involved, adcc is cleared of suspicion.

> Since adcc/psi4 are kind of separate modules but kind of depend on another (we need Psi4's SCF, Psi4 uses our ADC), I'm not that sure who is upstream and who is downstream 😄. And too much testing never hurts, was our rationale.

ah, right. build upstream/downstream and runtime upstream/downstream get complicated.

loriab:
Yeah, hdf5 to .5 should fix g2g 1. https://dev.azure.com/psi4/psi4/_build/results?buildId=1784&view=logs&j=96451287-da82-57a8-2c11-8da7db9ab71a&t=41b51342-9bfb-52c2-7efb-0050141a3f7e&l=399

mfherbst:
Great. Thanks for your input @loriab. So I think the simplest thing is to do is indeed to loosen the clamps on our package requirements and go with the hdf5 from the channel. I'll build and upload the adcc packages to conda accordingly and then we will see if this works on the Psi4 end.

mfherbst:
@maxscheurer Ready for review :smile:

PR #57, opened 2020-04-29
Ensure set_from_ndarray converts to C-style order

mfherbst:
While I was on it I also removed some unused functions.


mfherbst:
@maxscheurer Because I lazily merged master in, could you just squash it when you approve and merge, thanks!

maxscheurer:
Many thanks for the fix!

PR #64, opened 2020-05-21
Fix small inconsistencies in ExcitedStates

mfherbst:
I think there are a few inconsistencies in ExcitedStates, for example that we only allow using the length gauge for the plot_spectrum. Also the plot_spectrum function is getting a bit messy. I'll try to correct this here.

Some questions to @maxscheurer:
- What gauge is used for our rotatory strengths? Always velocity, right?
- Also do you have a good reference (paper, textbook) for the rotatory strength calculation. I think we should probably add one in the code.

### DIFF
--- a/adcc/ExcitedStates.py
+++ b/adcc/ExcitedStates.py
@@ -348,8 +348,8 @@ class ExcitedStates:
             ["eV", "au", "nm", "cm-1"]
         yaxis : str
             Quantity to plot on the y-Axis. Options are "cross_section",
-            "osc_strength", "dipole" (plots norm of transition dipole),
-            "rotational_strength" (ECD spectrum with rotational strength)
+            "oscillator_strength", "dipole" (plots norm of transition dipole),
+            "rotatory_strengths" (ECD spectrum with rotational strength),
         width : float, optional
             Gaussian broadening standard deviation or Lorentzian broadening
             gamma parameter. The value should be given in atomic units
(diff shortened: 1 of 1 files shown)


maxscheurer:
> What gauge is used for our rotatory strengths? Always velocity, right?

Yes, in `adcc` we only use velocity gauge.

> Also do you have a good reference (paper, textbook) for the rotatory strength calculation

We should ask @Drrehn, he's the expert. I guess it's somewhere in the Molecular Properties book, I'll check...

adcc/ExcitedStates.py
```diff
@@ -348,8 +348,8 @@ def plot_spectrum(self, broadening="lorentzian", xaxis="eV",
             ["eV", "au", "nm", "cm-1"]
         yaxis : str
             Quantity to plot on the y-Axis. Options are "cross_section",
-            "osc_strength", "dipole" (plots norm of transition dipole),
-            "rotational_strength" (ECD spectrum with rotational strength)
+            "oscillator_strength", "dipole" (plots norm of transition dipole),
+            "rotatory_strengths" (ECD spectrum with rotational strength),
```
maxscheurer, line 352: `_strength` or `_strengths`?
Or both should be accepted...

mfherbst: Probably plural, then it's exactly as the properties. Yes we should probably accept both.

maxscheurer: Now it should be the singular form 😄

PR #67, opened 2020-05-29
Update to most recent benchmark results

mfherbst:
The lazy implementation is not all that glorious on the smaller problems after all.


mfherbst:
From my point of view the last blocker before `v0.15.0`. I run the benchmarks at the moment and update this PR with the new figures.

PR #74, opened 2020-06-15
Amend modified transition moments

Drrehn:


adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,43 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
-        raise NotImplementedError("compute_modified_transition_moments "
-                                  "not implemented for CVS")
-
+        if method.name == "cvs-adc1":
+            return AmplitudeVector(dipole_operator['o2v1'])
+        else:
+            cvs_mtm = Adc2CVSModifiedTransitionMoments(method,
+                                                       ground_state,
+                                                       dipole_operator)
+            return cvs_mtm.cvs_adc2_mtm()
```
mfherbst, line 83: I don't see a good reason to make a class for this. Why don't you just implement it as free functions?

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,43 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
                         "libadcc.OneParticleOperator.")
 
     if method.is_core_valence_separated:
-        raise NotImplementedError("compute_modified_transition_moments "
-                                  "not implemented for CVS")
-
+        if method.name == "cvs-adc1":
+            return AmplitudeVector(dipole_operator['o2v1'])
```
mfherbst, line 78: That should also be a separate function further below. It feels strange to have this return early as a special case.

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,43 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+        self.cvs_matrix = AdcMatrix(method, ground_state)
+        self.intermediates = self.cvs_matrix.intermediates
+        self.dipole_operator = dipole_operator
+
+    def f1(self):
+        rho_ov = self.intermediates.cv_p_ov
+        rho_vv = self.intermediates.cv_p_vv
+        d_co = self.dipole_operator['o2o1']
+        d_cv = self.dipole_operator['o2v1']
+        return(
```
mfherbst, line 102: Space before "(".

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,43 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+            + d_cv
+            - einsum("Ib,ba->Ia", d_cv, rho_vv)
+            - einsum("Ij,ja->Ia", d_co, rho_ov)
+        )
+
+    def f2(self):
+        mp_t2_oovv = self.ground_state.t2("o1o1v1v1")
+        d_co = self.dipole_operator['o2o1']
+        return(
```
mfherbst, line 111: Space before "("

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,43 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+            + 0.5**0.5 * einsum("Ik,kjab->jIab", d_co, mp_t2_oovv)
+        )
+
+    def cvs_adc2_mtm(self):
```
mfherbst, line 115: I'd just have this as a free function.

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,43 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+                                                       ground_state,
+                                                       dipole_operator)
+            return cvs_mtm.cvs_adc2_mtm()
     mtm_cpp = libadcc.compute_modified_transition_moments(
         method.name, ground_state, dipole_operator
     )
     return AmplitudeVector(*mtm_cpp.to_tuple())
+
+
+class Adc2CVSModifiedTransitionMoments:
```
maxscheurer, line 90: How about:

```Python
def modified_transition_moments_cvs_adc2(gs_or_matrix, dipole_operator):
    # build AdcIntermediates from gs_or_matrix
    # compute f1
    f1 = ...
    # compute f2
    f2 = ...
    return AmplitudeVector(f1, f2)
```

adcc/modified_transition_moments.py
maxscheurer, line 83: Fail when other method requested:

```Python
else:
    raise NotImplementedError("...")
```

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,43 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+    def __init__(self, method, ground_state, dipole_operator):
+        self.ground_state = ground_state
+        self.cvs_matrix = AdcMatrix(method, ground_state)
+        self.intermediates = self.cvs_matrix.intermediates
```
mfherbst, line 94: If `gs_or_matrix` in the `compute_modified_transition_moments` is a matrix it should be used here in order to re-use the intermediates.

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,39 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+
+
+def modified_transition_moments_cvs(gs_or_matrix, dipole_operator,
+                                    method=None):
+    if isinstance(gs_or_matrix, AdcMatrix):
+        ground_state = gs_or_matrix.ground_state
+        intermediates = gs_or_matrix.intermediates
+    elif isinstance(gs_or_matrix, LazyMp):
+        ground_state = gs_or_matrix
+        intermediates = AdcMatrix(method, ground_state).intermediates
```
mfherbst, line 92: That's a bit of a quirk :smile:. Just construct the `AdcIntermediates` class, see https://github.com/adc-connect/adcc/blob/master/adcc/state_densities.py#L47.

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,39 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+
+    if method.name == "cvs-adc1":
+        return AmplitudeVector(dipole_operator['o2v1'])
```
mfherbst, line 95: I think I would do an `elif` for `"cvs-adc2"` and bail out with a `NotImplementedError` in an `else` block.

adcc/modified_transition_moments.py
```diff
@@ -72,10 +74,39 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+
+    mp_t2_oovv = ground_state.t2("o1o1v1v1")
+
+    f1 = (
+        + d_cv
+        - einsum("Ib,ba->Ia", d_cv, rho_vv)
+        - einsum("Ij,ja->Ia", d_co, rho_ov)
+    )
+
+    f2 = + 0.5**0.5 * einsum("Ik,kjab->jIab", d_co, mp_t2_oovv)
```
mfherbst, line 110: `from math import sqrt` and then `1 / sqrt(2)` is cleaner IMHO.

adcc/test_modified_transition_moments.py
```diff
@@ -32,7 +32,7 @@
 from pytest import approx
 
 # The methods to test
-methods = ["adc2"]
+methods = ["adc2", "cvs-adc1", "cvs-adc2"]
```
mfherbst, line 35: Unfortunately it's not qutie just that simple in this test, because you need to construct a cvs-ReferenceState as well. See https://github.com/adc-connect/adcc/blob/master/adcc/testdata/cache.py#L130 versus https://github.com/adc-connect/adcc/blob/master/adcc/testdata/cache.py#L122. See e.g. https://github.com/adc-connect/adcc/blob/805d2a40de30a1556fc08a88bc86c7bace9f590c/adcc/test_AdcMatrix.py#L42 how you can solve this.

adcc/modified_transition_moments.py
maxscheurer: Plus, I'd move the first `if`-statement for CVS-ADC(1) to the top and move the construction of intermediates etc. inside the `elif` block for CVS-ADC(2).

maxscheurer:
@Drrehn 
If you want to test `cvs-adc` as well, you could query
```Python
if 'd' in mtms[i].blocks:
    # add the doubles part as well
```
to avoid a lot of `if`-statements.

adcc/modified_transition_moments.py
Drrehn: I'm learnding

adcc/modified_transition_moments.py
```diff
@@ -76,37 +77,66 @@ def compute_modified_transition_moments(gs_or_matrix, dipole_operator,
...
+            + d_cv
+            - einsum("Ib,ba->Ia", d_cv, rho_vv)
+            - einsum("Ij,ja->Ia", d_co, rho_ov)
+        )
+
+        f2 = + 1.0 / sqrt(2) * einsum("Ik,kjab->jIab", d_co, mp_t2_oovv)
+        return AmplitudeVector(f1, f2)
+    else:
+        raise NotImplementedError("compute_modified_transition_moments "
+                                  "not implemented for", method.name)
```
mfherbst, line 121: I'd use a format string: `f"not implemented for {method.name}"`. That will automatically substitute the value of `method.name`.

adcc/modified_transition_moments.py
mfherbst: Sure, that's what reviews are for. I hope I don't come across as too harsh ... not my intention.

mfherbst:
@Drrehn I suppose this is ready to go in from your end?

@maxscheurer Feel free to squash-merge if you approve.

Drrehn:
yay!

mfherbst:
@Drrehn Do you need a new version?

maxscheurer:
I think we don't need to pin/release a new version at this point. Probably some other Gator-related stuff will need to be added this month?

mfherbst:
I don't mind waiting ...

PR #71, opened 2020-06-06
Implement ADC equations on the python side

mfherbst:
Still at a very early stage. Just to stash some work ...

Done in #77:
- [x] Update adccore to the version including the missing patches
- [x] diffdm
- [x] transition dms
- [x] MTMs
- [x] state-to-state transition dms will be done in #91 

Done in this PR:
- [x] Rebase to pick up on #77
- [x] Rebase to pick up #79 
- [x] Rebase to pick up #98
- [x] Rebase to pick up #91
- [x] Matrix block-wise applies
    - [x] ADC(0)
    - [x] ADC(1)
    - [x] ADC(2)
    - [x] ADC(2)-x
    - [x] ADC(3)
    - [x] CVS-ADC(0)
    - [x] CVS-ADC(1)
    - [x] CVS-ADC(2)
    - [x] CVS-ADC(2)-x
    - [x] CVS-ADC(3)
- [x] Timers on pyside routines
- [x] Compare timings
- [x] ~~Restructure adc expressions: Make the same function generate ph_pphh and pphh_ph blocks
      (because they should always be equivalent due to ADC matrix being Hermitian)~~ :arrow_right: Nah, causes issues with block-wise apply
- [x] Pass for some simple optimisations:
   - Store more (cheap) intermediates in hot applies
- [x] Pythonise AmplitudeVector
- [x] Pythonise AdcMatrix
- [x] Docstrings
- [x] Cleanup:
   * [x] Find deprecated usage of `blocks`, `"s"`, `"d"` etc.
   * [x] Outdated stuff in docs

Not so clear stuff:
- [x] Opinions on `hf.b.cv` versus `b.cv` versus `hf.b.Cv`  ??
- ~~[ ] Quick-access like functionality for T2 amplitudes and density matrices?~~ :arrow_right: postponed
- ~~[ ] Quick-access like functionality for other things of LazyMp?~~ :arrow_right: postponed

PR #76, opened 2020-06-16
Amend evaluate function

maxscheurer:
This is needed to achieve compatibility with stuff derived off tensors or amplitude vectors (e.g. a complex response vector)...

Probably some changes will also be made in the CG code if needed (I don't want to have an extra CG solver for complex response functions)

### DIFF
--- a/adcc/functions.py
+++ b/adcc/functions.py
@@ -150,6 +150,8 @@ def evaluate(a):
         return a
     elif isinstance(a, list):
         return [evaluate(elem) for elem in a]
+    elif hasattr(a, "evaluate"):
+        return a.evaluate()
     else:
         return libadcc.evaluate(a)
 


maxscheurer:
@mfherbst seems to work, also with the response vectors (that's why I can now use the CG from `adcc`). I don't know if there's some pitfall that we don't cover with our tests or if the change is acceptable...

mfherbst:
Actually ...
https://github.com/adc-connect/adcc/blob/75dd60b3bc0a5bba1e4f33168201a90b550de1b3/adcc/functions.py#L145-L152
... I just did the same thing in the other branch.

mfherbst:
Ups sorry ... I did not want to close.

mfherbst:
Ok. I think I'll split #71 and we merge some of it. There is no point in us repeating the work.

mfherbst:
I'll do this in #77 if you don't mind.

PR #77, opened 2020-06-16
Move MTMs, diffdms and transition dms to python

mfherbst:
- [x] Update adccore to the version including the missing patches
- [x] diffdm
- [x] transition dms
- [x] MTMs
- [x] Getattr in module file
- [x] as b more locally.
- [x] Check timings against master.


mfherbst:
@Drrehn @maxscheurer Now that I implemented all MTM equations in the same python file I an wondering whether the CVS `f2` contribution needs an `antisymmetrise` as well. Right here: https://github.com/adc-connect/adcc/blob/b10592e5dfeca87b8242b5eb73891ea3d27f7c32/adcc/kernel/modified_transition_moments.py#L73

Thoughts?

mfherbst:
Also we should bikeshed the name of the folder. From my end pro `kernel` is that it is an unused name in adcc (core is already connected to the C++ library). Cons is that it mathematically is something different. Both terms are also pretty much meaningless. Same as equations, which is too generic (and probably this folder will not be the only place with equations and this folder will not just be equations). 

I am tempted to leave it as `kernel` even though I don't like it. I am open for other suggestions.

mfherbst:
Also one thing we *could* do is not have so long names inside the kernel files. Since the files are already named like `modified_transition_moments` I think it's fine to name the functions just after the ADC method and drop the prefix. Thoughts?

maxscheurer:
I‘m giving this a thorough look first thing tomorrow morning!

maxscheurer:
- CVS `f2`: @Drrehn, please correct me if I'm wrong: IIRC the `antisymmetrise` is not needed because the symmetry only arises from the `ab` indices of the `t2`, which are retained during contraction.

- folder name: I'm against `kernel` and for `core`. The association with the adccore repo is just manifested by the name of the repo, not really in the Python code (`libadcc`). TBH I don't like either option... 

> I think it's fine to name the functions just after the ADC method and drop the prefix
Can you be more specific about this? I expect there to be clashes with duplicate names when the actual functions are being imported somewhere, i.e., 
```Python
from adcc.modified_transition_moments import adc2
from adcc.state_densities import adc2
```
Probably I misunderstand?

adcc/kernel/modified_transition_moments.py
```diff
@@ -0,0 +1,130 @@
...
+    "cvs-adc0": modified_transition_moments_cvs_adc0,
+    "cvs-adc1": modified_transition_moments_cvs_adc0,  # Identical to CVS-ADC(0)
+    "cvs-adc2": modified_transition_moments_cvs_adc2,
+}
+
+
+def modified_transition_moments(method, ground_state, dipole_operator=None,
+                                intermediates=None):
+    """Compute the modified transition moments (MTM) for the provided
+    ADC method with reference to the passed ground state.
```
maxscheurer, line 90: Just out of interest... is there a possibility to easily add LaTex equations to the documentation?
Then we could have $F_J(\op{d}) = \ISB{J}\op{d}\ket{\Psi_0}$ or something of the sort 😄

adcc/kernel/state_diffdm.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+# dict controlling the dispatch of the state_diffdm function
+DISPATCH = {
+    "adc0": state_diffdm_adc0,
+    "adc1": state_diffdm_adc0,  # same as ADC(0)
+    "adc2": state_diffdm_adc2,
+    "adc2x": state_diffdm_adc2,
+    "cvs-adc0": state_diffdm_adc0,
+    "cvs-adc1": state_diffdm_adc0,
```
maxscheurer, line 164: Comment `# same as CVS-ADC(0)` for consistency.

adcc/kernel/test_modified_transition_moments.py
```diff
@@ -39,39 +38,17 @@
...
-            if 'd' in mtms[i].blocks:
-                ref_d = ref["eigenvectors_doubles"][i]
-                mtm_np_d = [mtms[i]['d'].to_ndarray() for i in range(3)]
-
-                res_tdm -= np.array([
-                    np.sum(ref_d * mtm_d) for mtm_d in mtm_np_d
-                ])
-
+            excivec = state.excitation_vector[i]
+            res_tdm = -1.0 * np.array([excivec @ mtms[i] for i in range(3)])
```
maxscheurer, line 51: Nice! I don't know why I made this so complicated before 😄

extension/export_AdcMatrix.cc
```diff
@@ -63,13 +63,15 @@ void export_AdcMatrix(py::module& m) {
         .def_property_readonly("mospaces", &AdcMatrix::mospaces_ptr)
         .def_property_readonly("is_core_valence_separated",
                                &AdcMatrix::is_core_valence_separated)
-        .def("compute_apply", &AdcMatrix::compute_apply)
+        .def("compute_apply", &AdcMatrix::compute_apply, py::arg("block"), py::arg("in"),
```
maxscheurer, line 66: I think we should start using literal operators instead of `py::arg`, because it makes the code shorter:
```C++
.def("compute_apply", &AdcMatrix::compute_apply, "block"_a, "in"_a,
```

[Link](https://pybind11.readthedocs.io/en/stable/basics.html#keyword-arguments)

Note: needs a `using namespace pybind11::literals`.

extension/export_OneParticleOperator.cc
```diff
@@ -112,6 +114,8 @@ void export_OneParticleOperator(py::module& m) {
              "Set a block of the matrix (e.g. o1o1, o1v1)")
         .def("set_zero_block", set_zero_block_1,
              "Set a block of the matrix (e.g. o1o1, o1v1) to be explicitly zero.")
+        .def("set_zero_block", set_zero_block_2,
```
maxscheurer, line 117: I'd prefer `py::overload_cast` if possible.

```C++
.def("set_zero_block", py::overload_cast<std::vector<std::string>>(&OneParticleOperator::set_zero_block),
"Set a few blocks of the matrix (e.g. o1o1, o1v1) to be explicitly zero.")
```
Then it's all in one place.

[Link](https://pybind11.readthedocs.io/en/stable/classes.html?highlight=overload_cast#overloaded-methods)

Drrehn:
I agree with @maxscheurer that the antisymmetrise is not needed becaused the indices are not touched even though I don't know how to make my reply as fancy with bullet points

mfherbst:
@maxscheurer Thanks for the comments.

- folder name: I see your point. But keep in mind if we later distribute adccore separately as a conda / pip package than the name of the module will likely change to adccore as well for consistency. I never liked libadcc ... it was just a hack because adccore.so was already the statically compiled part, but if this has the python bindings included, there is no reason for libadcc. Of course we could also think of a different name for that library ... then I don't mind using core here.
- prefix: No you don't misunderstand and I see your point, but the idea is that the user will always use the dispatch function and never call into the specific methods. Unless of course he is doing low-level hacking and then using the full thing like `modified_transition_moments.adc2` is ok, I think. My point is that shorter names in the `modified_transition_moments.py` lead to shorter and simpler to digest code and thus allows you to better get an overview of the equations, which is what really matters here. and not the name. Also the use case of `from ... import adc2` is extremely rare.
- antisymmetrise: Ok then we leave it. It's just that for the amplitude vectors it *is* antisymmetrised and to me contracting MTMs with dips and amplitudes with TDMs was always kind of pendants, so I would expect by symmetry an antisymmetrisation to be needed in the MTMs.

@Drrehn https://guides.github.com/features/mastering-markdown/

adcc/block.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def __getattr__(attr):
+    """
+    Return a multi-dimensional block string like 'o1o2v1v1'
+    when requesting the attribute 'ocvv'.
+    """
+    if any(c not in ["o", "v", "c"] for c in attr):
+        raise AttributeError
+    mapping = {"o": "o1", "v": "v1", "c": "o2"}
```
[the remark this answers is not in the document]
mfherbst: Really ... but than it's three lines. Do you find the current version so confusing? It's not that there will be many more entries (maybe later for RI, but then it's also just one extra letter).

adcc/kernel/modified_transition_moments.py
```diff
@@ -0,0 +1,130 @@
...
+        - einsum("ik,kjab->ijab", dipop[b.oo], t2).antisymmetrise(0, 1)
+    )
+    return AmplitudeVector(f1, f2)
+
+
+def modified_transition_moments_cvs_adc0(mp, dipop, intermediates):
+    return AmplitudeVector(dipop[b.cv])
+
+
+def modified_transition_moments_cvs_adc2(mp, dipop, intermediates):
```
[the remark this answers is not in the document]
mfherbst: Yes that is an option. Also `tdm_bla` and `dm_bla` and so on. Sounds like a good compromise to me. I'll do that.

adcc/kernel/modified_transition_moments.py
mfherbst: Sure! It's processed by sphinx so you can just use RST like in https://raw.githubusercontent.com/adc-connect/adcc/master/docs/theory.rst. We have to be careful with this to not blow up the docstrings too much. I think a general equation in the dispatch function is fine, but I would be reluctant to document the method-specific implementations to save space.

extension/export_AdcMatrix.cc
mfherbst: Good call.

Drrehn:
> * antisymmetrise: Ok then we leave it. It's just that for the amplitude vectors it _is_ antisymmetrised and to me contracting MTMs with dips and amplitudes with TDMs was always kind of pendants, so I would expect by symmetry an antisymmetrisation to be needed in the MTMs.

I guess the idea is that they are already antisymmetrised in a and b from the t2-amplitude and i and j are have no symmetry because of core and valence

mfherbst:
Yes absolutely! That makes sense and that I buy.

extension/export_OneParticleOperator.cc
mfherbst: Ok, I'll try that. With some of the pybind features it tends to give very long lines, which destroys the overview a little (that's why I rarely use the lambdas), but this might work.

adcc/block.py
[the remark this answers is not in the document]
maxscheurer: Nah, it's okay 😄

adcc/kernel/modified_transition_moments.py
maxscheurer: Yeah, I was not talking about method-specific stuff. This would be insane. Just general equations, but I would not do it in this PR.

extension/export_OneParticleOperator.cc
maxscheurer: I guess if one breaks the line after the first argument, it should be okay.

adcc/kernel/modified_transition_moments.py
mfherbst: Agree.

PR #78, opened 2020-06-17
Pandas DataFrame export

maxscheurer:
This PR adds the feature to export `ExcitedStates` as a `pandas.DataFrame` for convenient post-processing and serialization.

- [x] basic functionality
- [x] export energy corrections as well
- [x] make excitation property keys accessible for everyone
- [x] add `pandas` to `setup.py` and `conda` requirements
- [x] small test to keep coverage high 😄 

### Questions:
- Any other properties that should be exported?


adcc/ExcitedStates.py
```diff
@@ -657,6 +659,48 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+        """
+        propkeys = extract_excitation_property_keys(self)
+        propkeys.extend(self.excitation_energy_corrections.keys())
+        data = {
+            "excitation": np.arange(0, self.size, dtype=int),
+            "kind": np.tile(self.kind, self.size)
+        }
+        # fill the dict with data, sort out stuff that we cannot represent
+        # in a table
+        from adcc import Tensor, AmplitudeVector, OneParticleOperator
```
mfherbst, line 680: Not a fan of this. Why not just `import adcc` at the top (or sth like that) and then `adcc.Tensor` etc. ?

adcc/ExcitedStates.py
```diff
@@ -657,6 +659,48 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+                if any(type(e) in unacceptable_types for e in d):
+                    continue
+                data[key] = d
+            elif isinstance(d, np.ndarray):
+                if d.ndim == 1:
+                    data[key] = d
+                elif d.ndim == 2 and d.shape[1] == 3 and export_vectors:
+                    component = ["x", "y", "z"]
+                    # TODO: ugly, but works...
+                    for i, p in enumerate(component):
```
mfherbst, line 693: Completely fine, I think. I'd even do it in one line.

adcc/ExcitedStates.py
```diff
@@ -657,6 +659,48 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
                 ret += separator
         return ret[:-1]
 
+    def to_dataframe(self, export_vectors=True):
```
mfherbst, line 662: Can you think of any case where you would want `export_vectors=False`?

adcc/Excitation.py
```diff
@@ -42,6 +42,31 @@ def inner(f, kwargs=kwargs):
     return inner
 
 
+def extract_excitation_property_keys(state):
```
mfherbst, line 45: I get the idea, but it feels a bit complicated to me. Sure there is no simpler way (like using the decorator to automatically build a list of properties or sth like that).

adcc/ExcitedStates.py
maxscheurer: Yeah, that's nicer!

adcc/ExcitedStates.py
maxscheurer: Say you're using a notebook and want to display the dataframe there... you might not want a df that runs too wide...

adcc/Excitation.py
maxscheurer: Then this would need to be done inside `ExcitedStates` itself. I can certainly do this.

adcc/ExcitedStates.py
mfherbst: Yes but then isn't it just a question of ordering the important columns first?

adcc/Excitation.py
mfherbst: Yes I agree! I have no clue what is better. Just try a bit. The problem with python is that the enormous flexibility sometimes lets you buy into too much metaprogramming ...

adcc/ExcitedStates.py
maxscheurer: Well... what is "important"? 😄 Might be different for everyone...
I'd say we drop the extra flag, such that you always get everything. you can then of course reduce the table on your own if you want to.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/Excitation.py
maxscheurer: Done. I like the newer version much better.

maxscheurer: I think it lives in the right spot now 😄

adcc/ExcitedStates.py
```diff
@@ -657,6 +659,66 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
                 ret += separator
         return ret[:-1]
 
+    def to_dataframe(self):
+        """
+        Exports the ExcitedStates object as :class:`pandas.DataFrame`.
+        All values are printed in atomic units.
```
mfherbst, line 665: printed? I'd just say atomic units are used for all values.

adcc/ExcitedStates.py
```diff
@@ -657,6 +659,66 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+        """
+        propkeys = self.excitation_property_keys.copy()
```
mfherbst, line 667: Why copy? a property returns a new object on every call (unless cached :smile:)

adcc/ExcitedStates.py
```diff
@@ -657,6 +659,66 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+        propkeys.extend(self.excitation_energy_corrections.keys())
+        data = {
+            "excitation": np.arange(0, self.size, dtype=int),
+            "kind": np.tile(self.kind, self.size)
+        }
+        # fill the dict with data, sort out stuff that we cannot represent
+        # in a table
+        unacceptable_types = [adcc.AmplitudeVector, adcc.OneParticleOperator,
+                              adcc.Tensor]
```
mfherbst, line 676: In these kind of situations there is always the question of blacklisting versus whitelisting. Generally I think whitelisting is better here, because it is more forward-compatible (you don't want this to break if you implement a new fancy property ... better it just does not show up). So I'd vote for whitelisting numbers and np arrays and lists and the few other things (if any). If you use proper subtypes this is not too many tests (numpy has base types for all numbers I think).

adcc/ExcitedStates.py
```diff
@@ -657,6 +659,66 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+                continue
+            if isinstance(d, list):
+                if not any(type(e) in unacceptable_types for e in d):
+                    data[key] = d
+            elif isinstance(d, np.ndarray):
+                if d.ndim == 1:
+                    data[key] = d
+                elif d.ndim == 2 and d.shape[1] == 3:
+                    for i, p in enumerate(["x", "y", "z"]):
+                        data[f"{key}_{p}"] = d[:, i]
```
mfherbst, line 691: This quite deeply nested. I don't have a good idea how to not do it, but maybe when you refactor you have one.

adcc/ExcitedStates.py
```diff
@@ -657,6 +659,66 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+                elif d.ndim > 2:
+                    raise TypeError(f"Exporting NumPy array with shape {d.shape}"
+                                    " not supported.")
+            else:
+                data[key] = d
```
mfherbst, line 696: This is easily overlooked. I would make it the first case.

mfherbst:
Regarding other properties: I think dominant amplitudes would be awesome, but that could be a bit of effort to properly map this. I'd rather have it as is than nothing, so your call if you want to do it or not.

adcc/ExcitedStates.py
maxscheurer: Probably we only need ndarrays... we don‘t really use lists for numeric types and I don’t think it should be encouraged :smile:
So I‘d say we only export numpy ndarrays which are subtypes of np.numeric. Currently I cannot imagine anything one might „compute“ and mark as excitation property that would be a string...

maxscheurer:
> Regarding other properties: I think dominant amplitudes would be awesome

One could do this, but probably as a separate dataframe. The format of amplitudes does not fit the concise layout of the currently exported dataframe. I have something in mind that I could test tomorrow.

adcc/ExcitedStates.py
mfherbst: I agree should be as specific as sensible. It's just the first few things which came to my mind, so I wrote them down, but just all subtypes of np.numeric sounds exactly what we need (just make sure this includes standard python ints and such).

mfherbst:
Awesome! Anything in this direction would be appreciated. If you can somehow get a nice overview of this that would just be great.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: That's correct 😄 Done.

adcc/ExcitedStates.py
maxscheurer: It's much simpler and easier to read now.

adcc/ExcitedStates.py
maxscheurer: Not as deeply nested in the new version 👍

adcc/ExcitedStates.py
maxscheurer: Obsolete now...

adcc/ExcitedStates.py
```diff
@@ -657,6 +658,60 @@ def describe_amplitudes(self, tolerance=0.01, index_format=None):
...
+            if not np.issubdtype(d.dtype, np.number):
+                continue
+            if d.ndim == 1:
+                data[key] = d
+            elif d.ndim == 2 and d.shape[1] == 3:
+                for i, p in enumerate(["x", "y", "z"]):
+                    data[f"{key}_{p}"] = d[:, i]
+            elif d.ndim > 2:
+                raise TypeError(f"Exporting NumPy array with shape {d.shape}"
+                                " not supported.")
```
mfherbst, line 689: It just occurred to me that this should maybe be a warning followed by a continue, since the same reasoning of forward-compatibility applies. Also I'd add the `key` to the error message.

adcc/ExcitedStates.py
mfherbst: I agree :+1:

adcc/ExcitedStates.py
maxscheurer: Yeah that's nice!

mfherbst:
About other properties: Absorption cross section? Also excitation is an index (0 to something) and the index of the dataframe itself is also 0 to something. I don't think I would keep excitation or did you have something particular in mind when you did that?

maxscheurer:
> Absorption cross section?

Done 😄 ... and I noticed that it was only accessible via plots before haha.

> I don't think I would keep excitation or did you have something particular in mind when you did that?

Yes, I actually did! Say you do several calculations and merge dataframes with `pd.concat` or something of the sort with `ignore_index=True`, then you loose information about the number of the excited states. Having the excitation as an extra column thus comes in handy when comparing multiple calculations, because you can more easily filter and process the data


__Update__ on amplitudes: I think right now I'd have to change stuff in `FormatExcitationVector` to access the "raw" data...

mfherbst:
> Say you do several calculations and merge dataframes
Ah good point. That I did not think about ...

Re amplitudes ... don't worry if it turns out to be too annoying.

adcc/ExcitedStates.py
```diff
@@ -397,6 +399,15 @@ def state_dipole_moment(self):
             for ddm in self.state_diffdm
         ])
 
+    @property
+    @mark_excitation_property()
+    def cross_section(self):
```
mfherbst, line 404: Maybe a short docstring?

maxscheurer:
Okay, so the amplitudes are too much fiddling... I'm going to squash-merge this when the CI is done. I already have a new branch for the amplitude dataframe stuff.

maxscheurer:
I guess #79 needs to go in to fix the checksum errors?

PR #83, opened 2020-07-15
Add LGTM

maxscheurer:
Add LGTM Badge and automatic PR review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 # adcc: Seamlessly connect your program to ADC
 | **Documentation**                                       | **Build Status**                                       |  **Installation**                                                                      |
 |:------------------------------------------------------- |:------------------------------------------------------ |:-------------------------------------------------------------------------------------- |
-| [![][docs-img]][docs-url] [![][binder-img]][binder-url] | [![][travis-img]][travis-url] [![][cov-img]][cov-url]  | [![][pypi-img]][pypi-url] [![][conda-img]][conda-url] [![][license-img]][license-url]  |
+| [![][docs-img]][docs-url] [![][binder-img]][binder-url] | [![][travis-img]][travis-url] [![][cov-img]][cov-url] [![Language grade: Python][lgtm-img]][lgtm-url]  | [![][pypi-img]][pypi-url] [![][conda-img]][conda-url] [![][license-img]][license-url]  |
 
 [docs-img]: https://img.shields.io/badge/doc-latest-blue.svg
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3,7 +3,7 @@
 # adcc: Seamlessly connect your program to ADC
 | **Documentation**                                       | **Build Status**                                       |  **Installation**                                                                      |
 |:------------------------------------------------------- |:------------------------------------------------------ |:-------------------------------------------------------------------------------------- |
-| [![][docs-img]][docs-url] [![][binder-img]][binder-url] | [![][travis-img]][travis-url] [![][cov-img]][cov-url]  | [![][pypi-img]][pypi-url] [![][conda-img]][conda-url] [![][license-img]][license-url]  |
+| [![][docs-img]][docs-url] [![][binder-img]][binder-url] | [![][travis-img]][travis-url] [![][cov-img]][cov-url] [![Language grade: Python][lgtm-img]][lgtm-url]  | [![][pypi-img]][pypi-url] [![][conda-img]][conda-url] [![][license-img]][license-url]  |
```
mfherbst, line 6: I would not put this badge in the table. It's not so important and it makes the table two-row, which is much harder to read. I'd add it somewhere else in the readme.

maxscheurer:
How about we change the table to this format:

Screen Shot 2020-07-15 at 10 53 08

mfherbst:
Actually I dislike that too. I think psi4 is overdoing it a little with the badges. I'd just add a section in the text (like the citations) about code quality or code status or sth like that?

maxscheurer:
I didn't mean that we should add all these badges 😄 But instead of having rows instead of columns for Documentation, Build Status and Installation. Then the code quality thing can be added to build status, where it fits best I guess. But if you like the extra section better, we can do that too 😉

mfherbst:
Ah I see. Hmm I don't mind rows. Let's see what it looks like.

mfherbst:
Much better. Now the last thing that bothers me is that the LGTM badge is quite wide. Since we don't really have any other language at the moment, can we just drop the logo and the language in the badge?

maxscheurer:
>  can we just drop the logo and the language in the badge?

I've dropped the logo, but the language cannot be removed unfortunately. But now it's already less wide...

mfherbst:
> language cannot be removed unfortunately

maxscheurer:
> Nobody stops you from making a custom shield.io logo

Done 😄 The label can be overridden...

PR #90, opened 2020-08-24
Fix error in Docs

maxscheurer:
Outdated name `state.eigenvectors` renamed to `state.excitation_vector` (reported by Marvin, thanks!)


mfherbst (approved):
Thanks @maxscheurer and thanks Marvin!

PR #91, opened 2020-09-09
State-to-state transition densities

maxscheurer:
Python-side equations for State-to-state transition densities.

- [x] canonical ADC
- ~~[ ] CVS-ADC (I'm not really in the mood to add this though 😄)~~ ➡️ postpone
- [x] figure out order stuff
- [x] equation cleanup
- [x] cleanup: add checks for vector spaces etc.
- [x] expose via `ExcitedStates`/`Excitation` (this is the hardest part) (**preliminary** version working)
- [x] add tests against reference data
    - [x] transition dipole moments
    - [x] transition densities
- [x] run full tests
- [x] squash

### DIFF
--- a/adcc/ElectronicTransition.py
+++ b/adcc/ElectronicTransition.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python3
+## vi: tabstop=4 shiftwidth=4 softtabstop=4 expandtab
+## ---------------------------------------------------------------------
+##
+## Copyright (C) 2020 by the adcc authors
+##
+## This file is part of adcc.
+##
+## adcc is free software: you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published
+## by the Free Software Foundation, either version 3 of the License, or
+## (at your option) any later version.
+##
+## adcc is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with adcc. If not, see <http://www.gnu.org/licenses/>.
+##
+## ---------------------------------------------------------------------
+import warnings
+import numpy as np
+
+from .misc import cached_property
+from .timings import timed_member_call
+from .visualisation import ExcitationSpectrum
+from .OneParticleOperator import product_trace
+
+from scipy import constants
+from matplotlib import pyplot as plt
+from .Excitation import mark_excitation_property
+
+
+class ElectronicTransition:
+    """
+    Documentation
+    """
+    @cached_property
+    @mark_excitation_property()
+    @timed_member_call(timer="_property_timer")
+    def transition_dipole_moment(self):
+        """List of transition dipole moments of all computed states"""
+        if self.property_method.level == 0:
+            warnings.warn("ADC(0) transition dipole moments are known to be "
+                          "faulty in some cases.")
+        dipole_integrals = self.operators.electric_dipole
+        return np.array([
+            [product_trace(comp, tdm) for comp in dipole_integrals]
+            for tdm in self.transition_dm
+        ])
+
+    @cached_property
+    @mark_excitation_property()
+    @timed_member_call(timer="_property_timer")
+    def transition_dipole_moment_velocity(self):
+        """List of transition dipole moments in the
+        velocity gauge of all computed states"""
+        if self.property_method.level == 0:
+            warnings.warn("ADC(0) transition velocity dipole moments "
+                          "are known to be faulty in some cases.")
+        dipole_integrals = self.operators.nabla
+        return np.array([
+            [product_trace(comp, tdm) for comp in dipole_integrals]
+            for tdm in self.transition_dm
+        ])
+
+    @cached_property
+    @mark_excitation_property()
+    @timed_member_call(timer="_property_timer")
+    def transition_magnetic_dipole_moment(self):
+        """List of transition magnetic dipole moments of all computed states"""
+        if self.property_method.level == 0:
+            warnings.warn("ADC(0) transition magnetic dipole moments "
+                          "are known to be faulty in some cases.")
+        mag_dipole_integrals = self.operators.magnetic_dipole
+        return np.array([
+            [product_trace(comp, tdm) for comp in mag_dipole_integrals]
+            for tdm in self.transition_dm
+        ])
+
+    @cached_property
+    @mark_excitation_property()
+    def oscillator_strength(self):
+        """List of oscillator strengths of all computed states"""
+        return 2. / 3. * np.array([
+            np.linalg.norm(tdm)**2 * np.abs(ev)
+            for tdm, ev in zip(self.transition_dipole_moment,
+                               self.excitation_energy)
+        ])
+
+    @cached_property
+    @mark_excitation_property()
+    def oscillator_strength_velocity(self):
+        """List of oscillator strengths in
+        velocity gauge of all computed states"""
+        return 2. / 3. * np.array([
+            np.linalg.norm(tdm)**2 / np.abs(ev)
+            for tdm, ev in zip(self.transition_dipole_moment_velocity,
+                               self.excitation_energy)
+        ])
+
+    @cached_property
+    @mark_excitation_property()
+    def rotatory_strength(self):
+        """List of rotatory strengths of all computed states"""
+        return np.array([
+            np.dot(tdm, magmom) / ee
+            for tdm, magmom, ee in zip(self.transition_dipole_moment_velocity,
+                                       self.transition_magnetic_dipole_moment,
+                                       self.excitation_energy)
+        ])
+
+    @property
+    @mark_excitation_property()
+    def cross_section(self):
+        """List of one-photon absorption cross sections of all computed states"""
+        # TODO Source?
+        fine_structure = constants.fine_structure
+        fine_structure_au = 1 / fine_structure
+        prefac = 2.0 * np.pi ** 2 / fine_structure_au
+        return prefac * self.oscillator_strength
+
+    def plot_spectrum(self, broadening="lorentzian", xaxis="eV",
+                      yaxis="cross_section", width=0.01, **kwargs):
+        """One-shot plotting function for the spectrum generated by all states
+        known to this class.
+
+        Makes use of the :class:`adcc.visualisation.ExcitationSpectrum` class
+        in order to generate and format the spectrum to be plotted, using
+        many sensible defaults.
+
+        Parameters
+        ----------
+        broadening : str or None or callable, optional
+            The broadening type to used for the computed excitations.
+            A value of None disables broadening any other value is passed
+            straight to
+            :func:`adcc.visualisation.ExcitationSpectrum.broaden_lines`.
+        xaxis : str
+            Energy unit to be used on the x-Axis. Options:
+            ["eV", "au", "nm", "cm-1"]
+        yaxis : str
+            Quantity to plot on the y-Axis. Options are "cross_section",
+            "osc_strength", "dipole" (plots norm of transition dipole),
+            "rotational_strength" (ECD spectrum with rotational strength)
+        width : float, optional
+            Gaussian broadening standard deviation or Lorentzian broadening
+            gamma parameter. The value should be given in atomic units
+            and will be converted to the unit of the energy axis.
+        """
+        if xaxis == "eV":
+            eV = constants.value("Hartree energy in eV")
+            energies = self.excitation_energy * eV
+            width = width * eV
+            xlabel = "Energy (eV)"
+        elif xaxis in ["au", "Hartree", "a.u."]:
+            energies = self.excitation_energy
+            xlabel = "Energy (au)"
+        elif xaxis == "nm":
+            hc = constants.h * constants.c
+            Eh = constants.value("Hartree energy")
+            energies = hc / (self.excitation_energy * Eh) * 1e9
+            xlabel = "Wavelength (nm)"
+            if broadening is not None and not callable(broadening):
+                raise ValueError("xaxis=nm and broadening enabled is "
+                                 "not supported.")
+        elif xaxis in ["cm-1", "cm^-1", "cm^{-1}"]:
+            towvn = constants.value("hartree-inverse meter relationship") / 100
+            energies = self.excitation_energy * towvn
+            width = width * towvn
+            xlabel = "Wavenumbers (cm^{-1})"
+        else:
+            raise ValueError("Unknown xaxis specifier: {}".format(xaxis))
+
+        if yaxis in ["osc", "osc_strength", "oscillator_strength", "f"]:
+            absorption = self.oscillator_strength
+            ylabel = "Oscillator strengths (au)"
+        elif yaxis in ["dipole", "dipole_norm", "μ"]:
+            absorption = np.linalg.norm(self.transition_dipole_moment, axis=1)
+            ylabel = "Modulus of transition dipole (au)"
+        elif yaxis in ["cross_section", "σ"]:
+            absorption = self.cross_section
+            ylabel = "Cross section (au)"
+        elif yaxis in ["rot", "rotational_strength", "rotatory_strength"]:
+            absorption = self.rotatory_strength
+            ylabel = "Rotatory strength (au)"
+        else:
+            raise ValueError("Unknown yaxis specifier: {}".format(yaxis))
+
+        sp = ExcitationSpectrum(energies, absorption)
+        sp.xlabel = xlabel
+        sp.ylabel = ylabel
+        if not broadening:
+            plots = sp.plot(style="discrete", **kwargs)
+        else:
+            kwdisc = kwargs.copy()
+            kwdisc.pop("label", "")
+            plots = sp.plot(style="discrete", **kwdisc)
+
+            kwargs.pop("color", "")
+            sp_broad = sp.broaden_lines(width, shape=broadening)
+            plots.extend(sp_broad.plot(color=plots[0].get_color(),
+                                       style="continuous", **kwargs))
+
+        if xaxis in ["nm"]:
+            # Invert x axis
+            plt.xlim(plt.xlim()[::-1])
+        return plots
--- a/adcc/ElectronicTransition.py
+++ b/adcc/ElectronicTransition.py
@@ -24,19 +24,135 @@ import warnings
 import numpy as np
 
 from .misc import cached_property
-from .timings import timed_member_call
+from .timings import Timer, timed_member_call
 from .visualisation import ExcitationSpectrum
 from .OneParticleOperator import product_trace
+from .AdcMethod import AdcMethod
 
 from scipy import constants
 from matplotlib import pyplot as plt
 from .Excitation import mark_excitation_property
+from .solver.SolverStateBase import EigenSolverStateBase
 
 
 class ElectronicTransition:
-    """
-    Documentation
-    """
+    def __init__(self, data, method=None, property_method=None):
+        """Construct an ElectronicTransition class from some data obtained
+        from an interative solver or other suitable data.
+
+        Parameters
+        ----------
+        data
+            Any kind of iterative solver state. Typically derived off
+            a :class:`solver.EigenSolverStateBase`.
+        method : str, optional
+            Provide an explicit method parameter if data contains none.
+        property_method : str, optional
+            Provide an explicit method for property calculations to
+            override the automatic selection.
+        excitation_energy_corrections : dict, optional
+            Provide a dictionary of functions to compute corrections for
+            excitation energies, called for each excitation individually
+        """
+        self.matrix = data.matrix
+        self.ground_state = self.matrix.ground_state
+        self.reference_state = self.matrix.ground_state.reference_state
+        self.operators = self.reference_state.operators
+
+        # List of all the objects which have timers (do not yet collect
+        # timers, since new times might be added implicitly at a later point)
+        self._property_timer = Timer()
+        self._timed_objects = [("", self.reference_state),
+                               ("adcmatrix", self.matrix),
+                               ("mp", self.ground_state),
+                               ("intermediates", self.matrix.intermediates)]
+        if hasattr(data, "timer"):
+            datakey = getattr(data, "algorithm", data.__class__.__name__)
+            self._timed_objects.append((datakey, data))
+
+        # Copy some optional attributes
+        for optattr in ["converged", "spin_change", "kind", "n_iter"]:
+            if hasattr(data, optattr):
+                setattr(self, optattr, getattr(data, optattr))
+
+        self.method = getattr(data, "method", method)
+        if self.method is None:
+            self.method = self.matrix.method
+        if not isinstance(self.method, AdcMethod):
+            self.method = AdcMethod(self.method)
+        if property_method is None:
+            if self.method.level < 3:
+                property_method = self.method
+            else:
+                # Auto-select ADC(2) properties for ADC(3) calc
+                property_method = self.method.at_level(2)
+        elif not isinstance(property_method, AdcMethod):
+            property_method = AdcMethod(property_method)
+        self._property_method = property_method
+
+        # Special stuff for special solvers
+        if isinstance(data, EigenSolverStateBase):
+            self._excitation_vector = data.eigenvectors
+            self._excitation_energy_uncorrected = data.eigenvalues
+        else:
+            if hasattr(data, "eigenvalues"):
+                self._excitation_energy_uncorrected = data.eigenvalues
+            if hasattr(data, "eigenvectors"):
+                self._excitation_vector = data.eigenvectors
+            # if both excitation_energy and excitation_energy_uncorrected
+            # are present, the latter one has priority
+            if hasattr(data, "excitation_energy"):
+                self._excitation_energy_uncorrected = \
+                    data.excitation_energy.copy()
+            if hasattr(data, "excitation_energy_uncorrected"):
+                self._excitation_energy_uncorrected =\
+                    data.excitation_energy_uncorrected.copy()
+            if hasattr(data, "excitation_vector"):
+                self._excitation_vector = data.excitation_vector
+
+        # Collect all excitation energy corrections
+        self._excitation_energy = self._excitation_energy_uncorrected.copy()
+
+    def __len__(self):
+        return self.size
+
+    @property
+    def size(self):
+        return self._excitation_energy.size
+
+    @property
+    def timer(self):
+        """Return a cumulative timer collecting timings from the calculation"""
+        ret = Timer()
+        for key, obj in self._timed_objects:
+            ret.attach(obj.timer, subtree=key)
+        ret.attach(self._property_timer, subtree="properties")
+        ret.time_construction = self.reference_state.timer.time_construction
+        return ret
+
+    @property
+    def property_method(self):
+        """The method used to evaluate ADC properties"""
+        return self._property_method
+
+    @property
+    @mark_excitation_property()
+    def excitation_energy(self):
+        """Excitation energies including all corrections in atomic units"""
+        return self._excitation_energy
+
+    @property
+    @mark_excitation_property()
+    def excitation_energy_uncorrected(self):
+        """Excitation energies without any corrections in atomic units"""
+        return self._excitation_energy_uncorrected
+
+    @property
+    @mark_excitation_property()
+    def excitation_vector(self):
+        """List of excitation vectors"""
+        return self._excitation_vector
+
     @cached_property
     @mark_excitation_property()
     @timed_member_call(timer="_property_timer")
--- a/adcc/ElectronicTransition.py
+++ b/adcc/ElectronicTransition.py
@@ -50,9 +50,6 @@ class ElectronicTransition:
         property_method : str, optional
             Provide an explicit method for property calculations to
             override the automatic selection.
-        excitation_energy_corrections : dict, optional
-            Provide a dictionary of functions to compute corrections for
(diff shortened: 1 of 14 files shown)


adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,119 @@
...
+    )
+    return dm
+
+
+# Ref: https://doi.org/10.1080/00268976.2013.859313
+DISPATCH = {"adc0": s2s_tdm_adc0,
+            "adc1": s2s_tdm_adc0,       # same as ADC(0)
+            "adc2": s2s_tdm_adc2,
+            "adc2x": s2s_tdm_adc2,
+            "adc3": s2s_tdm_adc2,
```
mfherbst, line 134: Really? There are no terms in ADC(3)?

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,119 @@
...
+
+    rx1 = einsum('ijab,jb->ia', t2, x1)
+    ry1 = einsum('ijab,jb->ia', t2, y1)
+
+    dm[b.oo] = (
+        p1_oo - 2.0 * einsum('ikab,jkab->ij', y2, x2) + (
+            + 0.5 * (
+                einsum('ik,kj->ij', p1_oo, p0_oo)
+                + einsum('ik,kj->ij', p0_oo, p1_oo)
+            )
```
mfherbst, line 65: This (and in other places) is an explicit symmetrise. Any reason you do it this way?

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,119 @@
...
+    dm[b.oo] = -einsum('ja,ia->ij', x1, y1)
+    dm[b.vv] = einsum('ia,ib->ab', x1, y1)
+    return dm
+
+
+def s2s_tdm_adc2(mp, amplitude_l, amplitude_r, intermediates):
+    dm = s2s_tdm_adc0(mp, amplitude_l, amplitude_r, intermediates)
+
+    x1 = amplitude_l["s"]
+    y1 = amplitude_r["s"]
```
mfherbst, line 46: Not sure about x and y. The rest of the code uses u1 and u2. Why not stick to u1 and x1 maybe?

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,119 @@
...
+
+    t2 = mp.t2(b.oovv)
+    p0_ov = mp.mp2_diffdm[b.ov]
+    p0_oo = mp.mp2_diffdm[b.oo]
+    p0_vv = mp.mp2_diffdm[b.vv]
+    p1_oo = dm[b.oo].evaluate()  # ADC(1) tdm
+    p1_vv = dm[b.vv].evaluate()  # ADC(1) tdm
+
+    rx1 = einsum('ijab,jb->ia', t2, x1)
+    ry1 = einsum('ijab,jb->ia', t2, y1)
```
mfherbst, line 58: Add TODOs similar to state_diffdm

adcc/ElectronicTransition.py
```diff
@@ -0,0 +1,269 @@
...
+            ]
+        else:
+            final_vec = self.parent_state.excitation_vector[self.final]
+            return [
+                adc_pp.state2state_transition_dm(self.property_method,
+                                                 self.ground_state,
+                                                 self.initial_excitation_vector,
+                                                 final_vec,
+                                                 self.matrix.intermediates)
+            ]
```
mfherbst, line 269: Move to a separate file.

adcc/ElectronicTransition.py
```diff
@@ -0,0 +1,269 @@
...
+    def __init__(self, parent_state, initial=0, final=None):
+        self.parent_state = parent_state
+        self.reference_state = self.parent_state.reference_state
+        self.ground_state = self.parent_state.ground_state
+        self.matrix = self.parent_state.matrix
+        self.property_method = self.parent_state.property_method
+        self.operators = self.parent_state.operators
+
+        self.initial = initial
+        self.final = final
```
mfherbst, line 227: Why is there a final? We don't have this in ExcitedStates either?

adcc/ElectronicTransition.py
```diff
@@ -0,0 +1,269 @@
...
+            plots.extend(sp_broad.plot(color=plots[0].get_color(),
+                                       style="continuous", **kwargs))
+
+        if xaxis in ["nm"]:
+            # Invert x axis
+            plt.xlim(plt.xlim()[::-1])
+        return plots
+
+
+class State2StateTransition(ElectronicTransition):
```
mfherbst, line 214: I'd shorten it to State2States (similar to ExcitedStates)

adcc/ElectronicTransition.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    Documentation
+    """
+    def __init__(self, parent_state, initial=0, final=None):
```
mfherbst, line 218: I'm not a full fan of the wrapping aspect. I would try to drop storing the parent_state in this guy and also try to make it possible to construct it more low-level (e.g. without passing a parent state, but rather all the properties like the property_method etc). Gives more flexibility and one can always make a higher-level wrapper later.

adcc/ElectronicTransition.py
```diff
@@ -0,0 +1,269 @@
...
+        """
+        # TODO: only states above self.initial
+        if self.final is None:
+            return [
+                adc_pp.state2state_transition_dm(self.property_method,
+                                                 self.ground_state,
+                                                 self.initial_excitation_vector,
+                                                 evec,
+                                                 self.matrix.intermediates)
+                for i, evec in enumerate(self.parent_state.excitation_vector)
```
mfherbst, line 258: That's exactly the part I hate about black. I think that is too many lines for such a call. Can you try to compress that?

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,119 @@
...
+from adcc import block as b
 from adcc.AdcMethod import AdcMethod
+from adcc.functions import einsum
 from adcc.AmplitudeVector import AmplitudeVector
+from adcc.OneParticleOperator import OneParticleOperator
 
 import libadcc
 
-DISPATCH = {}  # None implemented
+
```
mfherbst, line 31: Thanks for extending that (and no probs it does not yet have CVS). In general I think you are too much copying libadc, i.e. I feel these equations could be compressed better by using dummy temporary variables or more explicit symmetrise etc. Remember since our evaluation is lazy temporary assignments don't cost because they don't evaluate.

adcc/test_state_densities.py
```diff
@@ -146,3 +148,31 @@ def base_test(self, system, method, kind):
...
+        refdata = cache.reference_data[system]
+        state = cache.adc_states[system][method][kind]
+
+        state_to_state = refdata[method][kind]["state_to_state"]
+        refevals = refdata[method][kind]["eigenvalues"]
+        for i, exci in enumerate(state.excitations):
+            # Check that we are talking about the same state when
+            # comparing reference and computed
+            assert exci.excitation_energy == refevals[i]
+            # TODO: I (MS) have no idea how the densities in the ctx are sorted...
```
mfherbst, line 167: Ok. I'll take a look ...

mfherbst (commented):
I took a first quick look. Here are my comments.

adcc/ElectronicTransition.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @cached_property
+    @mark_excitation_property(transform_to_ao=True)
+    @timed_member_call(timer="_property_timer")
+    def transition_dm(self):
+        """
+        List of transition density matrices from
+        initial state to final state/s
+        """
+        # TODO: only states above self.initial
```
mfherbst, line 250: Yes. Otherwise I think the interface becomes very strange.

mfherbst: Ok this interface is very confusing (at least for me). This initial final business we need to talk about at some point.

adcc/ElectronicTransition.py
maxscheurer: Because one needs some reference for the "from"-state and the "to/final"-state...

adcc/ElectronicTransition.py
maxscheurer: I think the main problem is:

Do we allow for only one or multiple final states in State2States? Or both?

From my point of view, we could make it lazy, such that you can either get all the transitions from one initial state to all the others...

mfherbst:
There is a bug in the testdata generator where from and to are flipped over, see

https://github.com/adc-connect/adcc-testdata/blob/master/adcctestdata/dump_reference.py#L206

but even taking that into account into the tests only ADC(0) and ADC(1) work. Could you have made a small mistake at ADC(2) level, which magically compensates in the dipoles?

adcc/test_properties.py
```diff
@@ -81,3 +83,27 @@ def base_test(self, system, method, kind):
...
+        refevals = refdata[method][kind]["eigenvalues"]
+        for i, exci in enumerate(state.excitations):
+            assert exci.excitation_energy == refevals[i]
+            fromi_ref = state_to_state[f"from_{i}"]["transition_dipole_moments"]
+            for ii, j in enumerate(range(i + 1, state.size)):
+                assert state.excitation_energy[j] == refevals[j]
+
+                # TODO Note the reversal
+                #      ... this is a bug in the testdata generator!
+                state2state = State2StateTransition(state, initial=j, final=i)
```
mfherbst, line 107: Not too happy about the way you structure the tests here. Ideally I would construct only *one* `State2StateTransition` for each `i` and use that to test all `i -> j` tdms.

maxscheurer: Yes, I agree.

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Just tried to remove the explicit symmetrise, but got failing tests even when using the `symmetrise` function on the result 😭

mfherbst: No what I meant is replace the explicitly coded `symmetrise` by an appropriate call to `symmetrise` (to save lines of code).

adcc/ElectronicTransition.py
mfherbst: Really? Can't you just do "all" or is that too expensive?

adcc/ElectronicTransition.py
mfherbst: That's a good question. In my head it was one initial to all final (kind of as a pendant to ExcitedStates). But you probably have more insight into what is needed. I think it gets pretty tricky in the interfaces to support both at the same time. What one could do is have one class that does 1 inital -> all final and one that does the reverse (all initial -> 1 final). Or one that does truely everything (all -> all) and then some clever iterators or arrangement of the properties in multi-dimensional np arrays to make it easy to use the results. But I don't know how expensive that becomes. Do you have an idea about that?

maxscheurer: I think 1 initial -> all final is the best solution. Then it's most consistent with `ExcitedStates`. all -> all is probably too expensive by default...

So I'm going to adapt to 1 initial -> all final and then we can decide whether we like it or not 😄

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Yes, I know what you meant. 👍  Still, I could not figure out the correct spot to add the `symmetrise` call... There is already an overall `symmetrise` around the expression, but this seemingly does not suffice, and doing e.g. `einsum('ik,kj->ij', p1_oo, p0_oo).symmetrise()` also led to failing tests.

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: My bad... we only have ISR(2), but of course the mapping to ADC(3/2) properties happens one layer above 😄
I'm going to remove the `adc3` key here.

adcc/adc_pp/state2state_transition_dm.py
mfherbst: You kept in mind that unlike adcman we have an 0.5 inside our definition of symmetrise? (Sorry for asking these stupid questions, but it's just stuff I often forget).

maxscheurer: Yeah, so to make it clear, what I did was to convert this expression
```Python
0.5 * (einsum('ik,kj->ij', p1_oo, p0_oo) + einsum('ik,kj->ij', p0_oo, p1_oo))
```
to
```Python
einsum('ik,kj->ij', p1_oo, p0_oo).symmetrise()
```

mfherbst: That's strange, but ok ... let's not bother with this until we have a working version.

mfherbst:
> There is a bug in the testdata generator where from and to are flipped over, see

I just spent some time checking this and I am no longer convinced this is a bug by comparing with the adcman output. Before making a call on this issue we have to investigate more thoroughly.

maxscheurer:
I think you're right. I've made some changes to have a more consistent interface... but I have no idea why the densities are still different. Transition moments seem to be okay...

adcc/ElectronicTransition.py
maxscheurer: Done.

adcc/ElectronicTransition.py
maxscheurer: Done.

adcc/ElectronicTransition.py
maxscheurer: Done.

adcc/ElectronicTransition.py
maxscheurer: 1 initial -> all final now implemented.

adcc/test_properties.py
maxscheurer: Done.

maxscheurer:
I think we need a decision on how to proceed with this PR. The question is if we want to hunt down what causes the discrepancies in the densities (I currently don't have an idea why this could be, so I don't know where to start debugging). It's strange that already `adc0`/`adc1` densities are wrong...

mfherbst:
Ok so one thing is clear. We can't merge it before we don't understand the discrepancies.

So let's start over from the beginning step by step. First we make a list. Let's pick the water sto-3g (using the geometry in the adcc repo) as the simplest test case and just manually compare one more time:

- [ ] Ensure adcc results agree with adcman results
- [ ] Ensure testdata agrees with adcman output
- [ ] testdata agrees with adcc, i.e. tests pass.

If something comes to your mind, just add it. If you have done one thing, tick it off. I'll try to do the second thing tonight and then we see how to continue.

mfherbst:
Ah sorry that was wrong thinking, because the dipole moments actually agree, right? It's only the densities which differ?

mfherbst:
adcman sometimes does weird things where the order of from state and to state are weirdly mixed (so to, from or something like that). Could it be that we differ in that aspect? (Note: That might not be an error, that might just cause matrices to be transposed or something like that).

maxscheurer:
Yes, transition dipole moments are correct, but the densities are not in agreement.

maxscheurer:
They are indeed just transposed... 🤦

mfherbst:
Yeah ok ... but why? If we understand that we are good to go.

adcc/ElectronicTransition.py
maxscheurer: Done.

adcc/test_state_densities.py
maxscheurer: Done.

maxscheurer:
I'm investigating...

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Done.

adcc/ElectronicTransition.py
maxscheurer: I made a version which is low-level-constructible 😄

maxscheurer:
I think the testdata from/to are reversed as you said previously [here](https://github.com/adc-connect/adcc-testdata/blob/9332642fe7b21e2d0fcf249ee4799cb0f90c0070/adcctestdata/dump_reference.py#L207). The key format is definitely `from-to` according to my investigation.

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,157 @@
...
+            raise ValueError("Mismatch in subspaces doubles part "
+                             f"(== {u2.subspaces}), where "
+                             f"{b.o}{b.o}{b.v}{b.v} was expected.")
+
+
+def s2s_tdm_adc0(mp, amplitude_l, amplitude_r, intermediates):
+    _check_have_singles_block(amplitude_l, amplitude_r)
+    _check_singles_subspaces(amplitude_l, amplitude_r)
+    u1 = amplitude_l["s"]
+    x1 = amplitude_r["s"]
```
mfherbst, line 70: Sorry to be a bother but I changed my mind after the whole density transpose stuff. I think the `l` and `r` should be visible here. How about `ul1` and `ur1` ?

mfherbst:
> The key format is definitely from-to according to my investigation.

Hmm. Let's discuss. So my argument is: If you have a transition `x -> y` then the transition dipole moment is `tr (y* μ x)`. Now representing both states in a finite AO basis `C` (i.e projecting into the space spanned by `C`) gives the approximation `tr[y* C C* μ C C* x] = tr[M C* x (C* y)*]`, where `M` are the dipole moment integrals in AO. Denoting now `X = C*x` and similarly `Y` the representation of the states `y` and `x` in the same AO basis yields `tr(M X Y*)`, i.e. the transition density matrix `X Y*`. In other words the final state should be the one which gets an adjoint (no surprise actually if you look at the transition dipole integral `<y | μ x>`).

In your implementation `amplitude_r` gets the adjoint (see e.g. `s2s_tdm_adc0`), therefore it is the final state. In other words

https://github.com/adc-connect/adcc/blob/50e05531a3529742ad07ae678b7a5f8d4b31670e/adcc/State2States.py#L82-L86

is the right way round and our transition density matrices should be correct. Would you agree to that reasoning? What does @Drrehn think?

maxscheurer:
I have to check the Adcman paper again... the implementation should be in agreement with the published equations. So, if we can all agree on the way it is done in the Adcman paper, then I'm fine with that.☺️

maxscheurer:
After a long and painful struggle with these transition densities, it's looking really good to me now. 🎉

mfherbst:
OK. I'll try to take a look tonight.

adcc/ElectronicTransition.py
```diff
@@ -0,0 +1,323 @@
...
+from scipy import constants
+from matplotlib import pyplot as plt
+from .Excitation import mark_excitation_property
+from .solver.SolverStateBase import EigenSolverStateBase
+
+
+class ElectronicTransition:
+    def __init__(self, data, method=None, property_method=None):
+        """Construct an ElectronicTransition class from some data obtained
+        from an interative solver or other suitable data.
```
mfherbst, line 41: "other suitable data" is a bit vague. I think I would put at least one more example.

adcc/State2States.py
```diff
@@ -0,0 +1,89 @@
...
+            from an interative solver or other suitable data.
+
+            The class provides access to the results from an ADC calculation
+            as well as derived properties. Properties are computed lazily
+            on the fly as requested by the user.
+
+            By default the ADC method is extracted from the data object
+            and the property method in property_method is set equal to
+            this method, except ADC(3) where property_method=="adc2".
+            This can be overwritten using the parameters.
```
mfherbst, line 45: I don't think I would just copy the docs from the ExcitedStates class here.

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,131 @@
...
+from adcc.functions import einsum
 from adcc.AmplitudeVector import AmplitudeVector
+from adcc.OneParticleOperator import OneParticleOperator
 
 import libadcc
 
-DISPATCH = {}  # None implemented
+
+# TODO: just a sketch for what I'd like to have, because we need to check on
+# 2 amplitude vectors simultaneously
```
mfherbst, line 33: These functions should not be here. I think they are useful, though. I suppose they should go to some sort of util or common file in adc_pp.

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,131 @@
...
+
+def s2s_tdm_adc2(mp, amplitude_l, amplitude_r, intermediates):
+    _check_have_doubles_block(amplitude_l, amplitude_r)
+    _check_doubles_subspaces(amplitude_l, amplitude_r)
+    dm = s2s_tdm_adc0(mp, amplitude_l, amplitude_r, intermediates)
+
+    ul1 = amplitude_l["s"]
+    ur1 = amplitude_r["s"]
+    ul2 = amplitude_l["d"]
+    ur2 = amplitude_r["d"]
```
mfherbst, line 86: For these sort of things I tend to prefer
```
ul1, ul2 =  amplitude_l["s"],  amplitude_l["d"]
ur1, ur2 =  amplitude_r["s"],  amplitude_r["d"]
```

but I suppose one can argue about that.

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -20,12 +20,131 @@
...
+    # ADC(2) ISR intermediate (TODO Move to intermediates)
+    rul1 = einsum('ijab,jb->ia', t2, ul1).evaluate()
+    rur1 = einsum('ijab,jb->ia', t2, ur1).evaluate()
+
+    dm[b.oo] = (
+        p1_oo - 2.0 * einsum('ikab,jkab->ij', ur2, ul2)
+        + 0.5 * einsum('ik,kj->ij', p1_oo, p0_oo)
+        + 0.5 * einsum('ik,kj->ij', p0_oo, p1_oo)
+        - 0.5 * einsum('ikcd,lk,jlcd->ij', t2, p1_oo, t2)
+        + einsum('ikcd,jkcb,db->ij', t2, t2, p1_vv)
```
mfherbst, line 104: Again a question of style, but I think I would put a `1.0 * ` in such places.

mfherbst (approved):
Very nice. A few nits only.

Also: Can you add an issue about the missing CVS S2S TDMs, just that we can keep track.

adcc/ElectronicTransition.py
mfherbst: There are a few places where this occurs btw.

adcc/adc_pp/state2state_transition_dm.py
mfherbst: Also can you redirect the checks from the other implementations (MTM, other densities) to these new functions.

mfherbst:
Ah and another small thing: Could you add something to our docs about `State2States`. What I have in mind is maybe the way to plot a spectrum for spin-flip to the tutorial and one of the spin-flip examples could be good.

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Sure. MTMs don't involve any AmplitudeVectors though 😉

adcc/ElectronicTransition.py
maxscheurer: Done.

adcc/State2States.py
maxscheurer: Done.

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Done.

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Done.

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Done.

maxscheurer:
I've addressed the reviews and added a small paragraph to the docs.

adcc/adc_pp/util.py
```diff
@@ -0,0 +1,67 @@
...
+    if any("d" not in amplitude.blocks for amplitude in amplitudes):
+        raise ValueError("state2state_transition_dm at ADC(2) level and "
+                         "beyond expects an excitation amplitude with a "
+                         "singles and a doubles part.")
+
+
+def check_singles_subspaces(*amplitudes, core_space=False):
+    C = b.c if core_space else b.o
+    for amplitude in amplitudes:
+        ul1 = amplitude["s"]
```
mfherbst, line 53: ul1 makes little sense here.

adcc/adc_pp/util.py
```diff
@@ -0,0 +1,67 @@
...
+def check_have_singles_block(*amplitudes):
+    if any("s" not in amplitude.blocks for amplitude in amplitudes):
+        raise ValueError("ADC(0) level and "
+                         "beyond expects an excitation amplitude with a "
+                         "singles part.")
+
+
+def check_have_doubles_block(*amplitudes):
+    if any("d" not in amplitude.blocks for amplitude in amplitudes):
+        raise ValueError("state2state_transition_dm at ADC(2) level and "
```
mfherbst, line 45: state2state_transition_dm makes little sense here

adcc/adc_pp/util.py
```diff
@@ -0,0 +1,67 @@
...
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with adcc. If not, see <http://www.gnu.org/licenses/>.
+##
+## ---------------------------------------------------------------------
+from adcc import block as b
+
+
+def check_singles_amplitudes(*amplitudes, core_space=False):
```
mfherbst, line 26: This core space thing is weird. I think I would pass the subspaces list explicitly, since this makes it more explicit in the calling functions what happens here and what is checked for.

mfherbst (commented):
Perfect on the docs, that's exactly what I had in mind. Got a few nits regarding the refactor, though.

adcc/adc_pp/util.py
mfherbst: (This is at two places).

maxscheurer: Done.

adcc/adc_pp/util.py
maxscheurer: Done.

adcc/adc_pp/util.py
maxscheurer: Now explicit list for spaces.

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -31,7 +31,7 @@
 
 
 def s2s_tdm_adc0(mp, amplitude_l, amplitude_r, intermediates):
-    check_singles_amplitudes(amplitude_l, amplitude_r)
+    check_singles_amplitudes(amplitude_l, amplitude_r, spaces=[b.o, b.v])
```
mfherbst, line 34: Sorry to be a bother ... but for this one I would prefer brevity, so
```suggestion
    check_singles_amplitudes(amplitude_l, amplitude_r, [b.o, b.v])
```

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -42,7 +42,8 @@ def s2s_tdm_adc0(mp, amplitude_l, amplitude_r, intermediates):
 
 
 def s2s_tdm_adc2(mp, amplitude_l, amplitude_r, intermediates):
-    check_doubles_amplitudes(amplitude_l, amplitude_r)
+    check_doubles_amplitudes(amplitude_l, amplitude_r,
+                             spaces=[b.o, b.o, b.v, b.v])
```
mfherbst:
```suggestion
    check_doubles_amplitudes(amplitude_l, amplitude_r, [b.o, b.o, b.v, b.v])
```

adcc/adc_pp/util.py
```diff
@@ -20,17 +20,16 @@
...
 ##
 ## ---------------------------------------------------------------------
-from adcc import block as b
 
 
-def check_singles_amplitudes(*amplitudes, core_space=False):
+def check_singles_amplitudes(*amplitudes, spaces):
     check_have_singles_block(*amplitudes)
-    check_singles_subspaces(*amplitudes, core_space=core_space)
+    check_singles_subspaces(*amplitudes, spaces=spaces)
```
mfherbst, line 27:
```suggestion
    check_singles_subspaces(*amplitudes, spaces)
```

adcc/adc_pp/util.py
```diff
@@ -20,17 +20,16 @@
...
 
 
-def check_doubles_amplitudes(*amplitudes, core_space=False):
+def check_doubles_amplitudes(*amplitudes, spaces):
     check_have_doubles_block(*amplitudes)
-    check_doubles_subspaces(*amplitudes, core_space=core_space)
+    check_doubles_subspaces(*amplitudes, spaces=spaces)
```
mfherbst, line 32:
```suggestion
    check_doubles_subspaces(*amplitudes, spaces)
```

adcc/adc_pp/state2state_transition_dm.py
maxscheurer: Won't work this way because of the `*amplitudes` argument...

mfherbst: ah ... yikes you are right. That's annoying. Ok than leave it as is.

maxscheurer: I committed another suggestion 😄

mfherbst:
Good idea.

maxscheurer:
Should we take the current version? 😄

mfherbst:
apparently ...

mfherbst:
I thought that was clear ... sorry.

PR #93, opened 2020-11-06
Introduce InputError

mfherbst:
During the parsing stage in run_adc, adcc now throws an InputError as a specialised "ValueError"

### DIFF
--- a/adcc/__init__.py
+++ b/adcc/__init__.py
@@ -23,6 +23,8 @@
 import sys
 import warnings
 
+from libadcc import HartreeFockProvider, get_n_threads, set_n_threads
+
 from .LazyMp import LazyMp
 from .Tensor import Tensor
 from .Symmetry import Symmetry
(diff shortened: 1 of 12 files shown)


maxscheurer (approved):

maxscheurer:
What is wrong with the tests?! The error messages look obnoxious 🔥 🚒

mfherbst:
I'm on it. pybind11 found a memory issue in the way we use their interface in the HF provider, so now "our" way of doing it is no longer allowed for safety.

maxscheurer:
Okay, let me know if I can do anything. Is this because of the new pybind11 2.6?

mfherbst:
Yes indeed. The changes were introduced in 2.6.

mfherbst:
And now we are hitting the 4th of https://docs.h5py.org/en/stable/whatsnew/3.0.html#breaking-changes-deprecations ...

mfherbst:
It's always the same when you just want to change a little thing :smile:

maxscheurer:
Will the changes be backward compatible for pb11 and h5?

mfherbst:
For pybind11 it's a bit tricky. The HDF5 stuff I might manage.

maxscheurer:
Yes, pybind11 is probably not as bad.. but since h5 gives runtime errors...

mfherbst:
What do you mean by runtime errors?

maxscheurer:
Ahh never mind... I thought the pybind11 error was thrown at compile time... but it wasn't.. so yeah it does not matter 😄

mfherbst:
Hooray ... nah that was a piece of cake, wasn't it ;)

maxscheurer:
Utterly frustrating...

PR #95, opened 2020-11-11
Implementation of Block-Lanczos procedure

mfherbst:
Less useful for standard ADC use cases, but very useful for debugging and for special applications (e.g. large number of eigenpairs).

## TODO
- [x] Shift-and-invert matrix
- [x] Docstrings

### DIFF
--- a/adcc/solver/LanczosIterator.py
+++ b/adcc/solver/LanczosIterator.py
@@ -32,6 +32,31 @@ from .orthogonaliser import GramSchmidtOrthogonaliser
 class LanczosIterator:
     def __init__(self, matrix, guesses, ritz_vectors=None, ritz_values=None,
                  ritz_overlaps=None, explicit_symmetrisation=None):
+        """
+        Initialise an iterator generating :py:class:`LanczosSubspace` objects,
+        which represent a growing Krylov subspace started from the `matrix`
+        and guess vectors `guesses` or optionally using a thick restart from
+        the passed Ritz vectors, Ritz values and Ritz overlaps.
+
+        Parameters
+        ----------
+        matrix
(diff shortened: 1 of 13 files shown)


adcc/solver/LanczosIterator.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # To be initialised by __iter__
+        self.lanczos_subspace = []
+        self.alphas = []  # Diagonal matrix block of subspace matrix
+        self.betas = []   # Side-diagonal matrix blocks of subspace matrix
+        self.residual = guesses
+        self.n_iter = 0
+        self.n_applies = 0
+
+    def __iter__(self):
```
maxscheurer, line 80: Where are these functions actually called? 😄 
I'm a bit overwhelmed by the Python magic probably...

mfherbst: Well basically this is an iterator that "iterates" through the Lanczos subspace. They are implicitly called in the loop "for subspace in iterator" in the lanczos.py file and setup and advance the state.

maxscheurer: That's cool 👍

maxscheurer (approved):
I cannot comment on this from an "algorithm"/math point of view, but the code looks very clean, well documented, etc. 😄

mfherbst:
No worries. I hope it's doing good. I first implemented it in Julia where I really treated it super badly when it comes to numerics, then I transferred it to python and checked every step. It might still be wrong but the final residuals will always tell you, so at worst it just does not find the eigenvalues, but wrong results will not go unnoticed.

Thanks for looking at the code and documentation!

PR #96, opened 2020-11-13
Implement to_qcvars function

mfherbst:
Adding to_qcvars function, which could serve https://github.com/MolSSI/QCEngine/pull/277.

Actually getting this to be useful for Psi4 is not so easy, so I would not bother, but this poses the question why to even have this function in the first place at it might be pretty specific to QCEngine. On the other hand this allows us to add new properties to adcc and make them available to QCEngine without the need to modify code in that repo all the time, which is probably easier to maintain.

@maxscheurer Thoughts?

### DIFF
--- a/adcc/ExcitedStates.py
+++ b/adcc/ExcitedStates.py
@@ -719,6 +719,39 @@ class ExcitedStates:
                 ret.append(key)
         return ret
 
+    def to_qcvars(self, properties=False, recurse=False):
+        """
+        Return a dictionary with property keys compatible to a Psi4 wavefunction
+        or a QCEngine Atomicresults object.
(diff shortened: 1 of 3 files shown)


maxscheurer (approved):
I think it is worthwile having these functions despite some duplication in Psi4. In the long run we'll benefit a lot from qcng, so we can also maintain the convenience function for qcvars. Thanks for the nice wrap-up inside adcc!

mfherbst:
Ok then I guess we release after this PR goes in and modify QCEngine accordingly?

maxscheurer:
Fine by me. I guess no other adcc-side changes are required so far 😉

PR #80, opened 2020-06-18
Implement guesses in python

mfherbst:
- [x] Zero vectors of appropriate symmetry
- [x] Singles guesses
- [ ] Doubles guesses

PR #98, opened 2020-11-22
Implement singles guesses in python

mfherbst:
Superseeds #80. Needs **adccore 0.14.4**.


mfherbst:
Cool, seems to be working. All that's left is the macos deployment for adccore.

maxscheurer (approved):
👍

PR #99, opened 2020-11-24
Loosen platform restrictions

mfherbst:


maxscheurer:
I get the same error as the CI. In my case `os_platform_tag` equals "macosx_10_15_x86_64".

mfherbst:
Ah hmm. Can you try adapting the code and just push what works on your end?

mfherbst:
Ah no the issue is
https://github.com/adc-connect/adcc/blob/master/extension/AdcCore.py#L38
Can you just use the distutils call without calling get_platform?

extension/AdcCore.py
```diff
@@ -249,14 +249,14 @@ def agrees_with_os_platform(self):
...
+            os_platform_tag = distutils.util.get_platform()
             _, os_version, _ = os_platform_tag.split('-')
             os_version = tuple([int(x) for x in os_version.split(".")])
 
-            if "-" not in self.platform or not self.platform.startswith("macosx"):
+            if "_" not in self.platform or not self.platform.startswith("macosx"):
                 return False
             try:
-                core_version = self.platform.split('-')[1]
+                core_version = self.platform.split('_')[1]
```
mfherbst, line 259: I don't think this works as this should have the same format as the `get_platform`function, so you would need `[1:2]`

extension/AdcCore.py
```diff
@@ -249,14 +249,14 @@ def agrees_with_os_platform(self):
         if sys.platform == "linux":
             return get_platform() == self.platform
         elif sys.platform == "darwin":
-            os_platform_tag = get_platform()
+            os_platform_tag = distutils.util.get_platform()
             _, os_version, _ = os_platform_tag.split('-')
```
mfherbst: Actually since you cannot change the code for the `self.platfrom` I think I would vote for keeping the `get_platform` here as well and making both blocks identical.

mfherbst:
@maxscheurer  Could you try again on your machine?

maxscheurer:
Yes, I was on the road, so I couldn't do anything 😄 Will test it now...

maxscheurer (approved):
It works fine now.

mfherbst:
Ok cool, so we only need to build on 10.14 and we're able to support 10.15 out of the box.

PR #103, opened 2020-11-27
Fixes for conda on GHA

maxscheurer:


setup.py
```diff
@@ -64,6 +67,12 @@ def get_adccore_data():
 
     import AdcCore
 
+    if is_conda_build():
+        os.environ["LDFLAGS_LD"] = \
+            os.environ["LDFLAGS_LD"].replace("-dead_strip_dylibs", "")
+        os.environ["LDFLAGS"] = \
+            os.environ["LDFLAGS"].replace("-Wl,-dead_strip_dylibs", "")
```
mfherbst, line 74: Ok this obviously modifies the environment for the full built. Is this exactly what is needed?

mfherbst (approved):
Could you bump the GHA workflow to macOS 10.15?

setup.py
maxscheurer: Yes, because somehow, the destructive flags otherwise get picked up again somehow.
I find this also super annoying and ugly.

mfherbst: Me too, but ok. Let's leave it for the time being. Once we manage to redo our build system with the OSS libtensor we can revisit this. Can you add a TODO?

mfherbst:
Do you want to do a dry run first or shall we directly bump and release?

maxscheurer:
For `conda`, we don't have enough minutes left for a dry run. It's all or nothing. I added payment info to GH so that we can get this release out, no matter what 😄

PR #104, opened 2020-12-02
Python-side Møller-Plesset

maxscheurer:
## ToDo
- [x] all MP equations py-side
- [x] heavy rebase on #71 
- [x] caching functionality
- [x] remove `set_t2`
- [x] timer functionality
- [x] remove `libadcc.LazyMp`
- [x] merge `mp.py` and `LazyMp.py` once #71 is in
- [x] fix conda env issue 
- [x] cache statistics OK
- [x] cleanup


adcc/LazyMp.py
```diff
@@ -138,7 +143,9 @@ def t2eri(self, space, contraction):
         }
         return expressions[space + contraction].evaluate()
 
+    # TODO: correct caching with set_t2?
```
mfherbst, line 146: Or just skip supporting it. To do this properly we would have to extend our equations anyway (some implicit simplifications in the adc equations no longer hold IRC from discussions with Manu. So I suggest we just scratch this (unused) feature for now.

maxscheurer: I've come up with a solution for this. If the feature is not required, we could instead use `functools.lru_cache`, which I would prefer indeed.

adcc/misc.py
```diff
@@ -47,6 +48,35 @@ def get(self):
     return property(get)
 
 
+def cached_function(function):
```
mfherbst, line 51: Good idea, but what I don't like about it is that this way the cache is not really easily accessible, which could be an issue in the future. I have to think about this a bit more.

maxscheurer: Sure, you can just `print(mp.t2.cache)` the way I implemented this.

adcc/LazyMp.py
maxscheurer: With `lru_cache`, there is however no way at all to access the cache. Just a "clear" function can be called.

adcc/misc.py
mfherbst: Well ok, but what I meant is that there is no "central" place to manage the cache, but then otoh you can just loop over the cached functions from lazymp somehow. Hmm. Not sure yet if I like it or not at the moment. Just leave it for now and we'll see.

maxscheurer:
I think we should drop support for the `set_t2` for now. It's not really essential and complicates the code imho...

mfherbst:
Yes I agree. Feel free to remove it.

adcc/misc.py
maxscheurer: I have now implemented a "central" cache 😄

adcc/LazyMp.py
```diff
@@ -96,7 +96,8 @@ def t2(self, space):
     def td2(self, space):
         """Return the T^D_2 term"""
         if space != b.oovv:
-            raise NotImplementedError()
+            raise NotImplementedError("T^D_2 term not implemented "
+                                      f"for space {space}.")
```
mfherbst, line 100: I'm pretty sure we can lift this now that we don't rely on libadc any more.

maxscheurer: Yes, but then I have to implement the amplitudes for CVS ☹️
I can give it a try...

mfherbst: Don't bother if it's too annoying. It would just be nice to be able to get an MP(3) energy also for CVS-ADC(3).

maxscheurer: It's really annoying 😄 Let's do this in a follow-up, because it's taking too much time right now.

maxscheurer:
60 lines net addition to move all the MP stuff equations py-side... 😄 
I'm going to do the same with `OneParticleOperator` in a follow-up PR once I come up with a strategy how to pythonize it.

mfherbst:
Not super critical since it's not blocking the path to remove adccore.

adcc/LazyMp.py
```diff
@@ -60,39 +57,80 @@ def __getattr__(self, attr):
 
     def df(self, space):
         """Delta Fock matrix"""
-        # TODO This is so cheap to compute that it does not really deserve
-        #      any caching
-        return self.__cpp.df(space)
+        hf = self.reference_state
+        s1 = space[:2]
+        s2 = space[2:]
```
mfherbst, line 62: I would write a function that (a) checks that the spaces are sensible and (b) splits them up into chunks. What I have in mind is:
```
s1, s2 = split_spaces(space, 2)
```
That you can then reuse for all other similar tasks in the other functions as well.

adcc/LazyMp.py
```diff
@@ -60,39 +57,80 @@ def __getattr__(self, attr):
...
-
-    def set_t2(self, space, tensor):
-        """
-        Set the T2 amplitudes tensor. This invalidates all data depending on the T2
-        amplitudes in this class. Note, that potential other caches, such as
-        computed ADC intermediates are *not* automatically invalidated.
-        """
-        return self.__cpp.set_t2(space, tensor)
+        hf = self.reference_state
+        if space == b.oovv:
```
mfherbst, line 71: If you have the split-space function from above all these cases can probably unified into one code path. (where you just check that the last two spaces are `v` and then build the df spaces from the first two.

mfherbst: Then you can also scratch the compute_t2.

adcc/LazyMp.py
```diff
@@ -60,39 +57,80 @@ def __getattr__(self, attr):
...
         """Return the T^D_2 term"""
-        return self.__cpp.td2(space)
+        if space != b.oovv:
+            raise NotImplementedError("T^D_2 term not implemented "
+                                      f"for space {space}.")
+        t2erit = self.t2eri(b.oovv, b.ov).transpose((1, 0, 2, 3))
+        denom = direct_sum(
+            'ia,jb->ijab', self.df(b.ov), self.df(b.ov)
+        ).symmetrise(0, 1)
+        return evaluate(
```
mfherbst, line 95: Might be something we should debate, but for me calling `evaluate` is not the job of the expression, but the job of the caching.

adcc/LazyMp.py
```diff
@@ -60,39 +57,80 @@ def __getattr__(self, attr):
...
+            (+ 4.0 * t2erit.antisymmetrise(2, 3).antisymmetrise(0, 1)
+             - 0.5 * self.t2eri(b.oovv, b.vv)
+             - 0.5 * self.t2eri(b.oovv,  b.oo)) / denom
```
mfherbst, line 98: Just to check: `/ denom` does not screw up the symmetry, i.e. libadc does the same thing here?

mfherbst: Also two spaces before `b.oo`

adcc/LazyMp.py
```diff
@@ -60,39 +57,80 @@ def __getattr__(self, attr):
...
 
+    @cached_member_function
     def t2eri(self, space, contraction):
         """
         Return the T2 tensor with ERI tensor contraction intermediates.
         These are called pi1 to pi7 in libadc.
         """
-        return self.__cpp.t2eri(space, contraction)
+        hf = self.reference_state
+        expressions = {
```
mfherbst, line 108: Again you might be able to simplify this if you parse the spaces.

adcc/LazyMp.py
```diff
@@ -60,39 +57,80 @@ def __getattr__(self, attr):
...
         Return the MP2 differensce density in the MO basis.
         """
-        ret = OneParticleOperator.from_cpp(self.__cpp.mp2_diffdm)
+        hf = self.reference_state
+        ret = OneParticleOperator(self.mospaces, is_symmetric=True)
+        blocks = ["oo", "ov", "vv"]
+        if self.has_core_occupied_space:
+            blocks += ["cc", "co", "cv"]
+        for bl in blocks:
+            ba = getattr(b, bl)
```
mfherbst, line 132: I'm debating with myself whether being explicit and having the list in `o1o1` etc. is better or not.

adcc/LazyMp.py
```diff
@@ -122,9 +160,24 @@ def dipole_moment(self, level=2):
...
+                                      "not implemented.")
+        if level < 2:
+            return 0.0
+        hf = self.reference_state
+        if level == 2 and not self.has_core_occupied_space:
+            return compute_energy_correction(hf.oovv, self.t2oo)
+        elif level == 2 and self.has_core_occupied_space:
+            pairs = [(self.t2oo, hf.oovv),
+                     (2.0 * self.t2oc, hf.ocvv),
+                     (self.t2cc, hf.ccvv)]
```
mfherbst, line 177: This again feels like something which you should be able to unify if you loop over spaces (and a coefficient). In that way you can also scratch the far too simple `compute_energy_correction` function.

adcc/LazyMp.py
```diff
@@ -60,39 +57,80 @@ def __getattr__(self, attr):
...
         """
-        ret = OneParticleOperator.from_cpp(self.__cpp.mp2_diffdm)
+        hf = self.reference_state
+        ret = OneParticleOperator(self.mospaces, is_symmetric=True)
+        blocks = ["oo", "ov", "vv"]
+        if self.has_core_occupied_space:
+            blocks += ["cc", "co", "cv"]
+        for bl in blocks:
+            ba = getattr(b, bl)
+            ret[ba] = compute_mp2_diffdm(hf, self, bl, apply_cvs=False)
```
mfherbst, line 133: Don't forget to evaluate somewhere.

adcc/LazyMp.py
```diff
@@ -179,3 +232,105 @@ def mp2_dipole_moment(self):
...
+    applied during the density matrix computation even if it is
+    applied in the hf reference. This is done in order to be
+    able to obtain the *full* MP2 difference density (needed for
+    MP properties and MP energies), while keeping the CVS is needed
+    for the CVS-ADC(3) matrix for example.
+    """
+    if apply_cvs is None:
+        apply_cvs = hf.has_core_occupied_space
+
+    # TODO: too much if/else?
```
mfherbst, line 266: Absolutely.

I think there are two options and you have to see which one is better: Either you write the expressions to a dict (i.e.
```
expressions[b.oo] = ...
if cvs:
    expressions[b.oo] = ...
```
) or you have separate small functions for them and then dispatch. Perhaps also a mixture could be reasonable (one function for keeping cvs one for dropping it and inside you have the expression dicts.

setup.py
```diff
@@ -69,10 +69,12 @@ def get_adccore_data():
 
     # TODO: find a more sustainable and clean solution ASAP
     if is_conda_build():
-        os.environ["LDFLAGS_LD"] = \
-            os.environ["LDFLAGS_LD"].replace("-dead_strip_dylibs", "")
-        os.environ["LDFLAGS"] = \
-            os.environ["LDFLAGS"].replace("-Wl,-dead_strip_dylibs", "")
+        clean_flags = ["LDFLAGS_LD", "LDFLAGS"]
+        for flag in clean_flags:
```
mfherbst, line 73: Merge these lines and make the list a tuple (faster).

setup.py
```diff
@@ -69,10 +69,12 @@ def get_adccore_data():
...
+            if flag not in os.environ:
+                continue
+            os.environ[flag] = \
+                os.environ[flag].replace("-dead_strip_dylibs", "")
```
mfherbst, line 77: I usually prefer in for loops:
```
if flag in environ:
   do something
```

adcc/misc.py
```diff
@@ -47,6 +48,38 @@ def get(self):
...
+def cached_member_function(function):
+    """
+    Decorates a member function being called with
+    one or more arguments and stores the results
+    in field `_function_cache` of the class instance.
+    """
+    fname = function.__name__
+
+    @wraps(function)
+    def wrapper(*args):
```
mfherbst, line 60: `(self, *args)`

adcc/misc.py
```diff
@@ -47,6 +48,38 @@ def get(self):
...
+        self = args[0]
+        key = args[1:]
+        if not hasattr(self, "_function_cache"):
+            self._function_cache = {}
+        cache = self._function_cache
+        if fname not in cache:
+            cache[fname] = {}
```
mfherbst, line 67: The pythonic way is to catch the exceptions here, see the implementation of the function above. Typically gives rise to less code.

mfherbst (commented):
One important thing to check is actually timings of this class. I would suspect that the caching mechanism in C++ is way faster than in python, but it would be interesting to see how much we loose. Could you do that? Just compare between the LazyMp purely in python and the python/c++ mixed version.

adcc/misc.py
maxscheurer: I think in this case this may lead to too many `try` & `except` in a row, because we have to take care of a) the function and b) the correct parameters. I can give it a try, but I'm not a huge fan of the exception stuff.

adcc/LazyMp.py
maxscheurer: I think one needs another solution for this anyways, because if we use this dictionary, the tensor expressions won't be evaluated, but the integrals will be imported. This is against the 'first-demand'-driven strategy and we need to find another solution for this.

maxscheurer: Resolved.

adcc/misc.py
maxscheurer: Done.

adcc/misc.py
maxscheurer: Done with `try/except` now 👍 I like it better this way.

adcc/LazyMp.py
```diff
@@ -60,39 +63,103 @@ def __getattr__(self, attr):
...
+            ret[b.cv] = -0.5 * (
+                - 1.0 * einsum("jIbc,jabc->Ia", self.t2oc, hf.ovvv)
+                + 1.0 * einsum("jkIb,jkab->Ia", hf.oocv, self.t2oo)
+                + 1.0 * einsum("jMIb,jMab->Ia", hf.occv, self.t2oc)
+                + 1.0 * einsum("ILbc,Labc->Ia", self.t2cc, hf.cvvv)
+                + 1.0 * einsum("kLIb,kLab->Ia", hf.occv, self.t2oc)
+                + 1.0 * einsum("LMIb,LMab->Ia", hf.cccv, self.t2cc)
+            ) / self.df(b.cv)
+        for bb in ret.blocks_nonzero:
+            ret[bb].evaluate()
```
mfherbst, line 162: If we don't yet support ret.evaluate() directly, I would add it.

mfherbst: Also I would set ret.reference_state and then just return evaluate(ret), but that's a matter of taste, just makes the flow clearer to me somehow.

adcc/LazyMp.py
```diff
@@ -60,39 +63,103 @@ def __getattr__(self, attr):
...
+            ret[b.co] = -0.5 * (+ einsum("kIab,kjab->Ij", self.t2oc, self.t2oo)
+                                + einsum("ILab,jLab->Ij", self.t2cc, self.t2oc))
+            ret[b.cv] = -0.5 * (
+                - 1.0 * einsum("jIbc,jabc->Ia", self.t2oc, hf.ovvv)
+                + 1.0 * einsum("jkIb,jkab->Ia", hf.oocv, self.t2oo)
+                + 1.0 * einsum("jMIb,jMab->Ia", hf.occv, self.t2oc)
+                + 1.0 * einsum("ILbc,Labc->Ia", self.t2cc, hf.cvvv)
+                + 1.0 * einsum("kLIb,kLab->Ia", hf.occv, self.t2oc)
+                + 1.0 * einsum("LMIb,LMab->Ia", hf.cccv, self.t2cc)
+            ) / self.df(b.cv)
```
mfherbst, line 160: Much better!

I would homogenise them more. Sometimes you have a 1.0 * in front, sometimes the -0.5 appears on the same line as the = that's sometimes a bit unexpected, so I would rather do it one way.

adcc/LazyMp.py
```diff
@@ -60,39 +63,103 @@ def __getattr__(self, attr):
...
+            ret[b.oo] += -0.5 * einsum("iLab,jLab->ij", self.t2oc, self.t2oc)
+            ret[b.ov] += -0.5 * (
+                + einsum("jMib,jMab->ia", hf.ocov, self.t2oc)
+                + einsum("iLbc,Labc->ia", self.t2oc, hf.cvvv)
+                + einsum("kLib,kLab->ia", hf.ocov, self.t2oc)
+                + einsum("iMLb,LMab->ia", hf.occv, self.t2cc)
+                - einsum("iLMb,LMab->ia", hf.occv, self.t2cc)
+            ) / self.df(b.ov)
+            ret[b.vv] += (+ 0.5 * einsum("IJac,IJbc->ab", self.t2cc, self.t2cc)
+                          + 1.0 * einsum("kJac,kJbc->ab", self.t2oc, self.t2oc))
```
mfherbst, line 147: For some reason I don't like the + in front if it is not the first thing in the line .. but you decide, just a minor point.

mfherbst (commented):
Some notes ;)

setup.py
maxscheurer: Done.

adcc/LazyMp.py
maxscheurer: Resolved.

adcc/LazyMp.py
maxscheurer: Also resolved.

adcc/LazyMp.py
maxscheurer: Done.

setup.py
maxscheurer: Done.

maxscheurer:
## Cache statistics:

Task: retrieve `t2`,`td2`, and `mp2_diffdm` from the cache (computing time is excluded),
*total* time for 10000 repeats.
```
#
#-- water aug-cc-pvdz
#
nocc = 10   nvirt = 72
full   mp_cache       228.55 ms      53.43 ms =  -175.12 ms   -76.62%
full   cvs-mp_cache   300.25 ms      53.31 ms =  -246.94 ms   -82.24%
```

So the Python cache is actually faster than what we have in current `master`.

adcc/LazyMp.py
maxscheurer: The + after ( ?
I can remove it but then it's not aligned with the `+ 1.0` in the next line...

maxscheurer: Resolved.

adcc/LazyMp.py
maxscheurer: Done.

adcc/LazyMp.py
maxscheurer: It's already there, so this is resolved 😄

adcc/LazyMp.py
maxscheurer: libadc uses `div`, so I guess we're good.

adcc/LazyMp.py
maxscheurer: Yes, that's right. Changed now, but I added `cached_member_function` also to `mp.df` because then we make sure it's properly evaluated + consistent with other routines.

adcc/LazyMp.py
maxscheurer: Done.

adcc/LazyMp.py
maxscheurer: Implemented now, but needs some "cleanup"/improvement 😄

maxscheurer: Resolved.

adcc/LazyMp.py
```diff
@@ -22,11 +22,24 @@
...
 from .OneParticleOperator import OneParticleOperator, product_trace
+from .Intermediates import register_as_intermediate
+from .timings import Timer, timed_member_call
+from . import block as b
+
+
+def split_spaces(spacestr):
+    sp = textwrap.wrap(spacestr, 2)
+    try:
+        assert all(s in ('o1', 'o2', 'v1') for s in sp)
```
mfherbst, line 39: (a) I think this is more general and should live in a util / common type file (maybe block?)
(b) what about the other spaces (o3, v2, v3 etc). Should not be too hard to support this.

adcc/LazyMp.py
```diff
@@ -40,61 +53,128 @@ def __init__(self, hf):
...
+                                      f"for space {space}.")
+        t2erit = self.t2eri(b.oovv, b.ov).transpose((1, 0, 2, 3))
+        denom = direct_sum(
+            'ia,jb->ijab', self.df(b.ov), self.df(b.ov)
+        ).symmetrise(0, 1)
+        return (
+            + 4.0 * t2erit.antisymmetrise(2, 3).antisymmetrise(0, 1)
+            - 0.5 * self.t2eri(b.oovv, b.vv)
+            - 0.5 * self.t2eri(b.oovv, b.oo)
+        ) / denom
```
mfherbst, line 103: Perfect!

adcc/LazyMp.py
```diff
@@ -122,9 +202,27 @@ def dipole_moment(self, level=2):
...
+            return 0.0
+        hf = self.reference_state
+        if level == 2 and not self.has_core_occupied_space:
+            terms = [(1.0, hf.oovv, self.t2oo)]
+        elif level == 2 and self.has_core_occupied_space:
+            terms = [(1.0, hf.oovv, self.t2oo),
+                     (2.0, hf.ocvv, self.t2oc),
+                     (1.0, hf.ccvv, self.t2cc)]
+        elif level == 3:
+            terms = [(1.0, hf.oovv, self.td2(b.oovv))]
```
mfherbst, line 221: So MP(3) on CVS now works just like this? I think you might be missing a case here.

adcc/LazyMp.py
```diff
@@ -179,3 +277,22 @@ def mp2_dipole_moment(self):
         mp2corr = -np.array([product_trace(comp, self.mp2_diffdm)
                              for comp in dipole_integrals])
         return refstate.dipole_moment + mp2corr
+
+
+#
+# Register cvs_p0 intermedites
+#
+@register_as_intermediate
+def cvs_p0_oo(hf, mp, intermediates):
```
mfherbst, line 286: Make connection to mp2_diffdm expressions clear by mutual comments (easier to refactor later because otherwise you might miss an instance).

adcc/LazyMp.py
maxscheurer: MP3 on CVS is wrong anyways... I think we should throw an exception right here instead of returning a wrong result.

adcc/LazyMp.py
maxscheurer: It used to be in `block`, but then the python 3.6 were failing, because when you say `b.split_spaces`, it dispatches to `__getattr__` and then fails 😄 I think `block` would be the right place though.

adcc/LazyMp.py
mfherbst: Definitely!

adcc/LazyMp.py
mfherbst: Ah yes, that's true. Maybe just split_spaces.py ?

maxscheurer: Or in MoSpaces.py? That would make sense, too.

mfherbst: Yes, ok, that's fine with me.

mfherbst:
If you're happy I would merge this once the CI is done.

maxscheurer:
Yes, I have squashed this to 4 commit, so you can just create a regular merge commit 👍

maxscheurer:
Thanks, LGTM! 👍

mfherbst (approved):

PR #106, opened 2020-12-06
Make adcc fully open-source

### LINKED ISSUE #105: Move adccore to adcc
Just an issue to keep track of things that need to be done:

- [x] merge #71 
- [x] merge #104 
- [x] Deploy libtensor to github release (for MacOS and Linux)
- [x] Deploy libtensor to conda (WIP already)
- [x] Merge codes
- [x] Restructure repo
- [x] Update deployment infrastructure
- [x] transfer issues from core repository

mfherbst:
PR to track the progress of making adcc fully open-source. Closes #105.

## Progress
- [x] Simplify the number of define flags and commandline flags needed to talk to libtensor
- [x] Get libtensor from github releases
- [x] MacOS support
- [x] C++ tests build
- [x] C++ tests integrated into setup.py test
- [x] Better file structure (libadcc/adcc is a bit weird)
- [x] Timings (to check we have not messed up)
  - [x] "normal" build
  - [x] conda build
- [x] Update setup.py installation (manifest, installed files, package files, ...)
- [x] Update scripts (grep for adccore)
- [x] Fix building docs
- [x] Update documentation
- [x] Conda support and upload
- [x] Merge deployment jobs into single file & enable all py-versions for Conda
- ~~[ ] LGTM for C++~~ ➡️ postpone
- [x] Coverage for C++
- [x] Code-style test for C++ using clang-format


run_tests.sh
```diff
@@ -18,3 +18,25 @@ elif [ "$PART" == "valgrind_pytest" ]; then
...
+if [ ! -d "$BUILD_DIR" ]; then
+	echo "Could not find build directory $BUILD_DIR" >&2
+	exit 1
+fi
+
+PART="$1"
+shift
+
+if [ "$PART" == "cpp" ]; then
+	cd "$BUILD_DIR" && "./src/adcc/tests/adccore_tests" "$@"
```
maxscheurer, line 33: Are these paths still correct?

mfherbst: Obviously not. The only thing that works right now is setup.py test.

maxscheurer: Yes, just wanted to add a note that we don't forget...

maxscheurer:
In `export_Tensor.cc` [here](https://github.com/adc-connect/adcc/blob/733d32442a476167d42cb22d7389c4a9bf31f561/libadcc/export_Tensor.cc#L38)
can you do
```C++
for (auto tpl : permutations) {
    if (py::isinstance<py::int_>(tpl)) {
      iterator_of_ints = true;
      break;
    }
  }
```

setup.py
```diff
@@ -295,7 +295,12 @@ def libadcc_extension():
             found_libtensor = True
             extra_compile_args.extend(cflags)
             extra_link_args.extend(libs)
-            runtime_library_dirs.extend(extract_library_dirs(libs))
+            if sys.platform == "darwin":
+                extra_link_args.append("-Wl,-rpath,@loader_path")
+                for path in extract_library_dirs(libs):
+                    extra_link_args.append(f"-Wl,-rpath,{path}")
+            else:
+                runtime_library_dirs.extend(extract_library_dirs(libs))
```
maxscheurer, line 303: Classic.... 🙃

mfherbst: Indeed, some tricks you know after a while. Let's hope that does it.

maxscheurer:
Can I make a commit here or do you have WIP?

mfherbst:
I won't work on it tonight and the next things I have are to get the C++ tests working. So that's mostly stuff in libadcc and setup.py.

setup.py
```diff
@@ -247,6 +247,10 @@ def libadcc_extension():
     define_macros = []
     search_system = True
 
+    if sys.platform == "darwin" and is_conda_build():
```
mfherbst, line 250: Are you sure this is the "right" thing to test for. Isn't this rather a clang-specific thing?

.github/workflows/deploy_conda.yaml
```diff
@@ -0,0 +1,56 @@
+name: Conda Deployment
```
mfherbst, line 1: I think it would be a good idea to merge this with the deploy workflow because they are essentially triggered under the same circumstances.

setup.py
maxscheurer: Kind of... AppleClang does not complain, and the only macOS compiler on `conda` is clang... I can add a note though.

.github/workflows/deploy_conda.yaml
maxscheurer: Yes, I think GHA can still run them as separate "jobs", right?

mfherbst: sure. You can even make the jobs depend on each other. That might allow us to only run the tests once.

maxscheurer:
We also should disable automatic download of libtensorlight in conda builds, such that we don't accidentally link against our "self-provided" version instead of the conda one.

setup.py
mfherbst: But what if I use clang on ubuntu?

mfherbst:
You can. Just provide a custom siteconfig.

maxscheurer:
But then I have to create an extra file just for the `conda` deployment?

mfherbst:
Sure. But I would write it to the home, that's more reliable (you don't depend on the working dir). It searches for `~/.adcc/siteconfig.py` or so ... check the setup.py.

setup.py
maxscheurer: Yes... so we need to get the compiler name and check for that 😄

maxscheurer: Mhm do you know by heart how to get the compiler name outside the extension class? 🤕

maxscheurer: Otherwise, I think one can just add the two flags directly to `extra_compile_args` without a compiler/platform distinction.

mfherbst: > how to get the compiler name outside the extension class?

No, but I guess you can always get it *after* the extension has been created (then the user can't override it, but at least things compile.

I think in order to decide what to do you will need to test if this problem also occurs in other circumstances. If it's conda-specific I don't mind the way you solved it now. If it's not then I think it should be solved more generally.

(On that note I think the siteconfig solution is not the best way to tackle it, better disable it in the setup.py)

maxscheurer: It seems to be conda+darwin-specific. Clang on arch compiles without complaining.

mfherbst: Ok cool. Then I'm happy as it is.

maxscheurer: Giving conda another run with the current configuration... Then I can also run some timings using the conda builds 😄

maxscheurer:
If you make commits here, can you leave the conda builds turned on for the time being? I need a fresh adcc on dev (because of the breaking lt allocator changes).

scripts/upload_to_anaconda.sh
```diff
@@ -45,4 +34,4 @@ conda install conda-build anaconda-client conda-verify --yes
 # conda config --append channels pyscf
 
 # Running build and deployment
-conda build conda -c defaults -c conda-forge --user adcc --token $ANACONDA_TOKEN --label $LABEL
+conda build conda -c adcc/label/dev -c defaults -c conda-forge --user adcc --token $ANACONDA_TOKEN --label $LABEL
```
maxscheurer, line 37: Can now be changed to `-c adcc`

maxscheurer:
Nice twist, conda, not wanting the channel name that is accidentally identical with a local directory 😂😂😂

mfherbst:
@maxscheurer Do you know how to exclude a file from LGTM. The siteconfig_example.py should be excluded from the analysis, I'd say.

maxscheurer:
I think configuring LGTM is really tough, but maybe it works via a [lgtm.yml](https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html) config file?

mfherbst:
> I think configuring LGTM is really tough, but maybe it works via a lgtm.yml config file?

Could be, I can't be asked to crawl through the docs at the moment though. Would you be willing to take care of that. I tried the C++ thing a few days back, but it did not work out of the box so I have not yet done it either.

mfherbst:
Ah ... ok that's surpirisingly easy.

maxscheurer:
> Ah ... ok that's surpirisingly easy.

I don't know if it works...

BTW did you test the `siteconfig.py` by specifying `libtensor_autoinstall` to some other path than the default? On my workstation the file is read, but the `libtensor_autoinstall` I set is being ignored.

mfherbst:
I know of that problem. I tested it, but before a refactor that broke it. It's now fixed.

maxscheurer:
> I know of that problem. I tested it, but before a refactor that broke it. It's now fixed in the C++ coverage commit.

I fixed a typo in the setup, so it's working now 👍

maxscheurer:
I bet LGTM for C++ will only work once this PR is merged into master, otherwise, I think there's no way to trigger another language on a PR...

mfherbst:
I guess we do it later than.

maxscheurer:
It seems like LGTM is just ignoring the `.lgtm.yml` file 😒
Hopefully it works when the PR is merged, but currently this feels like a waste of time.

mfherbst:
Well now that I record C++ coverage the python coverage is ignored ^^.

maxscheurer:
Question: Should we disable running the full test suite in conda builds? Instead, we could only run them during pip deployment and make the conda jobs depend on successful pip deployment...

Another suggestion is to just run tests on a single case in conda (e.g. `h2o_sto3g`) to make sure the built `adcc` in principle works.

If we support 4 Python versions (3.6-3.9, like Psi4) and 2 OS, we'd run the entire tests 8x during the deployment...

mfherbst:
> Another suggestion is to just run tests on a single case in conda (e.g. h2o_sto3g) to make sure the built adcc in principle works.

That plus the dependency on the pip job sounds like a good compromise. Also I suppose on the master / dev conda packages we can be more restrictive and e.g. exclude MacOS, which takes forever and perhaps some older python versions.

maxscheurer:
> we can be more restrictive and e.g. exclude MacOS, which takes forever and perhaps some older python versions.

You mean wrt the test runs?

mfherbst:
No with the actual dev deployments.

mfherbst:
How are the timings by the way?

maxscheurer:
> No with the actual dev deployments.

Okay, I don't have a good idea how to hack this in right now...

maxscheurer:
> How are the timings by the way?

Looking good, I'm running another large job right now, but I'd say we did not loose any performance on the way 🚀

maxscheurer:
Okay, I've adapted the workflow for conda to only deploy py38 to `dev`. I'd like to give this maybe another try and see if it works..

mfherbst:
Yeah sure. I'm not going to work on this today any more. Also feel free to take a round at cleaning up the docs if you want :smile:.

maxscheurer:
Okay, why not 😄 How should the "adccore" stuff be called from now on?

mfherbst:
Basically `libadcc`, but quite some of it will just need to go (e.g. overview of the structure of the code talks about the ADC equations being implemented in C++ and all that).

Like in the API documentation we distinguish libadcc and adccore and now libadcc has a c++ and a python part, so a bit of restructuring is needed there. Just come up with a sensible solution. The docs are pretty messy at the moment anyways. We really really should have a solid pass where we think about a more sensible structure. Probably less should be in static html and more in interactive examples anyway, but that I suppose is for another day.

maxscheurer:
I've made a pass at the docs, such that no more "wrong information" is present in the documentation 😄

maxscheurer:
Conda packaging on Linux now takes 12m, 22m on macOS 🍎 ... that's nice 🚀 🎸

maxscheurer (approved):

mfherbst:
Cool. I'll try to make anther pass at the docs today or tomorrow and than this can go in from my end as well.

maxscheurer (approved):
Huge effort over a long time that led to this PR 🥳
Thanks for the joint effort!

mfherbst:
Yes pretty much half a year, no? I'm really happy we're close to merging. Thanks for the teamwork! :fireworks:

Too bad the situation and all that but we should have some :beers: some time over this

maxscheurer:
I'm having a 🍺 now already, but yes! Hopefully adcc-🍻 soon...

mfherbst:
:smile:, I'll get one, too ;).

maxscheurer:
Can I bump to `v0.15.6`? 😄

maxscheurer:
Or you can do it as well, I'm upgrading macOS now... 🔥

mfherbst:
Oh yikes ... have fun! Would be good to see if our code then still works ;)

By the way: There is actually an issue with our coda deployment, because in the current workflow version 0.15.6 would be deployed both as 0.15.6.dev (to the dev label) and  0.15.6 (to the main label), which does not make super much sense. I'll try to manually cancel that this time, but that we should fix.

maxscheurer:
Okay, if you cannot cancel it, I can remove the files from dev label. I don't know how to fix it yet...

mfherbst:
No worries I cancelled, but yes that's a bit annoying. You can mark a complete job with an `if:` in GHA. I suppose that's what we have to do somehow (like if it's building on the master branch and got a tag, than don't run coda deployment only if it's building on the tag branch and got a tag). But don't ask me how that would work.

PR #114, opened 2020-12-11
Various small fixes

mfherbst:
**Note:** Coverage drops because previously the coverage of the C++ test files themselves was also reported, which does not make that much sense in my opinion.


mfherbst:
Feel free to merge @maxscheurer.

maxscheurer (approved):

PR #119, opened 2020-12-27
Python OneParticleOperator

maxscheurer:
- [x] working py-side implementation
- [x] docstrings
- [x] more tests of py-side functions
- [x] cleanup


maxscheurer:
Nice, this is the test suite nightmare - transition properties are wrong, but the tests of the low-level functions (`product_trace`) don't hit any failure 😡

adcc/OneParticleOperator.py
```diff
@@ -52,41 +57,183 @@ def __init__(self, spaces, is_symmetric=True):
...
-            ret.set_block(b, cpp_operator.block(b))
+        occs = sorted(self.mospaces.subspaces_occupied, reverse=True)
+        virts = sorted(self.mospaces.subspaces_virtual, reverse=True)
+        self.orbital_subspaces = occs + virts
+        if self.is_symmetric:
+            combs = list(
+                combinations_with_replacement(self.orbital_subspaces, r=2)
+            )
+        else:
+            combs = list(product(self.orbital_subspaces, repeat=2))
```
mfherbst, line 68: Can you explain in a comment why you are doing it exactly like this?

adcc/OneParticleOperator.py
```diff
@@ -52,41 +57,183 @@ def __init__(self, spaces, is_symmetric=True):
...
+        self.blocks = ["".join(com) for com in combs]
+        self._tensors = {}
+
+    @property
+    def shape(self):
+        """
+        Returns the shape tuple of the OneParticleOperator
+        """
+        size = sum(self.mospaces.n_orbs(ss) for ss in self.orbital_subspaces)
+        return (size, size)
```
mfherbst, line 78: Maybe cache this is `self.__space` ??

adcc/OneParticleOperator.py
```diff
@@ -52,41 +57,183 @@ def __init__(self, spaces, is_symmetric=True):
...
+        Checks if block is explicitly marked as zero block.
+        Returns False if the block does not exist.
+        """
+        if block not in self.blocks:
+            return False
+        return block not in self.blocks_nonzero
+
+    def block(self, block):
+        """
+        Returns tensor of the given block
```
mfherbst, line 105: Add: Does not create a block in case it is marked as a zero block. Use `__getitem__` for that purpose.

adcc/OneParticleOperator.py
```diff
@@ -52,41 +57,183 @@ def __init__(self, spaces, is_symmetric=True):
...
+            sym = libadcc.make_symmetry_operator(
+                self.mospaces, block, self.is_symmetric, "1"
+            )
+            self._tensors[block] = Tensor(sym)
+        return self._tensors[block]
+
+    def __setitem__(self, block, tensor):
+        self.set_block(block, tensor)
+
+    def set_block(self, block, tensor):
```
mfherbst, line 127: Do we still need the `set_block` function at all? Maybe we should make it deprecated and remove it soon?

adcc/OneParticleOperator.py
maxscheurer: What exactly? The number of MOs per subspace?

adcc/OneParticleOperator.py
maxscheurer: Yes, I completely agree. Since everything is on the py-side now, no reason to keep it.

adcc/OneParticleOperator.py
maxscheurer: Sure!

adcc/OneParticleOperator.py
maxscheurer: Maybe I'm wrong, but isn't this just equivalent to:
```Python
size = self.mospaces.n_orbs("f")
return (size, size)
```
?

maxscheurer: Yeah it's indeed the case... Resolved 😄

adcc/OneParticleOperator.py
```diff
@@ -52,41 +58,201 @@ def __init__(self, spaces, is_symmetric=True):
...
+        if block not in self._tensors:
+            sym = libadcc.make_symmetry_operator(
+                self.mospaces, block, self.is_symmetric, "1"
+            )
+            self._tensors[block] = Tensor(sym)
+        return self._tensors[block]
+
+    def __getattr__(self, attr):
+        from . import block as b
+        return self.__getitem__(b.__getattr__(attr))
```
mfherbst, line 135: Nice idea, but if you introduce it, you should also use it: Can you replace old usage of the block strings for this "new" syntax in adcc? There are plenty of places in the equations where one can relplace the `p0_oo` aliases by a single `p0` alias now.

adcc/adc_pp/matrix.py
```diff
@@ -411,9 +411,9 @@ def adc3_i1(hf, mp, intermediates):
         p0_ov = intermediates.cvs_p0_ov
         p0_vv = intermediates.cvs_p0_vv
     else:
-        p0_ov = mp.mp2_diffdm[b.ov]
-        p0_oo = mp.mp2_diffdm[b.oo]
-        p0_vv = mp.mp2_diffdm[b.vv]
+        p0_ov = mp.mp2_diffdm.ov
+        p0_oo = mp.mp2_diffdm.oo
+        p0_vv = mp.mp2_diffdm.vv
```
mfherbst, line 416: Just `p0 = mp.mp2_diffdm`?

mfherbst (commented):
Looking really good now.

adcc/adc_pp/matrix.py
maxscheurer: Okay, but then I need to modify the CVS intermediate stuff above, too 👍

maxscheurer:
The PR can be squash-merged after approval.

I've changed the structure of the CVS ground state density intermediates a little bit. From a performance point of view, this should be no problem because the density is cached and if one block is required, all blocks are required, sooner or later...

adcc/OneParticleOperator.py
```diff
@@ -52,41 +58,210 @@ def __init__(self, spaces, is_symmetric=True):
...
+        self.__setitem__(block, tensor)
+
+    def set_zero_block(self, block):
+        """
+        Set a given block as zero block
+        """
+        if block not in self.blocks:
+            raise KeyError(f"Invalid block {block} set as zero block. "
+                           f"Available blocks are: {self.blocks}.")
+        self._tensors.pop(block)
```
mfherbst, line 176: I wonder if we should have the special syntax
```python
operator.oo = 0
```
for this purpose or is that too much "magic"? Alternative is of course `operator.set_zero_block("o1o1")`.

mfherbst: Actually I think that `set_zero_block` is used nowhere, so then we can just leave it as is, I guess?

mfherbst (approved):

adcc/OneParticleOperator.py
maxscheurer: Yes, let's leave it as is. The `dm.oo = 0` syntax would cause confusion, because you'd expect to zero the tensor elements, but not delete the entire block 👍🏻

PR #122, opened 2020-12-30
Test conda build with dev lt

maxscheurer:
Just a dummy PR to test the dev-deployed lt version (fixing current conda build issues on macOS hopefully)
See [here](https://github.com/adc-connect/libtensor/pull/11) for details.


maxscheurer:
Ok, it works. So we can release a new libtensorlight with the fixed conda setup 😄

PR #123, opened 2021-01-06
Updates for solvent models (new user interface, new schemes)

maxscheurer:
Preliminary PR, just to stash some work...

- [x] tests
- [x] add references
- [x] decide on how to select PE-related features from user perspective: pt. corrections (at least ptLR) must be disabled when using the coupling scheme
- [x] changes in `run_adc` and wrappers
- [x] adapt examples
- [x] add docs


adcc/PeAdcMatrix.py
```diff
@@ -0,0 +1,16 @@
+from adcc import AdcMatrix
+from .OneParticleOperator import OneParticleOperator
+
+
+class PeAdcMatrix(AdcMatrix):
```
maxscheurer, line 5: I wonder if this would best be integrated directly with `AdcMatrix` instead of creating a subclass... For "derived" computations, e.g., response properties, this would certainly make things easier.

maxscheurer:
The only reason this PR is not moving forward is that I've not been able to come up with a user-friendly solution for integrating this feature properly.

The following feature scenarios are possible:

1. "pt scheme": purely perturbative energy corrections (currently automatically computed and incorporated into `excitation_energy` if a PE ground state is present. Uncorrected excitation energies (i.e., the "actual" eigenvalues") are in `excitation_energy_uncorrected`
2. "postSCF scheme": coupling to the environment via an additional (excitonic) coupling term (via CIS-like transition density). In this case, the `excitation_energy_uncorrected` values already contain the "linear response" contribution, so the ptLR does not make sense anymore. The ptSS still makes sense. The eigenvalues of this modified ADC matrix however still correspond to `excitation_energy_uncorrected`. At the moment, both ptLR and ptSS corrections are always computed, also for the "postSCF scheme", which makes little sense.
3. "ground state": don't couple at all in ADC, just through the PE-HF ground state. This is currently not possible if a PE-HF ground state is given.

The same problematic will hold for other solvent models (PCM for example). So the following questions arise:
1. Which of the above methods should be the default? (I'd say 1. because for excitation energies it offers the best trade-off between performance and accuracy)
2. How does the user select the coupling scheme? This must be possible both from `run_adc` and via "matrix construction".
3. How do we implement the "logic" of disabling corrections etc. if a certain scheme is chosen. This is the most annoying part, because we'd like to provide transparent result to the user. 


Just wanted to leave this here for discussion, because I have not made any progress alone so far...

adcc/PeAdcMatrix.py
maxscheurer: I now think the best strategy is to exploit the very nice design strategy of `AdcBlock` to enable these extra terms. In this manner, the contributions are "injected" at the lowest level, such that basically no errors can be made in follow-up work (e.g. singles-space solver etc.), because the matrix apply will always be correct.

The current implementation resides in the test files and I'm not yet sure how to implement it for real. I thought about
adding a `custom_block` argument to the `AdcMatrix` ctor, where one could pass extra `AdcBlock`s that are
added to `self.blocks_ph`.

maxscheurer:
I hate `clang-format`, why is it changing this one file anyways? I've added the `.clang-format` file we used to have in `adccore`, but it does not seem to care. I cannot reproduce this locally, however 😠

mfherbst:
Different clang-format versions? Sometimes this happens then.

maxscheurer:
> Different clang-format versions? Sometimes this happens then.

I've tried both v11 and v9 on CI, both giving the same `diff`... I can make the change such that the CI runs without failure.

mfherbst:
I would not get too much held up with this. either make the change or disable the relevant line from formatting or sth like that.

maxscheurer:
I think we have quite the functionality we want now! If you got a couple of minutes, @mfherbst, it would be nice if you could give it a brief look. I'm now working on refactoring the user-facing changes in `run_adc` and some convenience wrappers, and then we can do refactoring etc. 🚀

maxscheurer:
Everything seems to be operational again 🎉

adcc/AdcMatrix.py
```diff
@@ -32,6 +32,48 @@
...
+            block = block_fun(
+                self.reference_state, self.ground_state, self.intermediates
+            )
+            self.blocks[space] = block
+
+    def __iadd__(self, other):
+        # disable in-place addition getting accidentally
+        # short-circuited to AdcMatrix.__radd__
+        raise NotImplementedError("In-place addition not implemented "
+                                  " for AdcExtraTerm")
```
mfherbst, line 74: This is why Julia is so much nicer ...

adcc/AdcMatrix.py
```diff
@@ -32,6 +32,48 @@
 from .AmplitudeVector import AmplitudeVector
 
 
+class AdcExtraTerm:
+    # NOTE: currently requires the matrix
+    # to allow for block construction with the same
+    # interface as the usual adc_pp, i.e.,
+    # reference_state, ground_state, intermediates
+    def __init__(self, matrix, blocks):
```
mfherbst, line 40: Why do you need to pass the matrix? If you pass ground state it should be sufficient, shouldn't it? (ok save the intermediates, but I think we should have an interface that does not require the matrix as well. It somehow feels "too much".

adcc/AdcMatrix.py
```diff
@@ -32,6 +32,48 @@
 from .AmplitudeVector import AmplitudeVector
 
 
+class AdcExtraTerm:
+    # NOTE: currently requires the matrix
+    # to allow for block construction with the same
+    # interface as the usual adc_pp, i.e.,
+    # reference_state, ground_state, intermediates
```
mfherbst, line 39: i don't understand this

adcc/AdcMatrix.py
```diff
@@ -112,19 +155,72 @@ def __init__(self, method, hf_or_mp, block_orders=None, intermediates=None):
...
+
+        Parameters
+        ----------
+        other : AdcExtraTerm
+            the extra term to be added
+        """
+        if isinstance(other, list):
+            for k in other:
+                self += k
+            return self
```
mfherbst, line 182: Don't support this. I don't see why this would be useful.

adcc/AdcMatrix.py
```diff
@@ -112,19 +155,72 @@ def __init__(self, method, hf_or_mp, block_orders=None, intermediates=None):
...
+        if not isinstance(other, AdcExtraTerm):
+            raise TypeError("Can only add AdcExtra term or"
+                            " lists thereof to AdcMatrix.")
```
mfherbst, line 185: That's not the right way to do this in python. if the types don't match throw NotImplemented. See docs on python operator overloading.

adcc/adc_pp/matrix.py
```diff
@@ -39,15 +38,53 @@
...
+            expression to the diagonal of the ADC matrix from this block
+        """
+        if not callable(apply):
+            raise TypeError("apply needs to be callable.")
+        if not isinstance(diagonal, (AmplitudeVector, int, float)):
+            raise TypeError("diagonal needs to be an AmplitudeVector or a number.")
+        self._applies = [apply]
+        self._diagonals = [diagonal]
+
+    def add_block(self, other):
```
mfherbst, line 62: This is just +=

adcc/AdcMatrix.py
```diff
@@ -112,19 +155,72 @@ def __init__(self, method, hf_or_mp, block_orders=None, intermediates=None):
...
+        ----------
+        other : AdcExtraTerm
+            the extra term to be added
+
+        Returns
+        -------
+        AdcMatrix
+            a copy of the AdcMatrix with the extra term added
+        """
+        # NOTE: re-computes the (expensive) diagonal...
```
mfherbst, line 214: Yes, we should have a way to short-circuit around that.

adcc/adc_pp/matrix.py
```diff
@@ -39,15 +38,53 @@
...
 
-#
-# Dispatch routine
-#
-"""
-`apply` is a function mapping an AmplitudeVector to the contribution of this
-block to the result of applying the ADC matrix. `diagonal` is an `AmplitudeVector`
-containing the expression to the diagonal of the ADC matrix from this block.
-"""
-AdcBlock = namedtuple("AdcBlock", ["apply", "diagonal"])
```
mfherbst, line 50: Whenever I add 60 lines of code I ask myself: Is it worth it. And here I am not so sure. The only thing you really need is ` return sum(app(ampl) for app in self._applies)` as the diagonal summation is done by the AdcMatrix class later anyway. And for the sum you could also just generate a `lambda ampl: sum(app(ampl) for app in self._applies)` and use that instead of the already existing apply function inside the AdcMatrix for the patched block. Given that this version is not much slower (which we should test) I would prefer that, because that keeps also the "block logic" in the AdcMatrix where it belongs and shortens this back to 5 lines instead of 65.

adcc/AdcMatrix.py
```diff
@@ -112,19 +155,72 @@ def __init__(self, method, hf_or_mp, block_orders=None, intermediates=None):
...
+                             " AdcMatrix that already exist.")
+        for sp in other.blocks:
+            ob = other.blocks[sp]
+            self.blocks_ph[sp].add_block(ob)
+        diag = sum(bl.diagonal for bl in other.blocks.values()
+                   if bl.diagonal)
+        # iadd does not work with numbers
+        self.__diagonal = self.__diagonal + diag
+        self.__diagonal.evaluate()
+        self.extra_terms.append(other)
```
mfherbst, line 197: This feels to complicated and indirect. See my comment in the AdcBlock class below.

adcc/ExcitedStates.py
```diff
@@ -38,6 +38,19 @@
 from .FormatDominantElements import FormatDominantElements
 
 
+class EnergyCorrection:
+    def __init__(self, name, function, description=None):
+        assert isinstance(name, str)
+        assert callable(function)
```
mfherbst, line 44: Obviously should be replaced by better error handling :smile:.

adcc/OperatorIntegrals.py
```diff
@@ -137,6 +137,26 @@ def nabla(self):
         """
         return self.import_dipole_like_operator("nabla", is_symmetric=False)
 
+    @property
+    def density_dependent_operators(self):
+        if not hasattr(self.provider_ao, "density_dependent_operators"):
+            return {}
```
mfherbst, line 143: No need. Done exactly the same by the rest of the function.

adcc/OperatorIntegrals.py
```diff
@@ -137,6 +137,26 @@ def nabla(self):
         """
         return self.import_dipole_like_operator("nabla", is_symmetric=False)
 
+    @property
+    def density_dependent_operators(self):
```
mfherbst, line 141: Should bikeshed a better name maybe ...

adcc/ReferenceState.py
```diff
@@ -155,7 +155,8 @@ def __init__(self, hfdata, core_orbitals=None, frozen_core=None,
             self.orbital_coefficients, self.conv_tol
         )
 
-        for name in ["excitation_energy_corrections"]:
+        self.solvent = None  # no solvent attached by default
+        for name in ["excitation_energy_corrections", "solvent"]:
```
mfherbst, line 159: solvent -> solvation ?

adcc/adc_pp/solvent.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
```
mfherbst, line 1: rename to solvation?

adcc/adc_pp/solvent.py
```diff
@@ -0,0 +1,42 @@
...
+    coupling density as described in 10.1021/ct300763v, eq 63.
+    Since the contribution depends on the input amplitude itself,
+    a diagonal term cannot be formulated.
+    """
+    op = hf.operators
+
+    def apply(ampl):
+        tdm = OneParticleOperator(mp, is_symmetric=False)
+        tdm.vo = ampl.ph.transpose()
+        vpe = op.density_dependent_operators["pe_induction_elec"](tdm)
```
mfherbst, line 40: Here I wonder why this is not directly op.pe_induction_elec. It does not really matter that this does not return a matrix but instead returns a function, right? It would just feel more natural.

adcc/backends/psi4.py
```diff
@@ -53,6 +54,17 @@ def magnetic_dipole(self):
     def nabla(self):
         return [-1.0 * np.asarray(comp) for comp in self.mints.ao_nabla()]
 
+    @property
+    def density_dependent_operators(self):
+        ret = {}
+        if hasattr(self.wfn, "pe_state"):
+            ret["pe_induction_elec"] = lambda dm: \
```
mfherbst, line 61: IMHO if you need to break a lambda make it a def

adcc/backends/test_backends_polembed.py
```diff
@@ -82,3 +88,53 @@ def template_pe_formaldehyde(self, basis, method, backend):
             state.pe_ptlr_correction,
             atol=1e-5
         )
+
+    def template_pe_coupling_formaldehyde(self, basis, method, backend):
+        if method != "adc2":
+            pytest.skip("")
```
mfherbst, line 94: some message would be good ;)

adcc/backends/test_backends_polembed.py
```diff
@@ -82,3 +88,53 @@ def template_pe_formaldehyde(self, basis, method, backend):
...
+        tm_result = tmole_data[basename]
+        pe_options = {"potfile": pe_potentials["fa_6w"]}
+        scfres = cached_backend_hf(backend, "formaldehyde", basis,
+                                   pe_options=pe_options)
+        assert_allclose(scfres.energy_scf, tm_result["energy_scf"], atol=1e-8)
+
+        matrix = adcc.AdcMatrix(method, scfres)
+        solvent = AdcExtraTerm(matrix, {'ph_ph': block_ph_ph_0_pe})
+
+        with pytest.raises(NotImplementedError):
```
mfherbst, line 105: If you do operators properly this should raise a TypeError automatically.

adcc/workflow.py
```diff
@@ -47,6 +47,7 @@ def run_adc(data_or_matrix, n_states=None, kind="any", conv_tol=None,
             n_guesses_doubles=None, output=sys.stdout, core_orbitals=None,
             frozen_core=None, frozen_virtual=None, method=None,
             n_singlets=None, n_triplets=None, n_spin_flip=None,
+            solvent_scheme=None,
```
mfherbst, line 50: solvation_scheme?
or just solvation?

adcc/workflow.py
```diff
@@ -191,13 +192,22 @@ def run_adc(data_or_matrix, n_states=None, kind="any", conv_tol=None,
     if eigensolver is None:
         eigensolver = "davidson"
 
+    # Setup solvent coupling terms and energy corrections
+    solvent_matrix_terms, solvent_energy_corrections = \
+        setup_solvent(matrix, solvent_scheme)
```
mfherbst, line 197: in these cases I prefer
```suggestion
    ret = setup_solvent(matrix, solvent_scheme)
    solvent_matrix_terms, solvent_energy_corrections = ret
```

adcc/workflow.py
```diff
@@ -495,3 +505,85 @@ def setup_solver_printing(solmethod_name, matrix, kind, default_print,
         def inner_callback(state, identifier):
             default_print(state, identifier, output)
         return inner_callback
+
+
+def setup_solvent(matrix, solvent_scheme):
```
mfherbst, line 510: solvation

adcc/workflow.py
```diff
@@ -495,3 +505,85 @@ def setup_solver_printing(solmethod_name, matrix, kind, default_print,
...
+    """
+    Setup solvent matrix terms and/or energy corrections.
+    Internal function called from run_adc.
+    """
+    # TODO: move someplace meaningful & reasonable...
```
mfherbst, line 515: Yes let's discuss ...

mfherbst (commented):
Some suggestions ...

adcc/AdcMatrix.py
maxscheurer: Yes... I agree haha

mfherbst: oh also this is again a case where you raise NotImplemented and not an error message

adcc/AdcMatrix.py
maxscheurer: I had the same thought in the beginning, but after having worked with it a couple of times now, it somehow feels right: "Why would you construct an extra term without having a matrix that it belongs to in the first place?" is my current take on this issue.

If you think it's too much, we can change the interface, but then we need to adapt the underlying block construction routine accordingly (which is then inconsistent wrt the normal adc_pp routines).

mfherbst: ok I see. I have not checked that closely. Let's leave it and consider the other aspects first.

adcc/adc_pp/solvent.py
maxscheurer: Like `op.pe_induction_elec(tdm)`? I'm okay with this 👍

mfherbst: Yes.

adcc/ReferenceState.py
maxscheurer: Maybe `embedding` or `environment` would be even better... I'd not call a protein modeled with PE "solvent" 😄

adcc/AdcMatrix.py
maxscheurer: Done.

adcc/AdcMatrix.py
maxscheurer: Done.

adcc/AdcMatrix.py
maxscheurer: Done.

adcc/backends/test_backends_polembed.py
maxscheurer: Done.

adcc/workflow.py
maxscheurer: Done.

adcc/AdcMatrix.py
maxscheurer: Ah I remember why I did this... to do "nothing" in case an empty list is passed in `workflow`, but I think it's nicer with a simple `if` in `workflow` 😄

adcc/backends/test_backends_polembed.py
maxscheurer: That's right 😄 Done.

maxscheurer: Damn I hate Python... now having the "correct" implementation, i.e., returning `NotImplemented`, Python tries to be super clever and inverts the argument order automatically to `matrix += solvent` 🤯
I don't know how one can disable this 👎

adcc/workflow.py
maxscheurer: My idea was a table in the docs with the long explanations and references to papers; plus a short description in the function documentation that one can view via `adcc.workflow.setup_solvent?`

What do you think?

adcc/adc_pp/matrix.py
maxscheurer: I'm not sure I understood everything...
Do you want to revert to `namedtuple`, if yes, where would `self._applies` come from?

adcc/OperatorIntegrals.py
maxscheurer: Now obsolete.

adcc/OperatorIntegrals.py
maxscheurer: Now obsolete.

adcc/ExcitedStates.py
maxscheurer: Done.

adcc/adc_pp/matrix.py
mfherbst: Nowhere. Instead of "adding the block" you form a lambda inside the AdcMatrix class or so that does the addition between the applies.

adcc/ReferenceState.py
mfherbst: Ok for both embedding or environment.

adcc/backends/psi4.py
maxscheurer: Done.

adcc/adc_pp/matrix.py
maxscheurer: I'm still a bit uncertain...
I tried the following now
```Python
for sp in other.blocks:
    original_apply = self.blocks_ph[sp].apply
    def patched_apply(ampl):
        return sum(
            app(ampl)
            for app in (original_apply, other.blocks[sp].apply)
        )
    self.blocks_ph[sp].apply = patched_apply
```
if that's what you meant... Problem is one cannot re-set the attributes of namedtuple.

adcc/workflow.py
mfherbst: Sounds perfect. Keywords should be short and descriptive. Perhaps they should actually be a class / named tuple (this way you can take optional "arguments" in case that might be necessary for more involved models, like how many times you iterate a response or which order of perturbation theory or sth like that).

adcc/adc_pp/matrix.py
mfherbst: Ah crap ... I forgot about that. But then change the named tuple to a simple class with just two attributes?

mfherbst: or maybe avoid storing the named tuple in the AdcMatrix and intstead only store the apply functions (and the diagonal is munched into the precomputed thing anyway)?

adcc/workflow.py
maxscheurer: What exactly should be a class or namedtuple? The collection of all keywords/options or one instance of the namedtuple for each scheme?
I think for now one could also stick with the scheme "names" and if we need more advanced settings in the future, these can be enabled via `**kwargs`, which is the usual way in `adcc` I guess.

mfherbst: Nope what I meant was instead of `solvent_model="ptss"` you do `solvent_model=ptss` which is an actually `ptss` class or named tuple. In that way you can later do `solvent_model=ptss(order=5)` in case we ever implemented 5-th order corrections.

maxscheurer: For me this feels too complicated for now... I think what you want could be introduced without breaking changes if we ever need this in the future, i.e., allow for either `solvent_model=["ptss", "ptlr"]` which chooses pre-defined configurations or provides full flexibility via `solvent_model=[ptss(order=100), ptlr(order=10)]`.

I don't know... I have to think about it. Probably one would not want to write more `import` statements in the Python script to run `adcc` just to be able to use the corrections.

mfherbst: Yes that's true. As long as there is a reasonable way to later include this I'm ok. I just want to avoid that we have 10000 keywords in run_adc on the one hand and 1000 possible combinations of solvent keywords on the other, that's why I threw that comment in. I agree that for now supporting a few strings is ok. But a list of strings already feels like something more involved, where a class with sensible arguments could be a more pythonic alternative, which we should consider for a sec.

maxscheurer: Difficult design choice... I have no clue which alternative is best in the long run.
If we go for a class/namedtuple to configure everything, how would that look?
This is the same ol' "struct of arrays" or "array of structs" problem... 

One scenario:
```Python
config = SolventConfig(ptss=True, ptlr=True)
adcc.adc2(scfres, n_singlets=5, solvent_config=config)
```

Second scenario:
```Python
config = [
    SolventConfig("ptss"), SolventConfig("ptlr")
]
```
or whatever.

The only thing I don't like about the current solution (string or list of strings) is that it's pretty
unclear what you are allowed to do. I mean the code will raise Exceptions if you select an invalid combination, but
it's a bit weird.

I mean you still have full flexibility and can compose anything you like via the new classes `EnergyCorrection` and
`AdcExtraTerm`.

adcc/adc_pp/solvent.py
maxscheurer: Done.

adcc/workflow.py
mfherbst: I would do neither. I'd do
```solvent_config=[ptss, ptlr]```
which could also accept a list of strings ... but then we are sort of back where we started. I don't know. I'm just ultimately not happy about strings because it gives infinite possibilities (also typos etc.) and it does not suggest what is valid and what not at all. Flags (like `(ptss=True, ptlr=True)` are typically much more meaningful and better for tab completion.

Comment (maxscheurer) [reply adcc/workflow.py:516]: Okay probably `SolventConfig(ptss=True, ptlr=True)` or so is the easiest and cleanest way to handle it.
The class could also take care of basic validation, i.e., which combinations are allowed and which not in general, e.g.,
`SolventConfig(ptss=True, ptlr=True, postscf=True)` should throw an error.
`SolventConfig(ptss=True, hf=True)` as well. So the class attributes don't suggest which *combinations* are valid,
but at least which keywords...


If we decide to go this way, we should disable the list of strings stuff.

Comment (mfherbst) [reply adcc/workflow.py:516]: Yes that's true. Otherwise it is too much. Also we should have a sane default and `SolventConfig` is not the best name ...

Comment (maxscheurer) [reply adcc/workflow.py:516]: The "sane default" is the only thing that is still bugging me... maybe I'll have better ideas tomorrow 🙄

Comment (maxscheurer) [reply adcc/workflow.py:516]: I have to be frank... I don't like either solution for this problem. Maybe the solvent stuff needs to become "advanced" use, i.e., you need to do everything manually...

The reason why I'm saying this is:

Let's say `cfg = SolventConfig()` is by default `SolventConfig(hf=True)`, i.e., coupling only through "solvated orbitals" nothing further. Then `cfg = SolventConfig(ptss=True)` would be invalid (HF + ptSS is mutually exclusive), and one would need to specify `cfg = SolventConfig(hf=False, ptss=True)`, which is a lot of code to type for
what it achieves.
`cfg = SolventConfig(hf=False)` would be invalid later on *if* you have an SCF result w/ solvent attached. And so on and so forth... the whole logic of this in itself is not user-friendly at all.

Comment (mfherbst) [reply adcc/workflow.py:516]: Perhaps the `solvent` argument takes either `None` (auto-decided), `False` (HF only) or `Correction(ptss=True, ptlr=True)` or something along those lines?

Comment (maxscheurer) [reply adcc/workflow.py:516]: I'm not sure... different behavior for `None` and `False` does not seem a good choice, and adding a class `HfOnly` that basically does not do anything except "storing" the information just seems too much.

Maybe discuss via Skype?

Comment (mfherbst) [reply adcc/workflow.py:516]: Ok, but not now. At 13?

File: adcc/adc_pp/matrix.py (hunk 9 above)
Comment (maxscheurer) [reply adcc/adc_pp/matrix.py]: Done.

File: adcc/AdcMatrix.py (hunk 10 above)
Comment (maxscheurer) [reply adcc/AdcMatrix.py]: Changed as per comment above.

File: adcc/workflow.py (hunk 20 above)
Comment (maxscheurer) [reply adcc/workflow.py]: Done.

File: adcc/ReferenceState.py (hunk 14 above)
Comment (maxscheurer) [reply adcc/ReferenceState.py]: Done.

File: adcc/adc_pp/solvent.py (hunk 15 above)
Comment (maxscheurer) [reply adcc/adc_pp/solvent.py]: Done.

**Comment by maxscheurer** [comment]
I've added some of the stuff we discussed 😄 
The interface currently has a bit more flexibility, but it feels good from a user perspective imho.
Let me know what you think @mfherbst 🧐

**Comment by maxscheurer** [comment]
Docs now have a section on performing PE calculations 🚀 It's all coming together... 😄

**Comment by mfherbst** [comment]
@maxscheurer Noted it. Not sure when I get to it. Have a deadline next week.

**Comment by maxscheurer** [comment]
Sure, I didn't want to be too annoying, sorry 😄 Then I'll start working on sth else and we move this forward when you have more time.

**Comment by mfherbst** [comment]
Don't worry ... I just wanted you to know.

File: adcc/AdcMatrix.py [hunk 24]
```diff
@@ -112,19 +148,84 @@ def __init__(self, method, hf_or_mp, block_orders=None, intermediates=None):
...
-            self.__diagonal.evaluate()
+            # TODO Rename to self.block in 0.16.0
+            self.blocks_ph = {bl: blocks[bl].apply for bl in blocks}
+            if diagonal_precomputed:
+                if not isinstance(diagonal_precomputed, AmplitudeVector):
+                    raise TypeError("diagonal_precomputed needs to be"
+                                    " an AmplitudeVector.")
+                if diagonal_precomputed.needs_evaluation:
+                    raise ValueError("diagonal_precomputed must already"
+                                     " be evaluated.")
```
mfherbst, line 167: Move to the beginning where other validation is done.

adcc/AdcMatrix.py
```diff
@@ -112,19 +148,84 @@ def __init__(self, method, hf_or_mp, block_orders=None, intermediates=None):
...
+        for sp in other.blocks:
+            orig_app = self.blocks_ph[sp]
+            other_app = other.blocks[sp].apply
+
+            def patched_apply(ampl, original=orig_app, other=other_app):
+                return sum(app(ampl) for app in (original, other))
+            self.blocks_ph[sp] = patched_apply
+        other_diagonal = sum(bl.diagonal for bl in other.blocks.values()
+                             if bl.diagonal)
+        # iadd does not work with numbers
```
mfherbst, line 197: That comment should be at the place where you return NotImplemented.

adcc/ExcitedStates.py
```diff
@@ -153,33 +178,45 @@ def __init__(self, data, method=None, property_method=None,
...
-                                   for exci in self.excitations])
-            setattr(self, key, correction)
-            self._excitation_energy += correction
+        self._excitation_energy_corrections = []
+        # copy energy corrections if possible
+        # and avoids re-computation of the corrections
+        if hasattr(data, "_excitation_energy_corrections"):
+            for eec in data._excitation_energy_corrections:
+                correction_energy = getattr(data, eec.name)
+                setattr(self, eec.name, correction_energy)
```
mfherbst, line 189: I think the pythonic way is to provoke a key error and catch it to do the caching.

adcc/ExcitedStates.py
```diff
@@ -153,33 +178,45 @@ def __init__(self, data, method=None, property_method=None,
...
+
+    def __iadd__(self, other):
+        if isinstance(other, EnergyCorrection):
+            self.__add_energy_correction(other)
+        elif isinstance(other, list):
+            for k in other:
+                self += k
+        else:
+            raise TypeError("Can only add EnergyCorrection (or list"
+                            " of EnergyCorrection) to"
```
mfherbst, line 212: use NotImplemented mechanism

docs/calculations.rst
```diff
@@ -685,6 +687,95 @@ In fact all three may be combined jointly with any available ADC method,
...
+
+   # now model the solvent through linear-response coupling
+   # in the ADC matrix, re-using the matrix from previous run.
+   # This will modify state_pt.matrix
+   state_lr = adcc.run_adc(state_pt.matrix, n_singlets=5, conv_tol=1e-5,
+                           environment='linear_response')
+
+   print(state_pt.describe())
+   print(state_lr.describe())
+
```
mfherbst, line 777: Could you show the output in the docs?

adcc/ExcitedStates.py
```diff
@@ -338,6 +375,16 @@ def describe(self, oscillator_strengths=True, rotatory_strengths=False,
             text += body.format(i=i, ene=self.excitation_energy[i],
                                 ev=self.excitation_energy[i] * eV, **fields)
         text += separator + "\n"
+        if len(self._excitation_energy_corrections):
+            head_corr = "|  excitation energy corrections included:"
+            text += head_corr
+            nspace = len(separator) - len(head_corr) - 1
+            text += nspace * " " + "|\n"
```
mfherbst, line 382: Could you show an example output? I'm curious what it looks like.

adcc/backends/pyscf.py
```diff
@@ -117,12 +128,30 @@ def pe_energy(self, dm, elec_only=True):
...
+                                                  elec_only=True)
+            )
+            ptss = EnergyCorrection(
+                "pe_ptss_correction",
+                lambda view: self.pe_energy(view.state_diffdm_ao,
+                                            elec_only=True)
+            )
+            # NOTE: I don't like the duplicate 'name',
+            # but this way it's easier to exctract the corrections
+            # directly
```
mfherbst, line 144: This comment should not really go here, but somewhere else.

I see the point, but can't you first add all the corrections to a list and then transform the list to a dict before returning? This way one cannot forget to synchronise the names. Generally duplicating info is not the problem, it's needing to specify it twice which is.

adcc/test_AdcMatrix.py
```diff
@@ -214,6 +216,72 @@ def test_matvec_adc2(self):
...
+        with pytest.raises(ValueError):
+            adcc.AdcMatrix("adc2", ground_state,
+                           diagonal_precomputed=matrix.diagonal() + 42)
+
+        shift = -0.3
+        shifted = AdcMatrixShifted(matrix, shift)
+        # TODO: need to do this to differentiate between
+        # diagonals for ph and pphh
+        # if we just pass numbers, i.e., shift
+        # we get 2*shift on the diagonal :(
```
mfherbst, line 238: I don't get this comment. Can you explain more?

adcc/workflow.py
```diff
@@ -495,3 +509,73 @@ def setup_solver_printing(solmethod_name, matrix, kind, default_print,
...
+    Setup environment matrix terms and/or energy corrections.
+    Internal function called from run_adc.
+    """
+    valid_envs = ["ptss", "ptlr", "linear_response"]
+    hf = matrix.reference_state
+    if hf.environment and environment is None:
+        raise InputError(
+            "Environment found in reference state, but no environment"
+            " configuration specified. Please select from the following"
+            f" schemes: {valid_envs}."
```
mfherbst, line 525: But `False`is also possible, which you don't tell the user here. That's confusing.

adcc/workflow.py
```diff
@@ -495,3 +509,73 @@ def setup_solver_printing(solmethod_name, matrix, kind, default_print,
...
+        )
+    elif environment and not hf.environment:
+        raise InputError(
+            "Environment specified, but no environment"
+            " was found in reference state."
+        )
+    elif not hf.environment:
+        environment = False
```
mfherbst, line 533: environment = {}

adcc/workflow.py
```diff
@@ -495,3 +509,73 @@ def setup_solver_printing(solmethod_name, matrix, kind, default_print,
...
+
+    if isinstance(environment, bool):
+        environment = {"ptss": True, "ptlr": True} if environment else {}
+    elif isinstance(environment, list):
+        environment = {k: True for k in environment}
+    elif isinstance(environment, str):
+        environment = {environment: True}
+    elif not isinstance(environment, dict):
```
mfherbst, line 541: I think I would allow that. Not document it, but allow it. It makes your life easier for normalising the parameters.

adcc/workflow.py
```diff
@@ -495,3 +509,73 @@ def setup_solver_printing(solmethod_name, matrix, kind, default_print,
...
+
+    for pt in ["ptss", "ptlr"]:
+        if not environment.get(pt, False):
+            continue
+        hf_corr = hf.excitation_energy_corrections
+        eec_key = f"{hf.environment}_{pt}_correction"
+        if eec_key not in hf_corr:
+            raise ValueError(f"{pt} correction requested, but could not find"
+                             f" the needed function {eec_key} in"
+                             f" reference state from backend {hf.backend}.")
```
mfherbst, line 569: Show what is available? Or maybe too much?

mfherbst (commented):
Sorry that my review took so long. Nice work. I have a few small things, but I think it's almost good to go.

adcc/test_AdcMatrix.py
maxscheurer: Yeah sure, it's because the diagonal is constructed by
```
other_diagonal = sum(bl.diagonal for bl in other.blocks.values() if bl.diagonal)
```

So, if you want the diagonal contribution from `__shift_ph` and `__shift_pphh`, both have in theory `shift` in the `diagonal` attribute of their `AdcBlock`. But if you don't represent this as an `AmplitudeVector`, the above code will
result in `other_diagonal = shift + shift`, i.e., just adding the two floats.

mfherbst: Ah ok now I get it. The issue I was struggeling is the structure of the test. That comment actually refers to something you do much down the line. Could you move the "raises" tests commonly to the beginning or the end to keep the successful test code closer together?

adcc/workflow.py
maxscheurer: I guess it's a bit too much for this print-out. 😄

adcc/AdcMatrix.py
maxscheurer: Sure, it's the `AmplitudeVector`'s `iadd`, but I can add the comment there.

adcc/ExcitedStates.py
maxscheurer: I'm not sure where a key error could/should be provoked in these lines?

adcc/ExcitedStates.py
maxscheurer: Will show up in the docs 😄

adcc/backends/pyscf.py
maxscheurer: Good suggestion!

adcc/test_AdcMatrix.py
maxscheurer: Moved them all together except for one, which required `extra` to be present.

adcc/workflow.py
maxscheurer: Good point!

adcc/workflow.py
maxscheurer: What exactly do you mean?

adcc/AdcMatrix.py
maxscheurer: Done.

adcc/AdcMatrix.py
maxscheurer: Done.

adcc/ExcitedStates.py
maxscheurer: Done.

docs/calculations.rst
maxscheurer: Done.

adcc/workflow.py
maxscheurer: Done.

adcc/workflow.py
maxscheurer: Done.

adcc/workflow.py
mfherbst: ok

adcc/workflow.py
mfherbst: To avoid exactly what you did here ... this statement causes a bug if  hf.environment is false, because than line 533 makes it a dict. I think if you remove this branch and allows dicts it would allow you to simplify the parsing logic.

adcc/backends/veloxchem.py
```diff
@@ -130,12 +131,20 @@ def pe_energy(self, dm, elec_only=True):
...
+                                                  elec_only=True)
+            )
+            ptss = EnergyCorrection(
+                "pe_ptss_correction",
+                lambda view: self.pe_energy(view.state_diffdm_ao,
+                                            elec_only=True)
+            )
+            ret.extend([ptlr, ptss])
+        ret = {ec.name: ec for ec in ret}
         return ret
```
mfherbst, line 148:
```suggestion
        return {ec.name: ec for ec in ret}
```

adcc/backends/psi4.py
```diff
@@ -97,12 +108,27 @@ def pe_energy(self, dm, elec_only=True):
...
+                                                  elec_only=True)
+            )
+            ptss = EnergyCorrection(
+                "pe_ptss_correction",
+                lambda view: self.pe_energy(view.state_diffdm_ao,
+                                            elec_only=True)
+            )
+            ret.extend([ptlr, ptss])
+        ret = {ec.name: ec for ec in ret}
+        return ret
```
mfherbst, line 125:
```suggestion
        return {ec.name: ec for ec in ret}
```

adcc/ExcitedStates.py
mfherbst: Ok. I would write "Excitation energy includes these corrections:". I think that's more clear.

adcc/ExcitedStates.py
mfherbst: Ah sorry I misread the code.

adcc/workflow.py
maxscheurer: Dicts are of course allowed, but at this point, all other types are converted to a dict. The branch is just to catch types that we don't convert to a dict.
Sorry, I'm confused 😞

mfherbst: I think it's again me (sorry I have a really busy week): I overlooked the `not`. All good ... in that case don't you think that the parsing logic can be made a little simpler? Maybe you can try to reduce on the number of branches / the additional on-line if ... then ... else thingy

adcc/backends/veloxchem.py
maxscheurer: Done.

adcc/backends/psi4.py
maxscheurer: Done.

adcc/workflow.py
maxscheurer: I have restructured this part a bit and made it a bit more pythonic 😄

adcc/ExcitedStates.py
maxscheurer: Done, also updated in docs example.

mfherbst (approved):
Awesome!

mfherbst:
Unless you want to rebase I would squash-merge once tests pass.

maxscheurer:
Squash-merge is fine 😄 🚀

maxscheurer:
I guess a release of `v0.15.9` would be adequate?

PR #124, opened 2021-01-13
Nuclear Gradients

maxscheurer:
Implemented together with @Drrehn.

## Features
- [x] MP2
- [x] ADC(0)
- [x] ADC(1)
- [x] ADC(2)
- [x] ADC(2)-x
- [x] ADC(3)
- [x] CVS-ADC(1) (by @iubr)
- [x] CVS-ADC(2) (by @iubr)
- [x] CVS-ADC(2)-x (by @iubr)

## ToDo
- [x] merge on `master` (#189)
- [x] rebase on #158 
- [x] finite difference data test generation (preliminary version working)
- [x] tests:
    - [x] individual components (densities etc.)
    - [x] reference tests (preliminary version working)
- [x] rebase once #123 is in (I've added some stuff from this branch here for testing purposes, will be removed)
- [x] ~add equation numbers from papers (Rehn 2019, Levchenko 2005)~


adcc/AmplitudeVector.py
```diff
@@ -94,7 +94,7 @@ def __setitem__(self, index, item):
             else:
                 raise KeyError(index)
         else:
-            super().__setitem__(index, item)
+            return super().__setitem__(index, item)
```
mfherbst, line 97: Julia error :smile:

mfherbst:
Awesome :+1:

mfherbst:
Of course tests need to work and all, but I wonder if a good first aim is to just get the things already implemented merged into master and release a new version?

maxscheurer:
The tests are horrible to implement... ensuring highly converged eigenvectors is difficult 😞 but in principle all the implemented gradient methods are correct 👍🏻

iubr:
Not sure if this is a good suggestion, but maybe we could run a couple of small molecules and save the results in a text file for pytest to compare agains (this is more or less how it's done in veloxchem). I can try out the cvs-adc gradients and find an optimal/good enough setup (I guess the question is of comparing analytical vs. numerical gradients)

maxscheurer:
The generation of finite-difference reference data works fine already, for both generic and CVS-ADC 😊👍🏻

mfherbst:
> The tests are horrible to implement... ensuring highly converged eigenvectors is difficult  

Can't you lower to tolerance or introduce perturbations in the molecule / orbitals to ensure they are sufficiently distinct (to avoid rotations in invariant subspaces etc.)

maxscheurer:
You mean the convergence tolerance/`atol` for asserts?

You mean geometry perturbations to break symmetries? Yes, I can try that for sure 👍🏻

mfherbst:
> You mean the convergence tolerance/atol for asserts?

Yes. It's not amazing, but it will at least catch forgotten factors of two or switched signs.

maxscheurer:
@iubr, the CVS-ADC2-x don't seem to be correct right now, at least I get an error for H2O of ~1e-5 (compared to 5-point finite differences). Maybe I made some mistake while cleaning up the code... Which accuracy could you obtain in your tests? I'll try to run your original branch again at some point to get to the bottom of this.

iubr:
The ~1e-5 is similar to what I get with my code (in comparison to the 2-point finite difference numerical gradient). I checked it today with a 5-point finite difference and it is a little smaller, but still same order (1e-5). So if there is a bug, it would be in my module as well. 
I had discussed this difference with Patrick and we had concluded that it could be due to the tolerance/convergence threshold for the adc step (I couldn't go better than 1e-6 because of the requirement for the HF reference). Of course, this might not be the right explanation for the difference. It's, of course, possible that there is still an undiscovered bug or mistake somewhere in the code/equations. I'll make another check next week. Let's see.

maxscheurer:
> I had discussed this difference with Patrick and we had concluded that it could be due to the tolerance/convergence threshold for the adc step

I converge the SCF to 1e-12, and the ADC procedure is converged to 1e-10 in the energy. Given the fact that all other gradient methods so far can achieve at least 1e-7 accuracy, I'm convinced that one can achieve the same for CVS-ADC2-x . I'll try to add canonical ADC2-x gradients now and see how accurate one can get.

iubr:
Thanks a lot!

mfherbst:
Super cool! I'll review this weekend.

How do we compare speed-wise to alternative ADC(3) gradient implementations?

iubr:
> 

Hej Max,

Sorry for the long silence. There have been a couple of things that I had to prioritize, so I couldn't focus on cvs-adc2x, but I have been debugging these last two weeks. I didn't find anything yet, but still have some ideas of what to try (I confirmed the excitation energy PDMs are correct and tested the lambda multipliers against numerical dipole moments -- the error is the same as in cvs-adc2). I am now re-checking all the omega multipliers, but I am not sure how much longer it will take me to get to the bottom of it.  I hope not too long. Let me see if I can do it over the weekend, otherwise please go ahead without cvs-adc2x.  What was the error in the  correlation energy TPDMs?

And thanks a lot for all the help! Super-exciting that you managed to add adc2x and adc3!!

/Iulia

maxscheurer:
I fixed the TPDMs for CVS-ADC(2)-x in [this](https://github.com/adc-connect/adcc/pull/124/commits/e6cba8090baa2abc0ea6105cb052a1f34612ed94#diff-d8db6a984e6ebd25e75ba9f0b0e9529403c05989ee9d45915448fc744fb88804R281), where the prefactor of `g2a.vvvv` was wrong.

The "correlation part" of the energy is tested [here](https://github.com/adc-connect/adcc/blob/c1481ee144584d270d39baf6a732212f9e9f5231/adcc/gradients/test_functionality_gradients.py#L78).

iubr:
Thanks! I'll look into it.

iubr:
> 

I double checked the equations for the TPDMs and and I do get a factor 2 in the VVVV block for cvs-adc2x. In my module, it's the only way I get the correct excitation energy. I also wrote a stand-alone code to compute the excitation energy from the density matrices and without this factor 2 I do not get the correct energy. I attach the code here as a txt file in case you'd like to have a look [cvs_adc2x_dms_from_scratch.txt](https://github.com/adc-connect/adcc/files/8055240/cvs_adc2x_dms_from_scratch.txt)
. I still have to have a more careful look at the adcc version to understand what's going on,  but next week is quite crazy for me up until Thursday. It could also be great to have a short meeting -- maybe you spot something I don't in the equations. What do you think?  

Regarding cvs-adc2x updates, over the weekend I still haven't managed to figure out the reason why analytical vs. numerical agree only up to 1e-5 and not 1e-7 (as for the other adc orders). Maybe it's best you go ahead without cvs-adc2x for the moment and we add it once this mystery is solved (which I hope will not be that much longer).

maxscheurer:
Thanks for the script, I'll have a look.

iubr:
Thanks! If you'd like an overview of the density matrices, they are from https://doi.org/10.1063/5.0058221, Appendix C, Table II, 3rd column (ADC matrix terms).

iubr:
Hi Max,

I just found the bug! It was a sign in one of the terms for the CV block of the lambda multipliers. See [here:](https://github.com/iubr/adcc/commit/a58ccd3886a1fded5fce96aec37e34b737ba301b). 
I tried to merge the latest version of adcc/gradients branch into my cvs_gradients branch, but for some reason I cannot get it running on my computer, so I fixed the bug using the old version of the code. If you could make the fix at your end and test cvs-adc2x again, it would be super great!! Thanks a lot!! I'll make a few more tests on my end just to make sure there's nothing else.

/Iulia

maxscheurer:
That's great news! I will try it out later. 👍🏻😊

iubr:
Thank you!! Let me know how it goes :D
(I still haven't figured out the puzzle of the factor 2 in the VVVV block between my module and adcc, but I'm planning to look into it again tomorrow)

maxscheurer:
I found the reason for the weird factor 2 behavior in [this](https://github.com/adc-connect/adcc/blob/c1481ee144584d270d39baf6a732212f9e9f5231/adcc/gradients/TwoParticleDensityMatrix.py#L60)
 block of code. The blocks `oooo` and `vvvv` were added _again_ in the case of CVS, such that they existed twice in the `TwoParticleDensityMatrix` object. 😁 

I found some other inconsistencies in the formation of `g2a_hf` and `g2a_oresp`, with these fixes altogether, everything works!!! 💪

mfherbst:
Cool! Great work.

iubr:
Wonderful ^-^!! Thanks a lot @maxscheurer!!

adcc/gradients/TwoParticleDensityMatrix.py
```diff
@@ -0,0 +1,290 @@
...
+            )
+        return (g2_ao_1.evaluate(), g2_ao_2.evaluate())
+
+    def to_ao_basis(self, refstate_or_coefficients=None):
+        """
+        Transform the density to the AO basis for contraction
+        with two-electron integrals.
+        ALL coefficients are already accounted for in the density matrix.
+        Two blocks are returned, the first one needs to be contracted with
+        prqs, the second one with -psqr (in Chemists' notation).
```
mfherbst, line 219: I don't get that comment.

adcc/gradients/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+from .orbital_response import (
+    orbital_response, orbital_response_rhs, energy_weighted_density_matrix
+)
+from .amplitude_response import amplitude_relaxed_densities
+
+
+@dataclass(frozen=True)
+class GradientComponents:
+    natoms: int
+    nuc: np.ndarray
```
mfherbst, line 43: Nuc is a bit short, maybe add a comment here to explain

adcc/gradients/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+    overlap: np.ndarray
+    hcore: np.ndarray
+    two_electron: np.ndarray
```
mfherbst, line 46: comment: Two electron integrals?

adcc/gradients/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+    custom: Optional[Dict[str, np.ndarray]] = None
+
+    @property
+    def total(self):
+        """Returns the total gradient"""
+        ret = sum([self.nuc, self.overlap, self.hcore, self.two_electron])
+        if self.custom is None:
+            return ret
+        for c in self.custom:
+            ret += self.custom[c]
```
mfherbst, line 56:
```suggestion
        if self.custom:
            return ret + sum(self.custom.values())
        else:
            return ret
```

adcc/gradients/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+    timer: Timer
+    g1a: Optional[OneParticleOperator] = None
+    g2a: Optional[TwoParticleDensityMatrix] = None
+
+    @property
+    def reference_state(self):
+        return self.excitation_or_mp.reference_state
+
+    @property
+    def _energy(self):
```
mfherbst, line 79: Why the underscore? If someone uses it it's sort of their fault, no? If you add a warning I think that's fine.

adcc/gradients/orbital_response.py
```diff
@@ -0,0 +1,340 @@
...
+    Solves the orbital response equations
+    for a given reference state and right-hand side
+    """
+    # TODO: pass solver arguments
+    A = OrbitalResponseMatrix(hf)
+    Pinv = OrbitalResponsePinv(hf)
+    x0 = (Pinv @ rhs).evaluate()
+    lam = conjugate_gradient(A, rhs=rhs, x0=x0, Pinv=Pinv,
+                             explicit_symmetrisation=None,
+                             callback=default_print)
```
mfherbst, line 339: Which tolerance is used and why? Should for sure be related to the convergence used for the ADC excitation vectors ... or maybe aim for relative residual decrease?

mfherbst (commented):
Just a few comments ... obviously I have not at all checked the equations.

adcc/backends/psi4.py
```diff
@@ -32,6 +33,72 @@
...
+        self.mints = psi4.core.MintsHelper(self.wfn)
+
+    def correlated_gradient(self, g1_ao, w_ao, g2_ao_1, g2_ao_2):
+        """
+        g1_ao: relaxed one-particle density matrix
+        w_ao:  energy-weighted density matrix
+        g2_ao_1, g2_ao_2: relaxed two-particle density matrices
+        """
+        natoms = self.mol.natom()
+        Gradient = {}
```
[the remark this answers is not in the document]
maxscheurer: Shamelessly copied from some `psi4numpy` tutorial, will change this 😄

adcc/backends/psi4.py
```diff
@@ -32,6 +33,72 @@
...
+        # Build Total OEI Gradient
+        Gradient["OEI"] = Gradient["T"] + Gradient["V"] + Gradient["S"]
+
+        # Build TE contributions
+        for atom in range(natoms):
+            deriv2 = self.mints.ao_tei_deriv1(atom)
+            for p in range(3):
+                deriv2_np = np.asarray(deriv2[p])
+                Gradient["TEI"][atom, p] += np.einsum(
+                    'pqrs,prqs->', g2_ao_1, deriv2_np, optimize=True
```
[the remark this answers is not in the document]
maxscheurer: True, also copy/pasted from the tutorial 😄

adcc/gradients/__init__.py
maxscheurer: Good idea, I think it would be good to make all attribute names a bit more verbose.

adcc/gradients/__init__.py
maxscheurer: Good point. Will rename it to `correlation_energy` or so to make it more clear what exactly is returned.

adcc/gradients/orbital_response.py
maxscheurer: I think it should rather be related to the SCF convergence criterion, because we compute the response of the HF ground state (kind of 😄). In addition, MP states don't have an additional convergence criterion, so I will make this depend on the SCF conv tol. 👍

maxscheurer:
## Performance Optimization: 2-RDM / ERI Derivative Contraction (PySCF Backend)

### Problem

The current gradient implementation has two major memory bottlenecks in `correlated_gradient`:

1. **Full AO transform of the 2-RDM**: `g2_total.to_ao_basis()` produces two full `(nao, nao, nao, nao)` arrays
2. **Full ERI derivative pre-computation**: `mol.intor('int2e_ip1', aosym='s1')` → shape `(3, nao, nao, nao, nao)`

Peak memory: ~5 × N⁴_AO × 8 bytes (at nao=100 → ~4 GB, at nao=150 → ~20 GB).

### How PySCF handles this (MP2/CCSD gradients)

PySCF uses a **two-part strategy** that avoids full N⁴ tensors entirely:

**Part 1: Correlated 2-RDM → shell-batched direct contraction**
- **Half-transforms** the 2-RDM: only virtual→AO, keeping occupied indices in MO basis. For MP2: `(nocc, nocc, nao, nao)` instead of `(nao, nao, nao, nao)`.
- **Shell-batched streaming** of `int2e_ip1` (memory-capped, with `aosym='s2kl'` for kl permutational symmetry).
- Completes the remaining occ→AO transform **on-the-fly** within the shell loop.
- Direct `einsum` contraction of the ERI shell-slice with the density batch → accumulate gradient.

**Part 2: Separable 1-RDM contributions → J/K gradient machinery**
- HF and orbital-response parts of the 2-RDM that factorize as products of 1-RDMs can be handled via `_vhf.direct_mapdm` (C-level integral-driven J/K with screening).
- PySCF builds `vhf` (J-K response) piggybacking on the same shell loop for "free".
- The correlated 2-RDM blocks (t2, u·u type) **cannot** use J/K — they require direct 4-index contraction.

### Proposed Optimization Strategy

New method: `PyScfGradientProvider.correlated_gradient_direct(g2_mo, g1, w, refstate)` that accepts the block-sparse `TwoParticleDensityMatrix` in MO basis.

**Combined approach — block-wise half-transform + shell-batched ERI contraction:**

| Block type | Strategy |
|---|---|
| `oovv` (MP2 t2-like) | Half-transform vv→AO → `(nocc², nao²)`, then complete occ→AO on-the-fly in shell loop |
| `ovov`, `ooov` | Half-transform the virtual indices → `(nocc·nvir, nao)` or similar, stream similarly |
| `oooo` | Tiny (`nocc⁴`), can be fully transformed cheaply |
| `ovvv`, `vvvv` | Large, but only appear at ADC(2)/ADC(3); will need batched strategies |
| HF `oooo` + orbital-response blocks | Factor as 1-RDM products → use `get_jk` gradient calls |

**Memory target**: O(nocc² × nao²) + O(shell_batch × nao³) instead of O(nao⁴).

### Implementation Plan

1. Start with **MP2** (only `oovv` block non-zero in g2a) as proof of concept
2. Keep existing naive `correlated_gradient` for validation
3. Generalize to ADC(1) (adds `ovov` + orbital-response separable blocks)
4. Then extend to ADC(2)/ADC(3) with all blocks

### Reference

- PySCF MP2 gradient: `pyscf/grad/mp2.py` — `grad_elec()`, `_shell_prange()`
- PySCF CCSD gradient: `pyscf/grad/ccsd.py` — `_rdm2_mo2ao()` (HDF5-backed half-transform)
- adcc gradient: `adcc/gradients/__init__.py`, `adcc/backends/pyscf.py`
- Theory: Rehn, Dreuw et al., J. Chem. Phys. 150, 164102 (2019), DOI: 10.1063/1.5085117

maxscheurer:
Closed in favor of #230

PR #125, opened 2021-01-17
Fix AmplitudeVector and coveralls

maxscheurer:
This PR fixes some weird behavior of the `AmplitudeVector` and coveralls upload.
`__setattr__` is now forwarded to `__setitem__`, in analogy to `__getattr__`

Test that would hit the previous error added. I'd vote for a `0.15.7` release after this fix, because this is a serious bug...


mfherbst (approved):
Looks good and I agree. Version bump after this.

PR #130, opened 2021-06-10
Add DIP scheme to adc-connect

Drvanon:
TODO:

- [x] Matrix multiplications
  - [x] Implement diagonal multiplications with AmplitudeVectors (adcc/adcc/adc_dip/matrix.py) 
  - [x] Implement offdiagonal multiplications with AmplitudeVectors (adcc/adcc/adc_dip/matrix.py)  
  - [ ] Edit transitions for DIP (adcc/adcc/adc_dip/{modified_transition_moments.py, state2state_transition.py, transition_dm.py})
- [x]  Guesses:
  - [x] Implement guess_zero for DIP singles and doubles (adcc/adcc/guess/guess_zero.py)
  - [x] Implement guesses_from_diagonal for DIP singles (adcc/adcc/guess/guess_zero.py)
  - [ ]  Implement guesses_from_diagonal for DIP doubles (libadc/libadcc/fill_dip_doubles_guesses.cc)
- [ ] Test:
  - [ ] Matrix multiplications
    - [ ] Make code more general so that the tests execute for both DIP and PP
    - [ ] Add test data 
  - [ ] Guesses
- [ ] Adapt AdcMatrix to support multiple schemes
  - [x] Make AdcMatrix subclassable in a way that it allows for  multipleschemes
  - [x] Make the PP AdcMatrix a normal subclass of AdcMatrix
  - [ ]  Make the PP AdcMatrix the default matrix in order to ensure backwards compatibility
  - [ ] Adapt the AdcMethod class so that scheme can be set (default should remain PP)
  - [ ] Change workflow.py so that scheme can be chosen through a parameter or by setting it in the AdcMethod
- [ ]  Make ExcitedStates file more general so that it will work for all schemes
  - [ ]  Adapt linewidth and format methods on FormatExcitationVector
  - [ ]  Edit ExcitedStates.describe so that it is compatible with DIP. Sugestion: generate a pandas.DataFrame and print it with extra headers.


adcc/guess/guess_zero.py
```diff
@@ -84,48 +84,65 @@ def guess_symmetries(matrix, spin_change=0, spin_block_symmetrisation="none"):
                          "You passed {}".format(spin_change))
 
     max_spin_change = 0
-    if "ph" in matrix.axis_blocks:
+    if any(s in matrix.axis_blocks for s in ["ph", "hh"]):
         max_spin_change = 1
-    if "pphh" in matrix.axis_blocks:
+    if any(s in matrix.axis_blocks for s in ["pphh", "phhh"]):
```
Drvanon, line 89: Use length as a way of calculating max spin. 
Assert length is even.

PR #132, opened 2021-07-12
LR-PCM ADC for PSI4

jonasleitner:
~~For some reason the PSI4 PE and PCM tests don't work when executed both. When executed separately they pass.~~
Solved


adcc/backends/psi4.py
```diff
@@ -64,6 +64,15 @@ def pe_induction_elec_ao(dm):
                 )[1]
             return pe_induction_elec_ao
 
+    @property
+    def pcm_potential_elec(self):
+        if self.wfn.PCM_enabled():
+            def pcm_potential_elec_ao(dm):
+                return psi4.core.PCM.compute_V(
+                    self.wfn.get_PCM(),
+                    psi4.core.Matrix.from_array(dm.to_ndarray()))
```
maxscheurer, line 73: Can you move the last `)` to the next line?
```suggestion
                return psi4.core.PCM.compute_V(
                    self.wfn.get_PCM(),
                    psi4.core.Matrix.from_array(dm.to_ndarray())
                )
```

adcc/backends/psi4.py
```diff
@@ -128,6 +137,8 @@ def environment(self):
         ret = None
         if hasattr(self.wfn, "pe_state"):
             ret = "pe"
+        if self.wfn.PCM_enabled():
```
maxscheurer, line 140: `elif`?

adcc/backends/pyscf.py
```diff
@@ -262,7 +262,7 @@ def import_scf(scfres):
 
 
 def run_hf(xyz, basis, charge=0, multiplicity=1, conv_tol=1e-11,
-           conv_tol_grad=1e-9, max_iter=150, pe_options=None):
+           conv_tol_grad=1e-9, max_iter=150, pe_options=None, pcm=None):
```
maxscheurer, line 265: I think we need some more general option passing for the environment here...
Maybe an `environment` dictionary with keywords `model` and `options` would be better.
The function is supposed to be `adcc`-internal, so we can decide what we want. If you have 2mins, can you give this a quick look, @mfherbst?

adcc/backends/testing.py
```diff
@@ -189,21 +189,33 @@ def cached_backend_hf(backend, molecule, basis, multiplicity=1, conv_tol=1e-12,
 
     global __cache_cached_backend_hf
 
+    # so you can only assign one of them.
+    pe_options = None
+    pcm = None
+    if env[0] == "pcm":
+        pcm = env[1]
+    elif env[0] == "pe":
+        pe_options = env[1]
```
maxscheurer, line 198: This could also be much simpler if we can generalize the environment keyword passing somehow....

adcc/backends/testing.py
```diff
@@ -189,21 +189,33 @@ def cached_backend_hf(backend, molecule, basis, multiplicity=1, conv_tol=1e-12,
...
     # with pyscf
     if backend == "pyscf":
         return payload()
 
-    key = (backend, molecule, basis, str(multiplicity))
+    # otherwise PE and PCM share the same key
+    if env[0]:
+        key = (backend, molecule, basis, str(multiplicity), env[0])
+    else:
+        key = (backend, molecule, basis, str(multiplicity))
```
maxscheurer, line 218: Same here....

adcc/testdata/cache.py
```diff
@@ -258,3 +258,4 @@ def read_yaml_data(fname):
 
 qchem_data = read_yaml_data("qchem_dump.yml")
 tmole_data = read_yaml_data("tmole_dump.yml")
+psi_data = read_yaml_data("psi_dump.yml")
```
maxscheurer, line 261: `psi4` for consistency within `adcc`...
```suggestion
psi4_data = read_yaml_data("psi4_dump.yml")
```

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,71 @@
...
+            adcc.run_adc(scfres, method=method, n_singlets=5)
+
+        # automatically add coupling term
+        state = adcc.run_adc(scfres, method=method, n_singlets=5,
+                             conv_tol=1e-7, environment="linear_response")
+        assert_allclose(
+            state.excitation_energy_uncorrected,
+            psi_result["excitation_energy"],
+            atol=1e-5
+        )
```
maxscheurer, line 66: You can also dump the oscillator strengths from Psi4 CIS, those should roughly be equivalent to ADC(1)/ISR(0) oscillator strengths. You can get the ISR(0) properties by:
```Python
state_cis = adcc.ExcitedStates(state, property_method="adc0")
```

I think we had some problems with ADC(0) dipole moments at some point, but one should give it a try at least.

adcc/testdata/psi_dump.yml
```diff
@@ -0,0 +1,12 @@
+formaldehyde_sto3g_pcm_adc1:
```
maxscheurer, line 1: Can you maybe add the Python file with which you created the reference data? 😄

maxscheurer (commented):
Some comments from a first scroll-through... 😄 Looks good!

adcc/backends/pyscf.py
mfherbst: Yes I agree. It's running out of hands with the arguments. Perhaps an `environment` dict with a key `pe` and `pcm` to the respective options ?

adcc/backends/testing.py
mfherbst: Agree. This looks horrible.

mfherbst:
But thanks for the contribution @jonasleitner! This actually solves part of the angle of a project I'm involved with, so thanks for doing some of my work ;).

adcc/backends/pyscf.py
maxscheurer: I thought about it again... Maybe generalizing this does not make sense, because the handling of PCM options is different in each host program. The reason why we can do this for PE is that CPPE is used throughout. So, for now I suggest to completely remove PCM stuff from the `run_hf` and `cached_backend_hf` functions, and implement some routine that runs the PCM-HF in the test file.
Anything else would be overengineering for things we don't even support right now.

jonasleitner: Do you think we need a lot of different PCM calculations for the tests?
If yes, I agree with you @maxscheurer. Otherwise I think it would be better to just add the few extra lines in the corresponding `run_hf` and then maybe use the dict with a bool or sth like that to trigger PCM

adcc/backends/test_backends_pcm.py
jonasleitner: seems to work fine, will add it.

adcc/testdata/psi_dump.yml
jonasleitner: You mean add it to the testdata folder?
Sure. But why?

adcc/backends/pyscf.py
jonasleitner: Misunderstood the comments... yes you are right @maxscheurer, a dict of this form would not make sense for PCM.
But at least at the moment with only PCM and PE it would be possible to use an if for PCM and else for PE. Depends on the other envs that you plan to add if this makes sense. And of course the PCM options would be rather fixed in `run_hf`, though maybe 1 or 2 options could be passed with the dict... not sure right now.

maxscheurer: I see your point, but I don't like that we would need to hard-code PCM parameters in the respective backends, even if it's just for testing purposes. The reason is that it could be 'misused' as a wrapper for running HF with some backend, and would produce unpredictable results in that case. Another point is that you'd need to raise `NotImplementedError`s for the unsupported backends etc., which would result in quite some code addition for just 1-2 test cases.
So, I'd say make a simple PCM-HF routine in your test file which serves the purpose - in my opinion one can generalize it when the time comes 😉

adcc/testdata/psi_dump.yml
maxscheurer: For reproducibility purposes, and in case somebody except yourself wants to generate more reference data with the same setup. I think there is also a script file that generates Q-Chem data or so. You could create a `generate_psi4_data.py` script that produces your .yml file 😃

adcc/backends/pyscf.py
jonasleitner: ok good points.

jonasleitner:
Moved the PCM HF calc to the test file... Using a dict atm, but not sure how flexible the PCM options need to be. (For now they could also be fixed)
The Yaml generator works, but I don't like the layout of the lists. Would prefer the lists to be printed as lists. (As in the psi4_dump.yml)

adcc/testdata/generate_psi4_data.py
```diff
@@ -0,0 +1,97 @@
...
+
+    ret = {"basis": basis,
+           "method": "adc1",
+           "molecule": molecule}
+
+    wfn, res = run_psi4_tdscf(geom, basis, pcm_options=pcm_options)
+
+    ret["energy_scf"] = wfn.energy()
+    # yaml safe_dump doesn't like np.floats
+    ret["excitation_energy"] = [round(float(r["EXCITATION ENERGY"]), 5)
```
maxscheurer, line 74: Convert to list:
```Python
list([round(float(r["EXCITATION ENERGY"]), 5) for r in res])
```

jonasleitner: it is already a list

maxscheurer: Ah yes, I was confused because of the yaml output stuff... I'm wondering why the print-out looks weird.

jonasleitner:
fixed the yaml generator. The documentation of the package is just awesome.

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,142 @@
...
+backends = [b for b in adcc.backends.available()
+            if b not in ["molsturm", "veloxchem", "pyscf"]]
+basissets = ["sto-3g", "cc-pvdz"]
+methods = ["adc1", "adc2", "adc3"]
+
+
+@pytest.mark.skipif(len(backends) == 0, reason="No backend found.")
+@expand_test_templates(list(itertools.product(basissets, methods, backends)))
+class TestPCM(unittest.TestCase):
+    def template_pcm_linear_response_formaldehyde(self, basis, method, backend):
```
maxscheurer, line 28: I think I'd create a Psi4-specific function here, and only expand templates on `basissets` and `methods` for the time being. We won't have the other backends supported for quite some time in the future I guess.

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,142 @@
+import pytest
+import unittest
+import itertools
+import adcc
+import adcc.backends
+import psi4
```
maxscheurer, line 6: Import in Psi-specific routine, otherwise loading this file fails (see CI) if Psi not available.

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,142 @@
...
+        psi4_result = psi4_data[basename]
+
+        # so you can adapt some options more easily if needed
+        psi4_pcm_options = {"weight": 0.3, "pcm_method": "IEFPCM", "neq": True,
+                            "solvent": "Water"}
+        pcm_options = {"psi4": psi4_pcm_options}
+
+        hf_name = f"{backend}_run_pcm_hf"
+        hf_function = globals()[hf_name]
+        scfres = hf_function(static_data.xyz["formaldehyde"], basis, charge=0,
```
maxscheurer, line 41: ... and then directly call psi-specific function here.

maxscheurer (commented):
Except for some cleanup, this looks good!

adcc/backends/test_backends_pcm.py
jonasleitner: unfortunately true

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,139 @@
...
+from adcc.testdata.cache import psi4_data
+from numpy.testing import assert_allclose
+from adcc.exceptions import InputError
+
+from adcc.adc_pp.environment import block_ph_ph_0_pcm
+from adcc.AdcMatrix import AdcExtraTerm
+
+# remove pyscf until implemented
+backends = [b for b in adcc.backends.available()
+            if b not in ["molsturm", "veloxchem", "pyscf"]]
```
maxscheurer, line 19: Remove.

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,139 @@
...
+basissets = ["sto-3g", "cc-pvdz"]
+methods = ["adc1", "adc2", "adc3"]
+
+
+@pytest.mark.skipif(len(backends) == 0, reason="No backend found.")
```
maxscheurer, line 24:
```suggestion
@pytest.mark.skipif("psi4" not in adcc.backends.available(), reason="Psi4 backend not found.")
```

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,139 @@
...
+@expand_test_templates(list(itertools.product(basissets, methods)))
+class TestPCM_psi4(unittest.TestCase):
```
maxscheurer, line 26:
```suggestion
class TestPCM(unittest.TestCase):
```

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,139 @@
...
+    def template_pcm_linear_response_formaldehyde(self, basis, method):
```
maxscheurer, line 27:
```suggestion
    def template_pcm_psi4_linear_response_formaldehyde(self, basis, method):
```

maxscheurer (commented):
Cool, just some more small comments 😄

maxscheurer:
Almost ready I think 👍 @mfherbst, can you give this another look if you have the time 😄 
The py39 pipeline fails due to the newest `h5py` not being compatible with `pyscf<2.0.0`, but `pyscf==2.0.0` is not yet on pypi... Coverage decreases because we don't test Psi4 in our CI pipeline. Maybe we should add a minimal conda build & test (only `h2o_sto3g`) with Psi4 in a follow-up PR?

jonasleitner:
I added the ptLR correction for PSI4 to this PR as we discussed @maxscheurer .
Since I'm not aware of another ptLR implementation, I validated the implementation by comparing to excitation energies from LR-PCM. For the Nitrobenzene Benchmark set I used to validate LR-PCM, the excitation energies agree up to 1e-3 Hartree. Additionally degenerate states also have degenerate corrections ;). So I would assume the implementation should be fine.

Concerning the tests... I'm not sure about the consistency test (compare to values calculated with the current versions of PSI4 and ADCc). Probably I should raise the tolerance, to avoid failures through some smaller, "irrelevant" changes in PSI4/PCMSolver. Also I'm not sure whether it is a good idea to add the generation of the Testdata to the _generate.py_ script, since after running the script the consistency test will be passed 100%.
Other than that I don't know at the moment what other tests to add, since PTE energies and reference values for corrections are not available.

maxscheurer:
Can you do a rebase against the master branch? 😬

adcc/OperatorIntegrals.py
```diff
@@ -187,6 +188,19 @@ def pe_induction_elec(self):
...
         return self.__import_density_dependent_operator(callback)
 
+    @property
+    def pcm_potential_elec(self):
+        """
+        Returns a function to obtain the (density-dependent)
+        electronic PCM potential operator in the molecular orbital basis
+        """
+        if "pcm_potential_elec" not in self.available:
+            raise NotImplementedError("electronic PCM potential operator "
```
maxscheurer, line 198:
```suggestion
            raise NotImplementedError("Electronic PCM potential operator "
```

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,206 @@
...
+from adcc.adc_pp.environment import block_ph_ph_0_pcm
+from adcc.AdcMatrix import AdcExtraTerm
+
+backends = [b for b in adcc.backends.available()
+            if b in ["psi4", "pyscf"]]
+basissets = ["sto-3g", "cc-pvdz"]
+methods = ["adc1", "adc2", "adc3"]
+
+
+@pytest.mark.skipif(len(backends) == 0, reason="Psi4 and Pyscf backends not found.")
```
maxscheurer, line 23:
```suggestion
@pytest.mark.skipif(len(backends) == 0, reason="No backend for PCM available.")
```

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,206 @@
...
+                        max_iter=150, pcm_options=pcm_options)
+
+        assert_allclose(scfres.energy_scf, result["energy_scf"], atol=1e-8)
+
+        state = adcc.run_adc(scfres, method=method, n_singlets=5,
+                             conv_tol=1e-7, environment="ptlr")
+
+        # compare ptLR result to LR data
+        assert_allclose(state.excitation_energy,
+                        result["lr_excitation_energy"], atol=5 * 1e-3)
```
maxscheurer, line 53:
```suggestion
                        result["lr_excitation_energy"], atol=5e-3)
```

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,206 @@
...
+
+        if backend == "psi4":
+            data = psi4_data
+            run_hf = psi4_run_pcm_hf
+            pcm_options = {"weight": 0.3, "pcm_method": "IEFPCM", "neq": True,
+                           "solvent": "Water"}
+        elif backend == "pyscf":
+            data = pyscf_data
+            run_hf = pyscf_run_pcm_hf
+            pcm_options = {"eps": 78.3553, "eps_opt": 1.78}
```
maxscheurer, line 77: This is identical to what you have in the other test above. Can probably be put in a `dict` inside the test class:

```Python
config = {
    "psi4": {"data": psi4_data, "run_hf": psi4_run_pcm_hf, "pcm_options": ...}
    "pyscf": ...
}
```
or sth of the sort...

adcc/backends/test_backends_pcm.py
```diff
@@ -0,0 +1,206 @@
...
+            state.excitation_energy_uncorrected,
+            result["lr_excitation_energy"],
+            atol=1e-5
+        )
+
+        if backend == "psi4":
+            # remove cavity files from PSI4 PCM calculations
+            for cavityfile in os.listdir(os.getcwd()):
+                if cavityfile.startswith(("cavity.off_", "PEDRA.OUT_")):
+                    os.remove(cavityfile)
```
maxscheurer, line 129: Also duplicate, maybe write a function `remove_cavity_psi4`?

maxscheurer (commented):
Some minor comments.

maxscheurer:
BTW @jonasleitner feel free to add yourself [here](https://github.com/adc-connect/adcc/blob/fe3d42c7cfece3cb4d0e477c62166f8ff73cbc6c/adcc/__init__.py#L72) and [here](https://github.com/adc-connect/adcc/blob/fe3d42c7cfece3cb4d0e477c62166f8ff73cbc6c/setup.py#L503).

adcc/backends/test_backends_pcm.py
jonasleitner: and I could also fix the method to adc1, because atm I only test for adc1 (although I could also adc2 etc. tests for ptLR)

maxscheurer: I'd say fix it to ADC1 outside, because it tests all the necessary machinery + we don't have any reference data for LR 👍

maxscheurer:
The only things left to do are:

- Adding an example file for these features for each host programs for users to have a good starting point. Maybe create a folder named "environment" or so in examples and also move all the PE examples in there.

- Adding documentation for the new features. There is already something about PE in `calculations.rst`, so it would be easy to show one of the examples in there as well.

adcc/backends/test_backends_pcm.py
jonasleitner: I just add the dict at the end of the file. If i define it as static variable, the program can't find the pcm_hf functions defined below.

mfherbst:
I'll try to do a review pass as well, might not get around to it before the weekend, though.

jonasleitner:
> Adding an example file for these features for each host programs for users to have a good starting point. Maybe create a folder named "environment" or so in examples and also move all the PE examples in there
You mean the two pna examples or is there anything else at another place?

maxscheurer:
> You mean the two pna examples or is there anything else at another place?

I think it's only the content of the pna folder...

jonasleitner:
Ok, beside the test calculations this should be pretty much it (and of course some typos probably :D).

examples/environment/psi4_adc2_pna_ptlr_pcm.py
```diff
@@ -0,0 +1,57 @@
...
+""")
+
+psi4.core.set_num_threads(4)
+
+scf_e, wfn = psi4.energy('scf', return_wfn=True)
+
+# Run an ADC2 calculation with ptLR
+state = adcc.adc2(wfn, n_singlets=5, conv_tol=1e-8,
+                  environment="ptlr")
+print(state.describe())
```
mfherbst, line 57: no line break

docs/calculations.rst
```diff
@@ -811,6 +811,98 @@ The output of the last two lines is::
...
+
+   scf_e, wfn = psi4.energy('scf', return_wfn=True)
+
+   # Run a ADC2 calculation with ptLR
+   state = adcc.adc2(wfn, n_singlets=5, conv_tol=1e-8,
+                     environment="ptlr")
+   print(state.describe())
+
+
+
```
mfherbst, line 904: You should strongly emphasise (even with a warning box) that one needs to adjust the dielectric constant to the optical value for pyscf.

mfherbst (approved):
Very nice work, @jonasleitner, thanks! I only got a few nits.

mfherbst:
I guess coverage goes down because the psi4 tests are not run, right?

maxscheurer:
> I guess coverage goes down because the psi4 tests are not run, right?

Yes, that's correct. The matrix setup etc. is however covered through `pyscf` tests.

docs/calculations.rst
jonasleitner: Yes your right. Then it would be probably the best to also show this in the example.
Fine with having two examples here?
Think Psi4 would be good to have for the rather complicated PCM input there

maxscheurer: No problem with more examples I guess 👍

maxscheurer:
Once you've addressed Michael's comments, can you confirm that the Psi4 tests are working locally for you? Afterwards, I can merge the PR 👍

docs/calculations.rst
mfherbst: Yeah why not, but than output from one is probably sufficient.

jonasleitner:
> Once you've addressed Michael's comments, can you confirm that the Psi4 tests are working locally for you?

Tests are running locally :+1:

docs/calculations.rst
mfherbst: or not ... it's just a small box, so let's leave it as is.

maxscheurer:
> Tests are running locally

Ok, so can I merge your first PR then? 😄

jonasleitner:
> Ok, so can I merge your first PR then?

Yes go on :+1: 
Everything should be fine :tada:

PR #134, opened 2021-12-14
Remove Python 3.6

maxscheurer:
As it says in the title. Psi4 already dropped support for Python 3.6 in the recently released `v1.5`, so let's do the same 😄


mfherbst (approved):
I'm fully with you. How about 3.10 support?

maxscheurer:
I think we should add 3.10 support, probably in a separate PR, I'd like to refactor some of the conda stuff, too...

mfherbst:
ok, so I merge this :+1:.

PR #137, opened 2022-01-06
Psi4 tests on CI

maxscheurer:
Psi4 tests on Conda Python CI. Requires #143.


maxscheurer:
Motivation: Would be nice to have this for more realistic coverage reports (including Psi4) & build on CI against c-f packages.

Issue #140, opened 2022-01-11
matplotlib and pandas as optional dependencies


maxscheurer:
To reduce the dependencies of `adcc`, maybe it would be good to remove matplotlib and pandas from the dependencies. They are only required for specialty features (`plot_spectrum` and `to_dataframe`) but carry a long list of dependencies themselves, even if they're not needed for "core" features.

The correct versions can be "enforced" in conda recipes by [`run_constrained`](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained) and also for pypi.

Suggested by @loriab.

maxscheurer:
Closed in #141.

PR #141, opened 2022-01-12
Make pandas and mpl dependencies for optional features

maxscheurer:
mpl is used in two functions, pandas in one function, both could be considered as optional features (see #140).
This PR removes the two modules as mandatory dependencies, making `adcc` a bit more lightweight. 🚀 

A new decorator for the functions requiring the modules checks upon call whether the modules are found on the system or not and gives instructions on how to obtain them.

Unrelated to that, I have fixed all broken example python scripts in `examples`.


mfherbst:
Could you also add this to the docs (installation instruction and the tutorial guide we have). People might expect the analysis features to "just work" only to be presented with an error when they try it ...

maxscheurer:
Yes, I was going to add it anyways, but was waiting for your feedback on the idea in general 👍

maxscheurer:
Docs added. Note that I won't update the local conda recipes anymore, they will hopefully be removed soon...

docs/installation.rst
```diff
@@ -18,6 +18,9 @@ Please get in touch
 by `opening an issue <https://github.com/adc-connect/adcc/issues>`_
 if you cannot get adcc to work.
 
+Some specialty features of adcc require extra dependencies (Pandas and
```
mfherbst, line 21: I would add an `e.g.`, that ensure the docs don't become outdated so easily (one quickly forgets to add a dependency here, while it's clear to put it to the docs below.

docs/installation.rst
```diff
@@ -154,6 +157,20 @@ the :ref:`devnotes` will provide
 you with some useful pointers to get started.
 
 
+.. _optional-dependencies:
+
+Optional dependencies for analysis features
+-------------------------------------------
+
```
mfherbst, line 164: From my experience users rarely read such a block text in the docs. I think it's wise to first show a list of "optional packages" along with a code block listing the conda or pip command to install them. You can then add the more detailed info in a paragraph afterwards.

mfherbst (approved):
A few more nits about the docs. In DFTK we actually started running the (small and fast) examples in CI to ensure the code is not suddenly broken. Maybe that's something we should think about as well?

docs/installation.rst
maxscheurer: Done.

maxscheurer:
I noticed that sphinx does not process some files, and as such they do not appear in the docs 😠 Wonder why we didn't notice it before...

docs/installation.rst
mfherbst: Sounds good to be.

mfherbst (approved):

mfherbst:
> Wonder why we didn't notice it before...

Probably because no-one checked ...

maxscheurer:
Now the docs are looking good again 😄

PR #143, opened 2022-01-13
Remove our own conda and add Conda Python to CI tests

maxscheurer:


mfherbst (approved):
Very nice to get rid of this!

mfherbst:
Yes, I think so (as you already did, @maxscheurer, right?).

maxscheurer:
Yes 😄

PR #144, opened 2022-01-14
Smoke Tests

maxscheurer:
Minimal testing of functionality.

### DIFF
--- a/adcc/test_smoke.py
+++ b/adcc/test_smoke.py
@@ -73,13 +73,15 @@ class TestSmoke(unittest.TestCase):
     conv_tol = 1e-7
 
     def _run_scf_h2o_sto3g(self):
+        # NOTE: cached_backend_hf not used because adcc.testdata
+        # is not included in installation
         scfres = adcc.backends.run_hf(
             "pyscf", xyz="""
             O 0 0 0
             H 0 0 1.795239827225189
             H 1.693194615993441 0 -0.599043184453037""",
-            basis="sto-3g", conv_tol=1e-12,
-            conv_tol_grad=1e-10
+            basis="sto-3g", conv_tol=self.conv_tol / 100,
+            conv_tol_grad=self.conv_tol / 10
         )
         return scfres
 
--- a/adcc/test_smoke.py
+++ b/adcc/test_smoke.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+## vi: tabstop=4 shiftwidth=4 softtabstop=4 expandtab
+## ---------------------------------------------------------------------
+##
+## Copyright (C) 2022 by the adcc authors
+##
+## This file is part of adcc.
+##
+## adcc is free software: you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published
+## by the Free Software Foundation, either version 3 of the License, or
+## (at your option) any later version.
+##
+## adcc is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with adcc. If not, see <http://www.gnu.org/licenses/>.
+##
+## ---------------------------------------------------------------------
+import unittest
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+import adcc
+from .misc import expand_test_templates
+
+methods = ["adc0", "adc1", "adc2", "adc2x", "adc3"]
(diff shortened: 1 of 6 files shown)


adcc/smoketests/__init__.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
```
mfherbst, line 1: Why is this file needed at all?

adcc/smoketests/test_smoke.py
```diff
@@ -0,0 +1,91 @@
...
+        scfres = cached_backend_hf("pyscf", "h2o", "sto-3g")
+        refstate = adcc.ReferenceState(scfres)
+        state_singlets = adcc.run_adc(refstate, method=method,
+                                      n_singlets=3, conv_tol=1e-7)
+        state_triplets = adcc.run_adc(refstate, method=method,
+                                      n_triplets=3, conv_tol=1e-7)
+        assert_allclose(state_singlets.excitation_energy,
+                        _singlets[method], atol=1e-6)
+        assert_allclose(state_triplets.excitation_energy,
+                        _triplets[method], atol=1e-6)
```
mfherbst, line 80: Comparing the residual norm of the eigenpairs did not work?

adcc/smoketests/test_smoke.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
```
mfherbst, line 1: I would've just put this in the main folder as well or is there any reason I miss why this should be in a subfolder?

mfherbst (commented):
Thanks!

adcc/smoketests/__init__.py
maxscheurer: I wanted to invoke with `pytest --pyargs adcc.smoketests`, but can as well go for `pytest --pyargs adcc`. Probably the latter version is better (= shorter)...

adcc/smoketests/test_smoke.py
maxscheurer: Mhmm I don't see how this could be done from the `ExcitedStates` object... am I missing something?

mfherbst: can't you just compute M * v - λ * v and check the norm is small? (M = ADC matrix)

maxscheurer: Yes, I can 😄

maxscheurer:
Running the SCF manually now, because `adcc.testdata.static_data` is not installed, and I don't think we should install it 😄 Repeated H2O/sto-3g SCF does not cost that much, so it's not a problem.

adcc/test_smoke.py
```diff
@@ -0,0 +1,103 @@
...
+    return np.array([r @ r for r in residuals])
+
+
+@pytest.mark.skipif("pyscf" not in adcc.backends.available(),
+                    reason="PySCF not found.")
+@expand_test_templates(methods)
+class TestSmoke(unittest.TestCase):
+    conv_tol = 1e-7
+
+    def _run_scf_h2o_sto3g(self):
```
mfherbst, line 75: Make a comment why this and not using the cache.

adcc/test_smoke.py
```diff
@@ -0,0 +1,103 @@
...
+        scfres = adcc.backends.run_hf(
+            "pyscf", xyz="""
+            O 0 0 0
+            H 0 0 1.795239827225189
+            H 1.693194615993441 0 -0.599043184453037""",
+            basis="sto-3g", conv_tol=1e-12,
+            conv_tol_grad=1e-10
```
mfherbst, line 82: could make this depend on the class variable `conv_tol` as well.

mfherbst (approved):
Great!

maxscheurer:
Drat... one cannot install pyscf via pip in the `conda-build` test environment... So we need to use hard-coded reference state data.

maxscheurer:
BTW I just noticed that when a wheel dist is built using `python -m pip install . -vv` from adcc source directory, all the test files, including `conftest.py` are getting installed. The `MANIFEST` file is only acting on `sdist` builds, such that all our `conda` builds (which ran the above command or `python setup.py install`) include all the test files, even if they don't work in practice 😬 

Now I don't know how to **exclude** these specific files from the distribution, because the `exclude` statement in `find_packages` only acts on submodules, but not individual py files. 👎

mfherbst:
That's annoying ... I don't know a fix either. I would say, let's try very hard to find something and otherwise ... that's life. I mean one could an explicit `rm` in the conda build stage, deleting all files but the static `test_smoke.py` ... or even better maybe have a script in the adcc sources which does this (cleanup_tests or so).

maxscheurer:
Yes, I vote for the most pragmatic solution 😄  I hate these rabbit holes with deployment etc.

PR #145, opened 2022-01-25
Fix bug with pyscf module import

maxscheurer:
In #132, `pyscf` backend does not work anymore when `cppe` is not present because of top-level module import.

This also got past me during code review 😄

### DIFF
--- a/adcc/backends/pyscf.py
+++ b/adcc/backends/pyscf.py
@@ -30,7 +30,7 @@ from ..exceptions import InvalidReference
 from ..ExcitedStates import EnergyCorrection
 
 from pyscf import ao2mo, gto, scf
-from pyscf.solvent import pol_embed, ddcosmo
+from pyscf.solvent import ddcosmo
 
 
 class PyScfOperatorIntegralProvider:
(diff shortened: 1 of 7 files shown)


mfherbst (approved):
Jikes. This is because all of use have CPPE on our computers ;)

maxscheurer:
Probably best to make a patch release, because `0.15.12` is not really usable then without installing cppe 😦

maxscheurer:
I'm gonna try this in a cppe-free setup first, to make sure everything is fine again.

maxscheurer:
Okay it seems there are some problems here to sort out with the way things get imported from `pyscf`... That's niiiiice 🙈

maxscheurer:
Still investigating... I'm at the point where I'm wondering why it **was** even working at some point...

mfherbst:
Hmmm. Maybe `pyscf` has restructured their imports. They did a major version bump, so that might have actually happened.

maxscheurer:
It's actually also broken with 1.7.6. I think I'll have a solution soon 😄

adcc/backends/pyscf.py
```diff
@@ -59,28 +59,34 @@ def nabla(self):
...
-            if isinstance(self.scfres.with_solvent, pol_embed.PolEmbed):
-                def pe_induction_elec_ao(dm):
-                    return self.scfres.with_solvent._exec_cppe(
-                        dm.to_ndarray(), elec_only=True
-                    )[1]
-                return pe_induction_elec_ao
+        if not hasattr(self.scfres, "with_solvent"):
+            return None
+        if not hasattr(self.scfres.with_solvent, "cppe_state"):
+            return None
```
mfherbst, line 65: Could also provoke an attributeerror by directly accessing self.scfres.with_solvent.cppe_state and if it is encountered return None.

maxscheurer: Yes, that's more pythonic 👍

maxscheurer: Can't do the same for COSMO though since there's no equivalent state object on `with_solvent` in that case + need to make sure we have the correct environment model at that spot.

mfherbst: ok

PR #149, opened 2022-02-28
Projectors and projected ADC matrix

mfherbst:


adcc/MoSpaces.py
```diff
@@ -110,7 +116,7 @@ def expand_to_list(space, entry, from_min=None, from_max=None):
     if key in spaceargs and spaceargs[key]:
         spaceargs[key] = np.concatenate((
             expand_to_list(key, spaceargs[key][0], from_max=noa),
-            expand_to_list(key, spaceargs[key][1], from_max=noa) + noa
+            expand_to_list(key, spaceargs[key][1], from_max=nob) + noa
```
maxscheurer, line 119: Was this a bug?

adcc/guess/__init__.py
```diff
@@ -28,6 +28,23 @@
...
+    computed states of the passed excitation `kind`.
+    """
+    if kind == "singlet":
+        return dict(spin_block_symmetrisation="symmetric", spin_change=0)
+    elif kind == "triplet":
+        return dict(spin_block_symmetrisation="antisymmetric", spin_change=0)
+    elif kind == "spin_flip":
+        return dict(spin_block_symmetrisation="none", spin_change=-1)
+    elif kind == "any":
+        return dict(spin_block_symmetrisation="none", spin_change=0)
```
maxscheurer, line 43: More pythonic to use a dict for the mapping instead?

adcc/test_AdcMatrix.py
```diff
@@ -319,4 +319,64 @@ def template_matmul(self, case):
...
+                              spin_block_symmetrisation=spin_block_symmetrisation)
+        vec.set_random()
+        pvec = projected.apply_projection(vec.copy())  # only apply projection
+
+        pres = projected @ vec
+        ores = matrix @ pvec
+        res_for_sym = matrix @ vec
+
+        assert_equal_symmetry(res_for_sym.ph, pres.ph)
+        assert_equal_symmetry(res_for_sym.pphh, pres.pphh)
```
maxscheurer, line 378: You mentioned above that `assert_equal_symmetry` is not super reliable? Is this not a problem here?

maxscheurer (approved):
Just some questions, otherwise looks great! Thanks for the example script, too 🚀

t-fransson (approved):

adcc/MoSpaces.py
mfherbst: yes, I think so.

adcc/guess/__init__.py
mfherbst: indeed.

adcc/test_AdcMatrix.py
mfherbst: Not quite, because passing `assert_equal_symmetry` is a sufficient condition for the symmetry to be actually identical, i.e. it is a more strict check then necessary.

PR #153, opened 2022-08-09
Full and approximate polaritonic ADC energies up to second order

BauerMarco:
In the following polaritonic is always referred to by QED (Quantum Electrodynamics)

**Features:**

- QED-MP up to second order from QED-HF reference and standard HF
- QED-ADC up to second order from QED-HF
- QED-ADC up to first order from standard HF
- Method, which approximates QED-ADC up to second order from QED-HF

All features have been tested thoroughly, although only the methods based on a QED-HF reference should be used, since they are available up to second order.
One can also request QED-ADC(2) from a QED-HF reference with the polaritonic coupling only expanded up to first order, which is correct, but it is not sure, how useful this feature really is...This has to be tested in a future benchmark.

**New Keywords:**

- coupl: Coupling strength to the cavity mode
- freq: Energy of the cavity mode
- qed_hf: Whether a QED-HF reference is provided or not (default is True)
- qed_approx: Use standard or approximated method (default is False, but should be set to True after a proper benchmark)
- qed_full_diag: If full diagonalization of the QED matrix is requested (with qed_approx = False), you should use this guess

**Additional comments:**

- The tests do not use a QED-HF reference, but still use qed_hf = True, so they can still be tested with any of the possible backend options adcc provides.
- As far as I know, there are two QED-HF reference implementations available from open source, which are the hilbert package developed by the DePrince group, which is a plugin to the psi4 programm package, and an other quite basic implementation in the Polaritonic section of the psi4numpy, which is meant for educational purposes, but can be used just as well as the hilbert package. I will provide some examples for the website, after the reviewing process is done.
- I should mention, that the corresponding paper is not published yet, so I could not reference it at the appropriate places in the code.

### DIFF
--- a/adcc/qed_matrix_from_diag_adc.py
+++ b/adcc/qed_matrix_from_diag_adc.py
@@ -21,7 +21,7 @@
 ##
 ## ---------------------------------------------------------------------
 import numpy as np
-import scipy.linalg as sp
+#import scipy.linalg as sp
 
 
 class qed_matrix_from_diag_adc:
@@ -76,10 +76,11 @@ class qed_matrix_from_diag_adc:
                             matrix_middle.reshape((len(matrix_middle), 1)),
                             matrix_lower))
 
-        if np.iscomplex(self.full_freq):
-            return sp.eig(matrix)
-        else:
-            return sp.eigh(matrix)
+        #if np.iscomplex(self.full_freq):
+        #    return sp.eig(matrix)
+        #else:
+        #    return sp.eigh(matrix)
+        return matrix
 
     def second_order_coupling(self):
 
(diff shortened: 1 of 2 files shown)


maxscheurer:
Thanks for the contribution! This will, however, take some time to digest, but there are probably ways to make the integration of QED less "invasive" in the existing code structure, right @mfherbst? 👍

mfherbst:
Cool! Thanks for the contribution @BauerMarco.

Indeed, quite a bit of code. Some looks like duplications that can be avoided. I have to admit that I lack the time to review the code without some explanation what's going on. Would be good to have a chat at some point.

BauerMarco:
> Cool! Thanks for the contribution @BauerMarco.
> 
> Indeed, quite a bit of code. Some looks like duplications that can be avoided. I have to admit that I lack the time to review the code without some explanation what's going on. Would be good to have a chat at some point.

I am available after the STC ends (22nd of september) and would very much appreciate to have a chat.

BauerMarco (commented):
initial review cycle

PR #158, opened 2022-11-04
Transition dipole moments for asymmetric operators

apapapostolou:
We have found that the second-order transition density matrices from the ground state, as currently implemented, can only be used for symmetric operators. For asymmetric operators, such as the magnetic dipole operator, we found a small error in the oo block.

Other changes:
- generalized the `modified_transition_moments` function for arbitrary one-particle operators (the previous implementation was only for symmetric ones)
- renamed `dipole_operator` to `operator` in the `modified_transition_moments` function
- extended the tests to the magnetic dipole operator

The equations were taken from J. Schirmer, Phys. Rev. A 26, 2395 (1982), and we also compared them to the equations @jonasleitner created with his code generator.

In addition, we believe that a factor of 0.5 is missing in one of the terms in the vc block of the cvs transition dm. This is also missing in the MTMs in the corresponding term. Since this hardly affects the oscillator strengths, it was probably not noticed before.

ToDo: If the changes are approved, the reference data (`state.transition_magnetic_dipole_moment` and `state.transition_dipole_moment_velocity`) must be adjusted accordingly.


adcc/adc_pp/test_modified_transition_moments.py
```diff
@@ -34,46 +35,58 @@
...
 
-        mtms = modified_transition_moments(method, state.ground_state)
+        if op_kind == "electric":
+            dips = state.reference_state.operators.electric_dipole
+            ref_tdm = state.transition_dipole_moment
+        elif op_kind == "magnetic":
+            dips = state.reference_state.operators.magnetic_dipole
+            ref_tdm = state.transition_magnetic_dipole_moment
+        else:
+            raise NotImplementedError(
```
maxscheurer, line 54: better to use `pytest.skip` in that case?

adcc/adc_pp/transition_dm.py
```diff
@@ -69,7 +69,7 @@ def tdm_cvs_adc2(mp, amplitude, intermediates):
     )
 
     # cvs_adc2_dp0_vc
-    dm.vc = u1.transpose() - einsum("ab,Ib->aI", p0.vv, u1)
+    dm.vc -= 0.5 * einsum("ab,Ib->aI", p0.vv, u1)
```
maxscheurer, line 72: Is `u1` accounted for by the lower-order tdm?

maxscheurer (commented):
Thanks a lot for this!!! 🚀 @mfherbst what is the best way to update the reference data from `dump_reference_adcc.py`? We need to update them because the deposited data for consistency checks of magnetic moments etc. are wrong 😄

adcc/adc_pp/transition_dm.py
apapapostolou: Exactly. In line 57, the zero-order contribution is already calculated, but in the previous implementation, this was then not used, but recalculated in line 72.

maxscheurer:
@apapapostolou can you maybe upload the testdata files here that need to be updated so @mfherbst can upload them to the server? Would that work in principle?

mfherbst:
> Would that work in principle?

Yes. I'd be happy to do that. I could also download them from somewhere (in NRW we have something like an owncloud for such things. Maybe you have something similar in BaWü).

mfherbst:
I remember we had some tests which were disabled for some low-order ADC variants because they always failed. Could have something to do with this and it might be worth checking if it's now possible to activate them. They are marked as xfail in the test suite I think.

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -30,47 +30,54 @@
 from adcc.AmplitudeVector import AmplitudeVector
 
 
-def mtm_adc0(mp, dipop, intermediates):
-    return AmplitudeVector(ph=dipop.ov)
+def mtm_adc0(mp, op, intermediates):
+    return AmplitudeVector(ph=op.ov) if op.is_symmetric \
+        else AmplitudeVector(ph=op.vo.transpose())
```
mfherbst, line 35: If you have to break the one-line if statement it's better to not have a one-line if statement and go for an if-else block in my opnion

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -30,47 +30,54 @@
...
 
 
-def mtm_adc1(mp, dipop, intermediates):
-    f1 = dipop.ov - einsum("ijab,jb->ia", mp.t2(b.oovv), dipop.ov)
+def mtm_adc1(mp, op, intermediates):
+    f1 = op.ov if op.is_symmetric else op.vo.transpose()  # zeroth order
```
mfherbst, line 39: Better not repeat. Just call `mtm_adc0` to reuse the expression from above. Also makes the structure of the equations and the orders more explicit. If I remember correctly we do this for the matrix equations as well. (have not checked)

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -30,47 +30,54 @@
...
+    f1 -= einsum("ijab,jb->ia", mp.t2(b.oovv), op.ov)  # first order
     return AmplitudeVector(ph=f1)
 
 
-def mtm_adc2(mp, dipop, intermediates):
+def mtm_adc2(mp, op, intermediates):
     t2 = mp.t2(b.oovv)
     p0 = mp.mp2_diffdm
 
+    op_vo = op.ov.transpose() if op.is_symmetric else op.vo
```
mfherbst, line 48: same story, better call above functions.

adcc/adc_pp/test_modified_transition_moments.py
```diff
@@ -21,59 +21,72 @@
...
+            dips = state.reference_state.operators.electric_dipole
+            ref_tdm = state.transition_dipole_moment
+        elif op_kind == "magnetic":
+            dips = state.reference_state.operators.magnetic_dipole
+            ref_tdm = state.transition_magnetic_dipole_moment
+        else:
+            skip(
+                "Tests are only implemented for"
+                "electric and magnetic dipole operators."
+            )
```
mfherbst, line 57: could compress this into 2 lines or not?

adcc/adc_pp/test_modified_transition_moments.py
```diff
@@ -21,59 +21,72 @@
...
+        mtms = modified_transition_moments(method, state.ground_state, dips)
+
         for i in range(n_ref):
             # computing the scalar product of the eigenvector
             # and the modified transition moments yields
             # the transition dipole moment (doi.org/10.1063/1.1752875)
             excivec = state.excitation_vector[i]
-            res_tdm = -1.0 * np.array([excivec @ mtms[i] for i in range(3)])
-            ref_tdm = ref["transition_dipole_moments"][i]
+            res_tdm = np.array([excivec @ mtms[i] for i in range(3)])
```
mfherbst, line 65: Nice. Feels much better.

apapapostolou:
> I remember we had some tests which were disabled for some low-order ADC variants because they always failed. Could have something to do with this and it might be worth checking if it's now possible to activate them. They are marked as xfail in the test suite I think.

The functionality tests for the CN ADC(0) transition dipole moments still fail, even with the new reference data.

apapapostolou:
I have now updated `dump_adcc_reference.py` to regenerate all reference data with `adcc`. The hdf5 files are named like the others, only `reference` was replaced by `adcc_reference`.

Since not all `hfdata` contained the magnetic dipole operator and the nabla operator, I also wanted to regenerate the `hfdata`. However, this has caused problems with some tests that still use the `atd` reference data, as these had been generated with the old `hfdata`. I have now solved the problem by continuing to use the old `hfdata` and only patching in the missing operators.

apapapostolou:
> Yes. I'd be happy to do that. I could also download them from somewhere (in NRW we have something like an owncloud for such things. Maybe you have something similar in BaWü).

I could share the new reference data with you via heiBOX if that's okay with you. The file is too big to upload here.

maxscheurer (approved):
Great, thanks for this important contribution 🚀

PR #160, opened 2022-11-21
Solvation effects via DDX

mfherbst:
So far just a crude copy n' paste implementation. Let's see what psi4 settles on after review and then do this properly.

- [x] Wait for https://github.com/psi4/psi4/pull/2767 to get merged
- [x] Wait for https://github.com/psi4/psi4/pull/2967
- [x] Refactor
- [x] Fix convergence issues with state-specific solvation via DDX (probably due to initial guess handling in Psi4)
- [x] Add tests


mfherbst:
That would be awesome @maxscheurer! Much appreciated :+1:.

mfherbst:
@maxscheurer Should be ready to go now. Tests pass on my machine, but let's see what the CI says.

PR #165, opened 2023-04-27
Newer conda gcc

maxscheurer:


maxscheurer:
Coveralls seems to have some server issues....

PR #166, opened 2023-05-24
Try psi4 on Conda CI

maxscheurer:
New version of #137.


maxscheurer:
> I think pulling in a dev build could get pretty brittle.

I agree 🫠

PR #187, opened 2024-10-30
more flexible cached_member decorator

jonasleitner:
There are 2 problems with the current version (see example below):
- It is not possible to call a decorated function with keyword arguments (quality of life)
- The decorator does not work with default arguments and might load unexpected results from the cache.

Neither of the points is currently a problem. Just to have it sorted for the future.

```
from adcc.misc import cached_member_function

class Test:
    @cached_member_function
    def foo(self, x, y = 1):
        print(x, y)

t = Test()

# this will not work
t.foo(x=2)

# this will print twice because the default argument
# is not correctly considered
t.foo(0)  # key for cache lookup: (0,)
t.foo(0, 1)  # key for cache lookup: (0, 1)
```

### DIFF
--- a/adcc/misc.py
+++ b/adcc/misc.py
@@ -22,6 +22,7 @@
 ## ---------------------------------------------------------------------
 import warnings
 import numpy as np
+import inspect
 from functools import wraps
 from pkg_resources import parse_version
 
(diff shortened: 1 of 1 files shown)


jonasleitner:
@apapapostolou @maxscheurer can one of you take a quick look?
Since I'm using this code snippet in the equation generator I am pretty sure everything should work.

maxscheurer:
I'm gonna play the bad guy 😂
Can you please add a unit test that tests the functionality and failures of the edge cases, e.g. kwarg only etc? Just a suggestion...

jonasleitner:
Should have seen that one coming 😂👌
I think this should cover most cases. Did I miss something?

maxscheurer (approved):

PR #189, opened 2025-01-07
Refactor tests

jonasleitner:
Changes:
- Added a `TestCase` class to collect the metadata for all test cases (geometry, charge, reference cases, ...) centralized.
- A new Interface to Q-Chem/adcman for the test data generation.
- New test data generators for HF and ADC reference data.
- Restructured the test data hdf5 files:
    - MP data is separated from the ADC data to avoid duplication.
    - The data for all reference cases (cvs, fc, ...) and ground state density orders (will be introduced in the `density_order` branch) is collected in a single file to reduce the number of files.
- Adapted the test data cache and the tests to the new infrastructure.
- Switched to `pytest.mark.parametrize` for the parametrization of tests.
- Tried to unify tests by reducing the number of special cases when possible.

While most of the tests itself are actually unchanged, the cases for which to run a certain test were modified for quite some tests. I tried to consistently cover all relevant cases. However, certainly I might have missed some cases. Would be good if you could also have a look @apapapostolou  @AdrianLDempwolff  @maxscheurer . Probably it is sufficient to verify tests that were originally parametrized with the `expand_test_templates` decorator.

Other changes:
- While adapting the veloxchem tests, @frieschneider found that the veloxchem interface was not functional anymore and updated it. Unrestricted calculations are now also possible with veloxchem.

Todo:
- [x] Decide where to store the test data and upload the data.
- [x] Add scripts to download and remove test data.
- [x] Fix CI (not all errors seem to be related to the introduced changes).


tests/generators/generate_hfdata.py
```diff
@@ -0,0 +1,125 @@
...
+        return None
+    print(f"Generating data for {test_case.file_name}.")
+    mf = run_pyscf(test_case, frac_occ)
+    hdf5_file = h5py.File(hdf5_file, "w")
+    dump_pyscf(mf, hdf5_file)
+    return hdf5_file
+
+
+def generate_ch2nh2():
+    cases = testcases.get(n_expected_cases=1, name="ch2nh2", basis="sto-3g").pop()
```
apapapostolou, line 67: Why is `n_expected_cases` needed?

jonasleitner: The idea is to raise an exception if another `TestCase` that matches your request is added. For instance you want to retrieve the `h2o_sto3g` and `h2o_def2tzvp` cases with `testcases.get(name="h2o")`, but later `h2o_ccpvdz` is added and now the request will return all 3 cases, which is something you might not want at this point.
The `get` function is from an earlier iteration anyway. Most of the code is now using `get_by_filename`, which similarly raises an exception if it can't find a `TestCase` for each filename.

tests/ExcitedStates_test.py
```diff
@@ -20,20 +20,21 @@
...
 from adcc.ExcitedStates import EnergyCorrection
-from .test_state_densities import Runners
 
+from .testdata_cache import testdata_cache
+
+
+class TestExcitedStates:
+    def test_excitation_view(self):
+        state = testdata_cache.adcc_states(
+            system="h2o_sto3g", method="adc2", case="gen", kind="singlet"
```
apapapostolou, line 35: Shouldn't CVS be tested here as well? And what about the other ADC methods? Before, everything was tested here.

tests/ReferenceState_refdata_test.py
```diff
@@ -20,18 +20,28 @@
...
+from . import testcases
 
 
-def compare_refstate_with_reference(data, reference, spec, scfres=None,
-                                    compare_orbcoeff=True, compare_eri="value"):
+def compare_refstate_with_reference(system: str, case: str,
+                                    data: dict, reference: dict,
+                                    scfres=None, compare_orbcoeff: bool = True,
+                                    compare_eri: str = "value"):
+    # - data is the data to build a ReferenceState on top. If scref is given
```
apapapostolou, line 36: typo: scfres?

tests/properties_test.py
```diff
@@ -0,0 +1,244 @@
...
+        res_dms = state.transition_dipole_moment_velocity
+        n_ref = len(state.excitation_vector)
+        assert_allclose(
+            res_dms, refdata["transition_dipole_moments_velocity"][:n_ref],
+            atol=1e-4
+        )
+
+    # Only adcc reference data available.
+    @pytest.mark.parametrize("system,case,kind", cases)
+    def test_rotatory_strengths(self, system: str, case: str, kind: str,
```
apapapostolou, line 227: This should be zero for everything that is not chiral. Should we also test it for something chiral like methyloxirane?

apapapostolou (approved):
After reviewing all the changed files, I would say that this PR is ready to be merged as soon as the new reference data is uploaded. @jonasleitner, thank you very much for all the work you have put in!

But before that, the CI should be fixed, which is independent of this PR, I think.

tests/ExcitedStates_test.py
jonasleitner: Here we don't actually test the implemented equations. We only test that the `Excitation` class gives the same result as the `ExcitedStates` class. Since `Excitation` just calls out to the `ExcitedStates` implementation, this test should in my opinion be independent of reference case, state kind and system (restricted/unrestricted). Therefore, I decided to only verify a single test case.

tests/properties_test.py
jonasleitner: You are the expert 😄
Mhhh... In this test, we verify that the rotatory strength is correctly calculated from the electronic and magnetic transition dipole moments. In principle testing for a zero result is also a verification. But testing for a nonzero result is probably better.
However, at the moment we don't have any MP/ADC reference data for methyloxirane. So we would have to generate, store and maintain the test data exclusively for this test.

apapapostolou: Okay, then let's leave it for now.

maxscheurer:
I'm not a big fan of the `tests` folder outside of the source repository (mostly due to personal taste, but that does not count). One problem I see though is that we rely on the `test_smoke` to be present for tests on the adcc conda-forge feedstock. The reason why tests need to be installed along with the Python package there is that the build and test happens in two different environments.

See also https://github.com/conda-forge/adcc-feedstock/blob/main/recipe/build.sh
and https://github.com/adc-connect/adcc/blob/master/scripts/remove_test_files.sh

As a solution, one could just copy the `test_smoke` file into the adcc source before the install command (that would have to happen on the feedstock side). 😁

jonasleitner:
Thanks for pointing that out! I will have a look at it, though I have no experience with conda-forge, yet.

> I'm not a big fan of the tests folder outside of the source repository (mostly due to personal taste, but that does not count).

While I personally prefer separating the tests from the source, I agree that this change seems rather random and unnecessary. The ability to easily track my progress was the main reason to introduce it. But I have no strong opinion on keeping the tests folder. There was simply no reason or opposition until now. Depending on how easy it is to adjust the conda feedstock, we can also move the tests back to the source.

maxscheurer (approved):

PR #195, opened 2025-04-06
Restructure ExcitedStates etc for IP/EA Integration

jonasleitner:
This PR includes some preliminary work on the `ElectronicTransition`, `ExcitedStates` and `Excitation` classes for https://github.com/adc-connect/adcc/pull/168 by restructuring the classes to easily enable the addition of other ADC schemes.

**Changes**
- Introduce a new `ElectronicStates`  base class which implements methods that should be common to all ADC variants
- Reworked the plotting and the describe methods.
- Introduced methods for all densities and properties that compute the corresponding quantity only for a single state. This enables a (potentially) more efficient implementation of the `Excitation` class, since properties can now be evaluated only for a single state, while maintaining the interface.
- Operator integrals are now cached in the MO basis (on `OperatorIntegrals`) to enable the per state calculation of properties.
- Introduced a `StateView` base class from which the views for other ADC variants can inherit and explicitly implemented the available methods on the classes so it is more clear what methods available.


jonasleitner:
@frieschneider and I thought that it might be a good idea to introduce a generic `block_orders` generation and verification for the `AdcMatrix` and `IsrMatrix`. This way we can more easily add higher order methods or other ADC variants (IP/EA/DIP/...).

maxscheurer (approved):
rubber stamp approval 😁

PR #196, opened 2025-06-04
update `pkg_resources`

loriab:
I realized adcc is calling `pkg_resources` when setuptools bumped to v80.9 and started throwing a warning (below) whenever adcc is imported (as psi4 does whenever possible), which interferes with QCEngine importing, which in turn makes tests fail.

```
pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.

    -----------------------------------------------------------------------
          Psi4: An Open-Source Ab Initio Electronic Structure Package
                               Psi4 1.[10](https://github.com/psi4/psi4/actions/runs/15426225369/job/43413689202?pr=3303#step:20:11)a1.dev[11](https://github.com/psi4/psi4/actions/runs/15426225369/job/43413689202?pr=3303#step:20:12)9
```

This does interfere with QCEngine detection, but that's easy to paper over in the medium term. This PR replaces the troublesome import the same way QCElemental did (https://github.com/MolSSI/QCElemental/pull/341). QCEl needed a follow-up PR https://github.com/MolSSI/QCElemental/pull/353, but that was for parsing unusual versions, which I don't think is the case here.

btw, that `from packaging.version import parse` needs to be exactly that, iirc. No variations in import statements.

EDIT: unimportant, no change in info.


maxscheurer:
@loriab should we mint a new release after merging?

loriab:
> @loriab should we mint a new release after merging?

That'd be nice. Otherwise I'd suggest a patch and bumping the build number in the conda recipe.

The lint error `./adcc/backends/__init__.py:38:5: F824 `global __status` is unused: name is never assigned in scope` doesn't seem related to my changes. Should I look into it anyway?

jonasleitner:
> The lint error `./adcc/backends/__init__.py:38:5: F824 `global __status` is unused: name is never assigned in scope` doesn't seem related to my changes. Should I look into it anyway?

No, this is not related to your PR. I think I addressed this in https://github.com/adc-connect/adcc/pull/195. So I think you can just ignore it for now

maxscheurer (approved):

PR #198, opened 2025-08-12
Fix editable install after introducing pyproject.toml

jonasleitner:
After moving to `pyproject.toml` @frieschneider found that imports from `libadcc` fail for an editable install of `adcc` when python is not executed from the top level project directory.
This is the result of the changed mechanism of editable installs, since they are now build following [PEP660](https://peps.python.org/pep-0660/):
- Previously the top level project directory was simply added to `sys.path`
- Now a custom `MetaPathFinder` and `PathEntryFinder` are generated to resolve all imports from the (editable) installed package

This change leads to the observed behavior: `import libadcc` imports the `libadcc` source folder as python namespace package instead of the `libadcc.[...].so` shared object - if the top level project directory is not available in `sys.path`.
Without going to much into detail, this problem can be resolved by renaming the `libadcc` source folder to basically anything else (or by moving it to another place).
In this PR @frieschneider and I suggest renaming it to `libadcc_src`.


frieschneider (approved):
I noticed one small thing in `docs/references.rst` and adjusted it myself. Other than that, everything looks good. 👍

PR #208, opened 2026-04-14
Refactor AdcMethod + ISR(1)-d

frieschneider:
###  Refactor `AdcMethod.py`:
- explicitly differentiate between ADC calculation (`AdcMethod`) and property calculation (`IsrMethod`)
- method validation does not rely anymore on creating a list with all valid methods
- added some minimal tests

### Include ISR(1)-d:
densities:
- reorganized different terms
- used already available `check_doubles_amplitudes` function to assess whether doubles excitation are present (this function throws a `ValueError` -> try except layout)
- renaming of the functions (e.g. `s2s_tdm_adc2` is now called `s2s_tdm_isr2`)

Modified transition moments and B matrix:
- ISR(1)-d B matrix: `block_orders` need to be explicitly set. Update `_default_block_orders`?
- renaming (adc -> isr)

no test yet for ISR(1)-d

### DIFF
--- a/adcc/AdcMatrix.py
+++ b/adcc/AdcMatrix.py
@@ -22,13 +22,14 @@
 ## ---------------------------------------------------------------------
 import itertools
 import numpy as np
+from typing import Union
 
 import libadcc
 
 from .LazyMp import LazyMp
 from .adc_pp import matrix as ppmatrix
 from .timings import Timer, timed_member_call
-from .AdcMethod import AdcMethod
+from .AdcMethod import AdcMethod, IsrMethod
 from .functions import ones_like
 from .Intermediates import Intermediates
 from .AmplitudeVector import AmplitudeVector
@@ -72,18 +73,22 @@ class AdcMatrixlike:
     """
 
     _special_block_orders = {
-        "adc2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
+        "2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
     }
 
     @classmethod
-    def _default_block_orders(cls, method: AdcMethod) -> dict[str, int]:
+    def _default_block_orders(cls,
+                              method: Union[AdcMethod, IsrMethod]
+                              ) -> dict[str, int]:
         """
         Determines the default block orders for the given adc method.
         """
         # check if we have a special method like adc2x
         # I guess base_method should also contain the adc_type prefix so
         # we don't need to separate different adc_types
-        block_orders = cls._special_block_orders.get(method.base_method.name, None)
+        block_orders = cls._special_block_orders.get(
+            method.base_method.name[3:], None
+        )
         if block_orders is not None:
             return block_orders.copy()
         # otherwise assume that we have a "normal" PP/IP/...-ADC(n) method
--- a/adcc/AdcMatrix.py
+++ b/adcc/AdcMatrix.py
@@ -28,7 +28,7 @@ import libadcc
 from .LazyMp import LazyMp
 from .adc_pp import matrix as ppmatrix
 from .timings import Timer, timed_member_call
-from .AdcMethod import AdcMethod, Method
+from .AdcMethod import AdcMethod, IsrMethod, Method
 from .functions import ones_like
 from .Intermediates import Intermediates
 from .AmplitudeVector import AmplitudeVector
@@ -73,6 +73,7 @@ class AdcMatrixlike:
 
     _special_block_orders = {
         "adc2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
+        "isr1s": {"ph_ph": 1, "ph_pphh": None, "pphh_ph": None, "pphh_pphh": None},
     }
 
     @classmethod
@@ -100,10 +101,23 @@ class AdcMatrixlike:
             raise ValueError(f"Unknown adc type {method.adc_type} for method "
                              f"{method.name}. Can not determine default block "
                              "orders.")
-        spaces = [
-            "p" * i + min_space + "h" * i
-            for i in range(0, (method.level.to_int() // 2) + 1)
-        ]
+        # ADC matrices have first-order coupling whereas ISR matrices
+        # have a zeroth-order coupling between adjacent excitation classes
+        # see https://doi.org/10.1063/1.1752875
+        if isinstance(method, AdcMethod):
+            # First-order coupling between adjacent excitation classes.
+            spaces = [
+                "p" * i + min_space + "h" * i
+                for i in range(0, (method.level.to_int() // 2) + 1)
+            ]
+        elif isinstance(method, IsrMethod):
+            # Zeroth-order coupling between adjacent excitation classes.
+            spaces = [
+                "p" * i + min_space + "h" * i
+                for i in range(0, ((method.level.to_int() + 1) // 2) + 1)
+            ]
+        else:
+            raise ValueError(f"Invalid method: {method.name}")
         # exploit the fact that the spaces are sorted from small to high:
         # If we walk the adc matrix in any direction we always have to subtract 1!
         # Therefore, we can determine the order according to the position of the
@@ -113,7 +127,8 @@ class AdcMatrixlike:
         for ((i1, bra), (i2, ket)) in \
                 itertools.product(enumerate(spaces), repeat=2):
             order = method.level.to_int() - i1 - i2
-            assert order >= 0
+            # For ISR matrices allow missing diagonal blocks.
+            order = None if order < 0 else order
             ret[f"{bra}_{ket}"] = order
         return ret
 
@@ -142,6 +157,7 @@ class AdcMatrixlike:
         for block, order in block_orders.items():
             if order is None:
                 continue
+            assert order >= 0
             # ensure that the block is valid for the given adc type
             bra, ket = block.split("_")
             if not cls._is_valid_space(bra, method) or \
--- a/adcc/AdcMatrix.py
+++ b/adcc/AdcMatrix.py
@@ -22,14 +22,13 @@
 ## ---------------------------------------------------------------------
 import itertools
 import numpy as np
-from typing import Union
 
 import libadcc
 
 from .LazyMp import LazyMp
 from .adc_pp import matrix as ppmatrix
 from .timings import Timer, timed_member_call
-from .AdcMethod import AdcMethod, IsrMethod
+from .AdcMethod import AdcMethod, Method
 from .functions import ones_like
 from .Intermediates import Intermediates
 from .AmplitudeVector import AmplitudeVector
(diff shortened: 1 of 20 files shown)


adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -43,9 +43,24 @@ def s2s_tdm_adc0(mp, amplitude_l, amplitude_r, intermediates):
     return dm
 
 
-def s2s_tdm_adc2(mp, amplitude_l, amplitude_r, intermediates):
+def s2s_tdm_isr1(mp, amplitude_l, amplitude_r, intermediates):
```
jonasleitner, line 46: This auto selects ISR(1)-d for vectors with a doubles part, which is good, since it is the correct way to calculate first order properties in these cases. However, currently it is not possible to calculate "conventional" ISR(1) properties that ignore the doubles part at all. (Ignoring the fact that you can ask for ISR(0) properties, since the first order contribution to the singles-singles block is zero for one-particle operators).
I am not entirely sure whether we want to enable this via some flag or special method string or something like that. What do you think?
(same for the other densities)

adcc/adc_pp/state2state_transition_dm.py
jonasleitner, line 100: Aren't these contributions handled by the ISR(1) function right now?

adcc/adc_pp/state_diffdm.py
```diff
@@ -96,14 +109,29 @@ def diffdm_adc2(mp, amplitude, intermediates):
...
+        p2_ov = -sqrt(2) * einsum("jb,ijab->ia", u1, u2)
+        dm.ov = p2_ov
+    except ValueError:
+        # no doubles contribution
+        pass
+    return dm
+
+
+def diffdm_cvs_isr2(mp, amplitude, intermediates):
+    dm = diffdm_isr0(mp, amplitude, intermediates)  # Get ISR(1) result
```
jonasleitner: Here you are still calling `isr0`?

adcc/adc_pp/state_diffdm_2p.py
```diff
@@ -42,7 +42,7 @@ def diffdm_adc0_2p(mp, amplitude, intermediates):
 
     dm = TwoParticleDensity(mp, symmetry=OperatorSymmetry.HERMITIAN)
 
-    # ADC(1) diffdm
+    # ISR(1) diffdm
```
jonasleitner:
```suggestion
    # one-particle ISR(0) diffdm
```

adcc/adc_pp/state_diffdm_2p.py
```diff
@@ -69,10 +69,10 @@ def diffdm_adc1_2p(mp, amplitude, intermediates):
 
     t2 = mp.t2(b.oovv)
     # TODO move to intermediates!
-    # ADC(1) diffdm
+    # ISR(1) diffdm
```
jonasleitner:
```suggestion
    # one-particle ISR(0) diffdm
```

adcc/tests/AdcMethod_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_adcmethod(self):
+        method = adcc.AdcMethod("adc2")
+        cvs_method = adcc.AdcMethod("cvs-adc2")
+
+        assert method.name == cvs_method.base_method.name
+
+        method_new_level = method.at_level(1)
+        assert method_new_level._method_base_name == method._method_base_name
+        assert method_new_level.level == 1
```
jonasleitner: Maybe also test the conversion to IsrMethod and vice versa?

adcc/AdcMatrix.py
```diff
@@ -72,18 +73,22 @@ class AdcMatrixlike:
...
     _special_block_orders = {
-        "adc2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
+        "2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
     }
 
     @classmethod
-    def _default_block_orders(cls, method: AdcMethod) -> dict[str, int]:
+    def _default_block_orders(cls,
+                              method: Union[AdcMethod, IsrMethod]
+                              ) -> dict[str, int]:
```
jonasleitner:
```suggestion
    def _default_block_orders(cls, method: Method) -> dict[str, int]:
```

adcc/AdcMatrix.py
```diff
@@ -72,18 +73,22 @@ class AdcMatrixlike:
...
         """
         Determines the default block orders for the given adc method.
         """
         # check if we have a special method like adc2x
         # I guess base_method should also contain the adc_type prefix so
         # we don't need to separate different adc_types
-        block_orders = cls._special_block_orders.get(method.base_method.name, None)
+        block_orders = cls._special_block_orders.get(
+            method.base_method.name[3:], None
```
jonasleitner: Mhhh... in principle level on the `Method` class should store "2x" or integers, since it is perfectly valid to define 2x densities and so on. Maybe we can change this and add isr2x to the dispatches (they should still compute isr2 properties)? Then we can just use level for the dict lookup here.
Does that also mean that we currently compute 2x properties for the B-matrix and ISR(2) properties for the densities by defaults?

adcc/AdcMethod.py
```diff
@@ -20,76 +20,82 @@
...
+        level = split[-1][len(self._method_base_name):]
+        if level == "2x":
+            self.level: int = 2
+        elif not level.isnumeric:
+            raise ValueError(f"{level} is not a valid method level")
         else:
-            return AdcMethod("adc" + str(newlevel))
+            self.level: int = int(level)
+
+        self._base_method: str = self._method_base_name + level
```
jonasleitner: if we also store "2x" in level, it would be less error prone to make `_base_method` a property, that builds the string based on `_method_base_name` and `level` on the fly and thus automatically propagates changes to either of them all the way through. But I don't know how many changes would be required to also allow a `str` as level

adcc/adc_pp/state2state_transition_dm.py
frieschneider: These contributions are still used as intermediates in the ISR(2) equations, see for example:

```
dm.ov = (
        p1_ov
        - einsum("ijab,bj->ia", t2, p1_vo)
        ...
)
```

What is your opinion? Should I rewrite the S2S TDM with in-place addition, or should we leave it as it is? However, this would result in quite long files if we add ISR(3) and all contributions are always repeated.

adcc/AdcMatrix.py
frieschneider: yes, that is indeed the case, we didnt realise because "adc2x" is not a testcase for `TestIsrMatrix`: only adc[0-2] is tested. 

What should we do about it? Implement adc2x densities? Or should we use an ISR(2) B matrix by default?

adcc/adc_pp/state2state_transition_dm.py
jonasleitner: Yes you are right. The S2S TDM seems to be the only density, which is currently not using in-place addition. Doing so should only introduce negligible computation costs. So think it should be fine to switch to in-place addition 👍🏻

adcc/AdcMatrix.py
jonasleitner: I think we don't want to alter the default for the densities right now (though we probably should have a look at that in the future). So I think using an ISR(2) B matrix should be fine for now?

adcc/adc_pp/state2state_transition_dm.py
frieschneider: I get your point. I think, however, there is no need to be able to compute the "conventional" ISR(1) for an ADC(2) calculation, since there a just first-order terms missing.  🤷‍♀️

adcc/adc_pp/state_diffdm_2p.py
```diff
@@ -91,11 +91,65 @@ def diffdm_adc1_2p(mp, amplitude, intermediates):
             einsum("jk,ikab->ijab", p1_oo, t2)
         ).antisymmetrise(0, 1)
     )
+
     return dm
 
 
-def diffdm_adc2_2p(mp, amplitude, intermediates):
-    dm = diffdm_adc1_2p(mp, amplitude, intermediates)  # Get ADC(1) result
+def diffdm_isr1_2p(mp, amplitude, intermediates):
```
jonasleitner, line 98: Should this function call `isr1s` to avoid the repetition for the first-order singles-singles contribution?

adcc/AdcMatrix.py
```diff
@@ -73,17 +73,22 @@ class AdcMatrixlike:
...
     _special_block_orders = {
         "adc2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
+        "isr1s": {"ph_ph": 1, "ph_pphh": None, "pphh_ph": None, "pphh_pphh": None},
     }
 
     @classmethod
-    def _default_block_orders(cls, method: AdcMethod) -> dict[str, int]:
+    def _default_block_orders(cls,
+                              method: Method
+                              ) -> dict[str, int]:
```
jonasleitner:
```suggestion
    def _default_block_orders(cls, method: Method) -> dict[str, int]:
```

adcc/AdcMethod.py
```diff
@@ -20,76 +20,125 @@
...
-        The name of the canonical method to use for computing properties
-        for this ADC method. This only differs from the name property
-        for the ADC(2)-x family of methods.
-        """
-        if self.__base_method == "adc2x":
-            return AdcMethod(self.name.replace("adc2x", "adc2")).name
-        else:
-            return self.name
+    def _base_method(self) -> str:
+        return self._method_base_name + self.level.to_str()
```
jonasleitner, line 115:
```suggestion
    def _base_method(self) -> str:
        assert self._method_base_name is not None
        return self._method_base_name + self.level.to_str()
```

adcc/AdcMethod.py
```diff
@@ -20,76 +20,125 @@
...
+    ONE = 1
+    TWO = 2
+    THREE = 3
+    FOUR = 4
+    FIVE = 5
 
+    # special levels
+    TWO_X = "2x"
+    ONE_S = "1s"
+    THREE_D = "3d"
```
jonasleitner: I think it would be good to add a comment for each special level that explains what the level means

adcc/AdcMethod.py
```diff
@@ -20,76 +20,125 @@
...
+            self.level: MethodLevel = MethodLevel(level)
+        self._validate_level(self.level)
+
+        assert self._base_method == split[-1]
+
+        # validate prefix
         split = split[:-1]
-        self.is_core_valence_separated = "cvs" in split
+        if split and split[0] not in ["cvs"]:
+            raise ValueError(f"{split[0]} is not a valid method prefix")
```
jonasleitner: I think we also need to verify that split is at most of length 1 with cvs being the only valid entry, right?

```suggestion
        valid_prefixes: tuple[str, ...] = ("cvs",)
        if len(split) > len(valid_prefixes):
            raise ValueError(f"Invalid number of method prefixes provided in {split}.")
        if any(pref not in valid_prefixes for pref in split):
            raise ValueError(f"Invalid method prefix in {split}.")
```

adcc/AdcMethod.py
```diff
@@ -20,76 +20,125 @@
...
-            if self.__base_method == "adc2x":
-                self.level = 2
-            else:
-                self.level = int(self.__base_method[-1])
-        except ValueError:
-            raise ValueError("Not a valid base method: " + self.__base_method)
+    def _validate_level(self, level: MethodLevel) -> None:
+        if isinstance(level.value, int):
+            if level.value <= self.max_level:
+                return
```
jonasleitner:
```suggestion
        if isinstance(level.value, int) and level.value <= self.max_level:
            return
```

jonasleitner (approved):

PR #210, opened 2026-05-08
ISR(3) properties implementation

whyshna:
- State difference density matrices (`state_diffdm`) - `diffdm_isr3` added
- Ground-to-excited transition density matrices (`transition_dm`) -  `tdm_isr3` added
- State-to-state transition density matrices (`state2state_transition_dm`) -  `s2s_tdm_isr3` added
- Modified transition moments (`modified_transition_moments`) - `mtm_isr3` added
- ISR(3) blocks in `bmatrix`:
 -  third order singles block (`block_ph_ph_3`)
 -  couplings block - ph, pphh at second order (`block_ph_pphh_2`)
 - couplings block - pphh, ph at second order (`block_pphh_ph_2`)
 -  doubles block at first order (`block_pphh_pphh_1`)

- `workflow.py` - `isr_order` parameter added, ISR logic centralized here 
- `isr_order` can now be used as a keyword while running an ADC calculation, specifying the `isr_order` 
- For eg:  `state = adcc.adc3(scfres, n_singlets=3, isr_order=3)`

- `ElectronicStates`/`ExcitedStates` - accept `property_method` directly
- `State2States` - inherits `property_method` from parent state automatically
- HDF5 structure updated from `{case}/{gs_density_order}/{kind}` to `{case}/{gs_density_order}/{isr_order}/{kind}` - now the adc3 datas contains properties at both `isr_order=3`, and `isr_order=2`(`isr_order=2` is stored as `None`)
- ADC(3) still takes `isr_order=2` as default, unless explicitly mentioned

### Known Limitations
- CVS-ADC(3) with ISR(3) is not yet implemented


adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +222,794 @@ def apply(ampl):
...
+        return AmplitudeVector(
+            ph=(
+                # zeroth order
+                +1 * einsum("ib,ab->ia", ur1, d_vv)  # N^3: O^1V^2 / N^2: V^2
+                - 1 * einsum("ja,ji->ia", ur1, d_oo)  # N^3: O^2V^1 / N^2: O^1V^1
+                # second order
+                + 1
+                * einsum(
+                    "ab,ib->ia", einsum("kj,jbka->ab", d_oo, t2sq), ur1
+                )  # N^4: O^2V^2 / N^4: O^2V^2
```
jonasleitner: Can you please put the prefactor and the outer most einsum call in the same line? (For all equations)

```suggestion
                + 1 * einsum(
                    "ab,ib->ia", einsum("kj,jbka->ab", d_oo, t2sq), ur1
                )  # N^4: O^2V^2 / N^4: O^2V^2
```

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +222,794 @@ def apply(ampl):
...
+        ur1 = ampl.ph
+        return AmplitudeVector(
+            pphh=0.5
+            * (
+                # zeroth order
+                # Apply (1 - P_ij - P_ab + P_ijP_ab) to:
+                4
+                * einsum("ja,bi->ijab", ur1, d_vo)  # N^4: O^2V^2 / N^4: O^2V^2
+                .antisymmetrise(0, 1)
+                .antisymmetrise(2, 3)
```
jonasleitner: Please attach the `antisymmetrise` (or `symmetrise`) calls directly to the closing round brackets. (For all equations)

```suggestion
                # N^4: O^2V^2 / N^4: O^2V^2
                4 * einsum(
                    "ja,bi->ijab", ur1, d_vo
                ).antisymmetrise(0, 1).antisymmetrise(2, 3)
```

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -77,20 +77,130 @@ def mtm_cvs_isr2(mp, op, intermediates):
...
+def mtm_cvs_isr3(mp, op, intermediates):
+    raise NotImplementedError("CVS-ADC(3) is not implemented yet")
+
+
 DISPATCH = {
     "isr0": mtm_isr0,
     "isr1s": mtm_isr1,  # Identical to ISR(1)
     "isr1": mtm_isr1,
     "isr2": mtm_isr2,
+    "isr3": mtm_isr3,
```
jonasleitner, line 170: Please add `isr3d` to the dispatch and just forward to `mtm_isr3`

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -124,8 +234,8 @@ def modified_transition_moments(method, ground_state, operator=None,
         unpack = True
         operator = [operator]
     if method.name not in DISPATCH:
-        raise NotImplementedError("modified_transition_moments is not "
-                                  f"implemented for {method.name}.")
+        raise NotImplementedError(
+            "modified_transition_moments is not " f"implemented for {method.name}.")
```
jonasleitner:
```suggestion
        raise NotImplementedError(
            f"modified_transition_moments is not implemented for {method.name}."
        )
```

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -116,16 +116,491 @@ def s2s_tdm_isr2(mp, amplitude_l, amplitude_r, intermediates):
     return dm
 
 
+def s2s_tdm_isr3(mp, amplitude_l, amplitude_r, intermediates):
```
jonasleitner: This is the `s2s_tdm_isr3d`. For the full isr3 we are missing triples contributions, which we will add later. Therefore, just rename this function and add another function that then calls the `isr3d` function. Something like
```
def s2s_tdm_isr3(mp, amplitude_l, amplitude_r, intermediates):
    dm = s2s_tdm_isr3d(mp, amplitude_l, amplitude_r, intermediates)
    try:
        check_triples_amplitudes([b.o, b.o, b.o, b.v, b.v, b.v], amplitude_l, amplitude_r)
    except ValueError:
        return dm

    raise NotImplementedError(
        "Consistent ISR(3) including triples is not implemented yet."
    )
```

(Same for the state_diffdm)

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -116,16 +116,491 @@ def s2s_tdm_isr2(mp, amplitude_l, amplitude_r, intermediates):
...
+    # mp2 second order correction
+    p0_2_oo = mp.mp2_diffdm.oo
+    p0_2_vv = mp.mp2_diffdm.vv
+
+    # third order correction
+    p0_3_oo = mp.mp3_dm_correction.oo
+    p0_3_vv = mp.mp3_dm_correction.vv
+    p0_3_ov = mp.mp3_dm_correction.ov
+
+    dm.oo = (
```
jonasleitner: `dm` already contains all contributions through second-order. Instead of overwriting all blocks and recomputing the lower order contributions use in-place addition and only implement the new ISR(3) contributions.

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -77,20 +77,130 @@ def mtm_cvs_isr2(mp, op, intermediates):
...
+    t2sq = einsum("ikac,jkbc->iajb", mp.t2oo, mp.t2oo).evaluate()
+
+    d_vv, d_vo, d_ov, d_oo = op.vv, op.vo, op.ov, op.oo  # density operators
+
+    p0_3 = mp.mp3_dm_correction
+    p0_2 = mp.mp2_dm_correction
+    p0_3_oo, p0_3_vv, p0_3_ov = p0_3.oo, p0_3.vv, p0_3.ov
+    p0_2_oo, p0_2_vv = p0_2.oo, p0_2.vv
+
+    f1 = (
```
jonasleitner: Instead of recalculating all contributions, please call `mtm_isr2` and only compute the new ISR(3) contributions in this function like it is done in the other `mtm_isrn` functions

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -116,16 +116,491 @@ def s2s_tdm_isr2(mp, amplitude_l, amplitude_r, intermediates):
...
-            "isr1s": s2s_tdm_isr0,  # Identical to ISR(0)
-            "isr1": s2s_tdm_isr1,
-            "isr2": s2s_tdm_isr2,
-            }
+DISPATCH = {
+    "isr0": s2s_tdm_isr0,
+    "isr1s": s2s_tdm_isr0,  # Identical to ISR(0)
+    "isr1": s2s_tdm_isr1,
+    "isr2": s2s_tdm_isr2,
+    "isr3": s2s_tdm_isr3,
```
jonasleitner, line 479: Please add `isr3d` to the dispatch

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -155,11 +630,11 @@ def state2state_transition_dm(method, ground_state, amplitude_from,
         intermediates = Intermediates(ground_state)
 
     if method.name not in DISPATCH:
-        raise NotImplementedError("state2state_transition_dm is not implemented "
-                                  f"for {method.name}.")
+        raise NotImplementedError(
+            "state2state_transition_dm is not implemented " f"for {method.name}.")
```
jonasleitner:
```suggestion
        raise NotImplementedError(
            f"state2state_transition_dm is not implemented for {method.name}."
        )
```

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +222,794 @@ def apply(ampl):
             + 1.0 * einsum('kc,knce,inaf,ef->ia', ampl.ph, t2, t2, op.vv)
         ))
     return IsrBlock(apply)
+
+#
+#
+# ISR(3)
+#
+#
```
jonasleitner: Please remove the comments and integrate the functions according to the existing ordering, i.e., all zeroth order diagonal blocks next to each other and so on.

adcc/adc_pp/state_diffdm.py
```diff
@@ -159,14 +160,270 @@ def diffdm_cvs_isr2(mp, amplitude, intermediates):
...
+
+    t2_1 = mp.t2(b.oovv)  # first order doubles
+    t1_2 = mp.diffdm(level=2).ov  # second order singles
+    t2_2 = mp.td2(b.oovv)  # second order doubles
+    t3_2 = mp.tt2(b.ooovvv)  # second order triples
+
+    p0_3 = mp.mp3_dm_correction  # third order dm correction
+    p0_2 = mp.mp2_dm_correction  # second order dm correction
+
+    dm.oo = (  # adc3_p_oo
```
jonasleitner: You re-evaluate all the isr2 terms that already have been computed in `diffdm_isr2`. Please use in-place addition and only evaluate the ISR(3) contributions

adcc/adc_pp/state_diffdm.py
```diff
@@ -159,14 +160,270 @@ def diffdm_cvs_isr2(mp, amplitude, intermediates):
...
+                einsum("ic,jkac->ijka", ur1, t2_1),
+            )
+        ).symmetrise()
+    )
+
+    return dm
+
+
+def diffdm_cvs_adc3(mp, amplitude, intermediates):
+    raise NotImplementedError("CVS-ADC(3) is not implemented yet")
```
jonasleitner: There is no need for this function, since the `state_diffdm`function below already raises an exception since `cvs-isr3` is not contained in the `DISPATCH` dictionary. Please remove it.

adcc/adc_pp/state_diffdm.py
```diff
@@ -159,14 +160,270 @@ def diffdm_cvs_isr2(mp, amplitude, intermediates):
...
+
+
 # dict controlling the dispatch of the state_diffdm function
 DISPATCH = {
     "isr0": diffdm_isr0,
-    "isr1s": diffdm_isr0,   # Identical to ISR(0)
+    "isr1s": diffdm_isr0,  # Identical to ISR(0)
     "isr1": diffdm_isr1,
     "isr2": diffdm_isr2,
+    "isr3": diffdm_isr3,
```
jonasleitner, line 352: Please add `isr3d` dispatch

adcc/adc_pp/state_diffdm.py
```diff
@@ -198,8 +455,9 @@ def state_diffdm(method, ground_state, amplitude, intermediates=None):
         intermediates = Intermediates(ground_state)
 
     if method.name not in DISPATCH:
-        raise NotImplementedError("state_diffdm is not implemented "
-                                  f"for {method.name}.")
+        raise NotImplementedError(
+            "state_diffdm is not implemented " f"for {method.name}."
+        )
```
jonasleitner, line 389:
```suggestion
        raise NotImplementedError(
            f"state_diffdm is not implemented for {method.name}."
        )
```

adcc/adc_pp/transition_dm.py
```diff
@@ -100,11 +100,90 @@ def tdm_isr2(mp, amplitude, intermediates):
...
+    t2_3 = mp.td3(b.oovv)
+
+    p0_3_ov = mp.mp3_dm_correction.ov
+    p0_3_oo = mp.mp3_dm_correction.oo
+    p0_3_vv = mp.mp3_dm_correction.vv
+
+    p0_2_oo = mp.mp2_dm_correction.oo
+    p0_2_vv = mp.mp2_dm_correction.vv
+
+    dm.oo = (
```
jonasleitner: `dm` already contains all ISR(2) contributions. Please, only evaluate and implement the new ISR(3) contributions

adcc/adc_pp/transition_dm.py
```diff
@@ -100,11 +100,90 @@ def tdm_isr2(mp, amplitude, intermediates):
...
+
+    return dm
+
+
 DISPATCH = {
     "isr0": tdm_isr0,
     "isr1s": tdm_isr1,  # Identical to ISR(1)
     "isr1": tdm_isr1,
     "isr2": tdm_isr2,
+    "isr3": tdm_isr3,
```
jonasleitner, line 178: Please add `isr3d` to the dispatch

adcc/adc_pp/transition_dm.py
```diff
@@ -137,8 +216,8 @@ def transition_dm(method, ground_state, amplitude, intermediates=None):
         intermediates = Intermediates(ground_state)
 
     if method.name not in DISPATCH:
-        raise NotImplementedError("transition_dm is not implemented "
-                                  f"for {method.name}.")
+        raise NotImplementedError(
+            "transition_dm is not implemented " f"for {method.name}.")
```
jonasleitner:
```suggestion
        raise NotImplementedError(
            f"transition_dm is not implemented for {method.name}."
        )
```

adcc/tests/adc_pp/modified_transition_moments_test.py
```diff
@@ -41,20 +49,26 @@
...
-@pytest.mark.parametrize("method", methods)
+@pytest.mark.parametrize("adc_method, isr_order", methods)
 @pytest.mark.parametrize("system,case,kind", cases)
 @pytest.mark.parametrize("op_kind", operator_kinds)
-def test_modified_transition_moments(system: str, case: str, method: str, kind: str,
-                                     op_kind: str):
+def test_modified_transition_moments(system: str, case: str, adc_method: str,
+                                     isr_order, kind: str, op_kind: str):
+
+    isr_key = "3" if isr_order == 3 else "None"
```
jonasleitner: You dont need to define an `isr_key`. You can just use `str(isr_order)` as key below

adcc/tests/adc_pp/modified_transition_moments_test.py
```diff
@@ -66,10 +80,16 @@ def test_modified_transition_moments(system: str, case: str, method: str, kind:
...
-    mtms = modified_transition_moments(method, state.ground_state, dips)
+    if "cvs" in case and "cvs" not in adc_method:
+        adc_method = f"cvs-{adc_method}"
+    if isr_order is not None:
+        mtm_level = f"isr{isr_order}"
+    else:
+        level = min(AdcMethod(adc_method).level.to_int(), 2)
+        mtm_level = f"isr{level}"
+    if "cvs" in case:
+        mtm_level = f"cvs-{mtm_level}"
```
jonasleitner: Can't we just use the `property_method` from `state`? Then we could simply remove this code block

adcc/tests/adc_pp/state_diffdm_test.py
```diff
@@ -52,14 +48,7 @@ def calculate_adcn_excitation_energy(self, state):
...
         if state.method.name == "adc3":
-            # so we don't forget to switch to the actual implementation
-            with pytest.raises(NotImplementedError):
-                method = IsrMethod("isr3")
-
-            # TODO switch to ISR(3) implementation
-            method = IsrMethod("isr2")
-            method.level = MethodLevel(3)
-
+            method = IsrMethod("isr3")
```
jonasleitner: I think this if statement might not be needed anymore. If `isr_order` is set on the state, we can always use `state.property_method`. Otherwise `state.method.as_method(IsrMethod)` should work if I am not mistaken.

adcc/tests/generators/dump_adcc.py
```diff
@@ -101,10 +117,13 @@ def dump_excited_states(states: ExcitedStates, hdf5_file: h5py.Group,
             eigenvectors[exdegree + 1].append(getattr(
                 states.excitation_vector[n], block  # type: ignore
             ).to_ndarray())
-    kind_data = {}
-    # eigenvalues
+
+        kind_data = {}
```
jonasleitner: This is indented one too much and should not be part of the loop, right?

adcc/tests/generators/generate_adcman_data.py
```diff
@@ -45,29 +46,36 @@ def generate_adc(test_case: testcases.TestCase, method: AdcMethod, case: str,
...
-    if f"{case}/{gs_density_order}" in hdf5_file:
+
+    key = f"{case}/{gs_density_order}/{isr_order}"
+    if key in hdf5_file:
         return None
     # skip cvs-adc(0), since it is not available in qchem.
-    if "cvs" in case and method.level == 0:
+    if "cvs" in case and method.level is MethodLevel.ZERO:
+        return None
+    # skip cvs-adc(3), since it is not implemented yet.
```
jonasleitner:
```suggestion
    # skip cvs-isr(3), since it is not implemented yet.
```

adcc/tests/ExcitedStates_test.py
```diff
@@ -33,7 +33,7 @@
 class TestExcitedStates:
     def test_excitation_view(self):
         state = testdata_cache.adcc_states(
-            system="h2o_sto3g", method="adc2", case="gen", kind="singlet"
+            system="h2o_sto3g", method="adc3", case="gen", kind="singlet"
```
jonasleitner: Why did you change the method from `adc2` to `adc3` in this file? This test should not be related to any of your changes

adcc/tests/functionality_test.py
```diff
@@ -59,7 +59,9 @@ def base_test(self, system: testcases.TestCase, case: str, method: str,
         refdata = getattr(testdata_cache, f"{generator}_data")(
             system=system, method=method, case=case
         )
-        ref: dict = refdata[kind]
+
+        hdf5_key = "None"
+        ref: dict = refdata[hdf5_key][kind]
```
jonasleitner:
```suggestion
        ref: dict = refdata["None"][kind]
```

adcc/tests/LazyMp_test.py
```diff
@@ -307,9 +342,6 @@ class TestGroundstateDensity:
     def calculate_mpn_energy(self, mp, level):
         hf = mp.reference_state
         if level == 3:
-            # TODO move to ISR(3) implementation
-            with pytest.raises(NotImplementedError):
-                mp.density(level)
```
jonasleitner: I think the if statement should not be needed anymore, since `mp.density(level)` should now also work for `level=3`. Also please delete the `mp3_diffdm` method on the `TestGroundstateDensity` class, since it is now obsolete.

adcc/tests/properties_test.py
```diff
@@ -39,9 +39,18 @@
...
-methods = ["adc2"]
+
+methods = [
+    ("adc0", None),
+    ("adc1", None),
+    ("adc2", None),
+    ("adc2x", None),
+    ("adc3", None),
+    ("adc3", 3),
+]
```
jonasleitner: As it says in the comment above, the tests in this file should be independent of the ADC method and therefore also independent of the isr_order. Thus I think it is sufficient to only test ADC(2/2)

```suggestion

methods = [
    ("adc2", None),
]
```

adcc/tests/properties_test.py
```diff
@@ -56,21 +65,26 @@
...
+@pytest.mark.parametrize("adc_method, isr_order", methods)
 class TestProperties:
     @pytest.mark.parametrize("generator", generators)
     @pytest.mark.parametrize("system,case,kind", cases)
     def test_transition_dipole_moments(self, system: str, case: str, kind: str,
-                                       method: str, generator: str):
-        if "cvs" in case and AdcMethod(method).level == 0 and generator == "adcman":
+                                       adc_method: str, isr_order, generator: str):
+
+        isr_key = "3" if isr_order == 3 else "None"
```
jonasleitner: This is not necessary, since we can use `str(isr_order)` to access the data below.
(Same for the other tests in this file)

adcc/tests/state_densities_test.py
```diff
@@ -48,23 +55,31 @@
...
 class TestStateDensities:
     @pytest.mark.parametrize("system,case,kind", cases)
-    def test_state_diffdm(self, system: str, case: str, kind: str, method: str,
-                          generator: str):
-        if ("cvs" in case and AdcMethod(method).level.to_int() == 0
-                and generator == "adcman"):
+    def test_state_diffdm(self, system: str, case: str, kind: str, adc_method: str,
+                          isr_order, generator: str):
+
+        isr_key = "3" if isr_order == 3 else "None"
```
jonasleitner: Please use `str(isr_order)` directly to extract the data below
(Same for the other tests in this file)

adcc/tests/state_densities_test.py
```diff
@@ -48,23 +55,31 @@
...
+        if (
+            "cvs" in case and AdcMethod(adc_method).level.to_int() == 0
+            and generator == "adcman"
+        ):
             pytest.skip("No CVS-ADC(0) adcman reference data available.")
+        if 'cvs' in case and adc_method == "adc3":
+            pytest.skip("CVS-ADC3 not implemented yet")
```
jonasleitner: CVS-ADC(3) should be available. You probably mean ISR(3)?

```suggestion
        if 'cvs' in case and isr_order == 3:
            pytest.skip("CVS-ISR(3) not implemented yet")
```
(same for the other tests in this file)

adcc/tests/testdata_cache.py
```diff
@@ -213,11 +217,16 @@ def _make_mock_adc_state(self, system: Union[str, testcases.TestCase],
             system, method=method, case=case, source=source,
             gs_density_order=gs_density_order
         )
-        adc_data = data.get(kind, None)
-        if adc_data is None:
-            raise ValueError(f"No data available for kind {kind} in {case} "
-                             f"{method} {system}.")
-        # load the reference state and build a matrix on top
+        hdf5_key = "None" if isr_order != 3 else "3"
+        isr_data = data.get(hdf5_key, None)
```
jonasleitner:
```suggestion
        isr_data = data.get(str(isr_order), None)
```

adcc/tests/testdata_cache.py
```diff
@@ -257,10 +270,11 @@ def _make_mock_adc_state(self, system: Union[str, testcases.TestCase],
                 evec[blocks[1]].set_from_ndarray(
                     adc_data["eigenvectors_doubles"][i], 1e-14
                 )
-        return ExcitedStates(states)
+        return ExcitedStates(states, property_method=states._property_method)
```
jonasleitner: There should be no need to define `_property_method` on `states`

adcc/AdcMatrix.py
```diff
@@ -74,6 +74,7 @@ class AdcMatrixlike:
     _special_block_orders = {
         "adc2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
         "isr1s": {"ph_ph": 1, "ph_pphh": None, "pphh_ph": None, "pphh_pphh": None},
+        "isr3": {"ph_ph": 3, "ph_pphh": 2, "pphh_ph": 2, "pphh_pphh": 1},
```
jonasleitner: This should be `isr3d`

adcc/AdcMethod.py
```diff
@@ -162,5 +162,6 @@ class AdcMethod(Method):
 
 class IsrMethod(Method):
     _method_base_name = "isr"
-    max_level = 2
+    max_level = 3
     special_levels = (MethodLevel.ONE_S,)
```
jonasleitner: please add `THREE_D` as special level, but keep `max_level = 3`

adcc/ElectronicStates.py
```diff
@@ -84,13 +85,14 @@ def __init__(self, data, method: str = None,
 
         if property_method is None:
             if self.method.level in [MethodLevel.TWO_X, MethodLevel.THREE]:
-                # Auto-select ISR(2) properties for ADC(2)-x and ADC(3) calc
-                warnings.warn(f"ISR({self.method.level.to_str()}) not implemented."
-                              f" Property method is selected as ISR(2).")
+                warnings.warn(
+                    f"property_method defaults to ISR(2) for {self.method.name}."
+                )
```
jonasleitner:
```suggestion
                warnings.warn(
                    "By default ISR(2) is selected as property method "
                    f"for an {self.method.name} calculation."
                )
```

adcc/ElectronicStates.py
```diff
@@ -84,13 +85,14 @@ def __init__(self, data, method: str = None,
...
                 property_method = self.method.at_level(2).as_method(IsrMethod)
             else:
                 property_method = self.method.as_method(IsrMethod)
-        elif not isinstance(property_method, IsrMethod):
+
+        if not isinstance(property_method, IsrMethod):
```
jonasleitner:
```suggestion
        elif not isinstance(property_method, IsrMethod):
```

adcc/LazyMp.py
```diff
@@ -148,6 +197,64 @@ def t2eri(self, space, contraction):
...
+        if level in [0, 1]:
+            raise ValueError()
+        elif level == 2:
+            return self.mp2_diffdm
+        elif level == 3:
+            return self.mp3_diffdm
+        else:
+            raise NotImplementedError(
+                "diffdm for order above 3 not implemented yet."
+            )
```
jonasleitner: I am not sure if we want an `mp3_diffdm` method. I think we should instead accumulate the densities within this function.

```suggestion
        if level == 2:
            return self.mp2_diffdm
        elif level == 3:
            return self.mp2_dm_correction + self.mp3_dm_correction
        else:
            raise NotImplementedError(
                "Only 2nd and 3rd-order difference densities are implemented."
            )
```

adcc/LazyMp.py
```diff
@@ -148,6 +197,64 @@ def t2eri(self, space, contraction):
...
+
     @cached_property
     @timed_member_call(timer="timer")
     def mp2_diffdm(self):
```
jonasleitner: Please remove the decorators from this method, move the implementation to `mp2_dm_correction` and call `mp2_dm_correction` from this method.

```suggestion
    def mp2_diffdm(self):
```

adcc/LazyMp.py
```diff
@@ -208,11 +358,45 @@ def density(self, level=2):
...
+                + self.mp2_dm_correction
+                + self.mp3_dm_correction
+            )
         else:
             raise NotImplementedError("Only densities for level 0, 1 and 2"
                                       " are implemented.")
 
+    @cached_property
+    @timed_member_call(timer="timer")
+    def mp3_diffdm(self) -> OneParticleDensity:
```
jonasleitner: I don't think we need this method. We should instead use the `diffdm` method and only keep the `mp2_diffdm` for consistency.

adcc/LazyMp.py
```diff
@@ -292,6 +476,8 @@ def dipole_moment(self, level=2):
             return self.reference_state.dipole_moment
         elif level == 2:
             return self.mp2_dipole_moment
+        elif level == 3:
+            return self.mp3_dipole_moment
         else:
             raise NotImplementedError("Only dipole moments for level 0, 1 and 2"
                                       " are implemented.")
```
jonasleitner: I don't think we need an `mp3_dipole_moment` method either.
We should be able to generalize this method. We can keep `mp2_dipole_moment` for consistency, but just call `dipole_moment` from there. Caching should then also happen on this method and not on `mp2_dipole_moment`

```
@cached_member_function()
def dipole_moment(self, level=2):
"""
Return the MP dipole moment at the specified level of
perturbation theory.
"""
dipole_integrals = refstate.operators.electric_dipole
density = self.density(level)
return np.array(
    product_trace(comp, density) for comp in dipole_integrals
)
```

adcc/LazyMp.py
```diff
@@ -392,10 +578,19 @@ def ssq(self, level=2):
         ssq_2p = product_trace(ssq_2p_op, self.density_2p(level))
         return (ssq_1p + ssq_2p)
 
+    @cached_property
+    def mp3_dipole_moment(self):
```
jonasleitner: I think we can remove this method, but keep `mp2_dipole_moment` for consistency

adcc/State2States.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, data, method=None, property_method=None, initial=0):
                 Provide the index of the excited state from which transitions
                 to all other higher-lying states are to be computed.
         """
+
+        if property_method is None and hasattr(data, '_property_method'):
+            property_method = data._property_method
```
jonasleitner: I think this should rather be handled in `__init__` of `ElectronicStates` to also automatically use the property method when constructing an `ElectronicStates` instance from another one.

adcc/workflow.py
```diff
@@ -108,6 +108,9 @@ def run_adc(data_or_matrix, n_states=None, kind="any", conv_tol=None,
         preference over `n_guesses` and `n_guesses_doubles`, such that these
         parameters are ignored.
 
+    isr_order: int, optional
+        Order of the ISR expansion used for property calculations.
```
jonasleitner: isr_order can also be a string, e.g. `isr1s`.

examples/water/pyscf_sto3g_adc2.py
```diff
@@ -24,3 +24,6 @@
 print(singlets.describe())
 print()
 print(triplets.describe())
+
+
+
```
jonasleitner: please remove the empty lines

adcc/workflow.py
```diff
@@ -183,6 +186,12 @@ def run_adc(data_or_matrix, n_states=None, kind="any", conv_tol=None,
         data_or_matrix, core_orbitals=core_orbitals, frozen_core=frozen_core,
         frozen_virtual=frozen_virtual, method=method)
 
+    if isr_order is not None:
+        if not isinstance(isr_order, int):
+            raise TypeError("isr_order must be an integer")
+        if isr_order < 0:
+            raise ValueError("isr_order must be >=0")
+
```
jonasleitner: When we construct the property_method this should be verified automatically.

```suggestion
```

adcc/workflow.py
```diff
@@ -209,7 +218,14 @@ def run_adc(data_or_matrix, n_states=None, kind="any", conv_tol=None,
         matrix, n_states, kind, guesses=guesses, n_guesses=n_guesses,
         n_guesses_doubles=n_guesses_doubles, conv_tol=conv_tol, output=output,
         eigensolver=eigensolver, **solverargs)
-    exstates = ExcitedStates(diagres)
+    if isr_order is not None:
+        property_method = matrix.method.at_level(isr_order).as_method(IsrMethod)
+    elif matrix.method.level in [MethodLevel.TWO_X, MethodLevel.THREE]:
+        property_method = matrix.method.at_level(2).as_method(IsrMethod)
+    else:
+        property_method = matrix.method.as_method(IsrMethod)
```
jonasleitner: Please shift this above the `diagonalise_adcmatrix` call. Also you only need to handle the case `isr_order is not None`. Everything else should already be handled in `ExcitedStates`

adcc/tests/ExcitedStates_test.py
whyshna: that was accidental mistake

adcc/LazyMp.py
whyshna: yeah but if im not wrong `mp3_diffdm` is called in a lot of other places, but i could also do the changes in those places, let me know if the latter one is a better idea :) (its an easy fix)

adcc/LazyMp.py
whyshna: with the above code you mentioned, the `nuclear_dipole` is not included, right?

adcc/adc_pp/bmatrix.py
```diff
@@ -182,6 +182,12 @@ def apply(ampl):
     return IsrBlock(apply)
 
 
+#
+# 1st order doubles block
+#
+block_pphh_pphh_1 = block_pphh_pphh_0
```
frieschneider: I would place that under 1st order main. E.g. `block_pphh_pphh_0` is also placed under 0th order main

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +228,584 @@ def apply(ampl):
             + 1.0 * einsum('kc,knce,inaf,ef->ia', ampl.ph, t2, t2, op.vv)
         ))
     return IsrBlock(apply)
+
+
+#
+# Second order coupling
```
frieschneider:
```suggestion
# 2nd order coupling
```

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +228,584 @@ def apply(ampl):
...
+
+    p0_2 = ground_state.mp2_dm_correction
+    p0_2_vv, p0_2_oo = p0_2.vv, p0_2.oo
+
+    def apply(ampl):
+        ur1 = ampl.ph
+        return AmplitudeVector(
+            pphh=0.5 * (
+                # zeroth order
+                # Apply (1 - P_ij - P_ab + P_ijP_ab) to:
```
frieschneider: I would remove these symmetrisation comments. What do you think @jonasleitner ?

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +228,584 @@ def apply(ampl):
...
+                    )  # N^6: O^4V^2 / N^4: O^2V^2
+                ).antisymmetrise(0, 1)
+            )
+        )
+
+    return IsrBlock(apply)
+
+
+#
+# Third order main
```
frieschneider:
```suggestion
# 3rd order main
```

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -77,20 +77,116 @@ def mtm_cvs_isr2(mp, op, intermediates):
...
     "cvs-isr1": mtm_cvs_isr0,  # Identical to CVS-ISR(0)
     "cvs-isr2": mtm_cvs_isr2,
 }
 
 
-def modified_transition_moments(method, ground_state, operator=None,
-                                intermediates=None):
+def modified_transition_moments(
+    method, ground_state, operator=None, intermediates=None
+):
```
frieschneider, line 182: Is there a reason why you changed the alignment? :)

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -116,16 +116,439 @@ def s2s_tdm_isr2(mp, amplitude_l, amplitude_r, intermediates):
...
+            t2_1,
+        )
+    )
+    return dm
+
+
+def s2s_tdm_isr3(mp, amplitude_l, amplitude_r, intermediates):
+    dm = s2s_tdm_isr3d(mp, amplitude_l, amplitude_r, intermediates)
+    try:
+        check_triples_amplitudes([b.o, b.o, b.o, b.v, b.v, b.v],
```
frieschneider, line 462: you need to add `check_triples_amplitudes` to `adc_pp/util.py` :) Or wait until #211 is merged, there it is added.

adcc/tests/generators/dump_adcc.py
```diff
@@ -60,6 +61,21 @@ def dump_groundstate(ground_state: LazyMp, hdf5_file: h5py.Group,
...
     gs_data[f"{gs}2/dm_bb_b"] = dm_bb_b.to_ndarray()
+
+    if not ground_state.has_core_occupied_space:
+        # MP3 density: MO basis
+        dm_blocks = ["dm_o1o1", "dm_o1v1", "dm_v1v1"]
+
+        for block in dm_blocks:
+            blk = block.split("_")[-1]
+            gs_data[f"{gs}3/{block}"] = ground_state.mp3_diffdm[blk].to_ndarray()
+    # MP3 density: AO basis
```
frieschneider: Please make sure that the comment is indented correctly.

adcc/LazyMp.py
```diff
@@ -208,11 +356,45 @@ def density(self, level=2):
...
+        ret.ov = (
+            mp2_dm.ov  # 2nd order
+            + mp3_dm.ov  # 3rd order
+        )
+        ret.vv = (
+            mp2_dm.vv  # 2nd order
+            # 3rd order
+            + mp3_dm.vv
+        )
+        return evaluate(ret)
```
frieschneider:
```suggestion
        return evaluate(mp2_dm + mp3_dm)
```

Addition of two one particle operators is implemented. You don't need to do it explicitly like that. :)

adcc/LazyMp.py
```diff
@@ -392,10 +572,11 @@ def ssq(self, level=2):
         ssq_2p = product_trace(ssq_2p_op, self.density_2p(level))
         return (ssq_1p + ssq_2p)
 
-
 #
 # Register cvs_p0 intermediate
 #
+
+
```
frieschneider, line 310:
```suggestion
```

adcc/State2States.py
```diff
@@ -58,6 +58,7 @@ def __init__(self, data, method=None, property_method=None, initial=0):
                 Provide the index of the excited state from which transitions
                 to all other higher-lying states are to be computed.
         """
+
```
frieschneider, line 62:
```suggestion
```

adcc/tests/LazyMp_test.py
frieschneider, line 222: I think we should add a `test_td3` as well and what about tests for the `sigma_inf_ov`? We should tests the consistency of these, too.

adcc/adc_pp/bmatrix.py
jonasleitner: Yes, I agree. It is better to remove them, since they may even be misleading in some cases once triples symmetry is taken into account.

adcc/adc_pp/modified_transition_moments.py
whyshna: nope, i used a tool to fix all the flake8 errors, and it changed the alignment of these too. I can revert back to the previous alignment :)

adcc/tests/LazyMp_test.py
whyshna: yeah, then i also need to dump them and regenerate the data right? :)

adcc/adc_pp/bmatrix.py
```diff
@@ -182,11 +185,17 @@ def apply(ampl):
     return IsrBlock(apply)
 
 
+#
+# 1st order doubles block
+#
+block_pphh_pphh_1 = block_pphh_pphh_0
```
jonasleitner: already defined above :)

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +231,576 @@ def apply(ampl):
...
+            )
+        )
+
+    return IsrBlock(apply)
+
+
+def block_pphh_ph_2(ground_state, op):
+    d_vo, d_ov, d_oo, d_vv = op.vo, op.ov, op.oo, op.vv
+
+    t1_2 = ground_state.diffdm(level=2).ov
```
jonasleitner:
```suggestion
    t1_2 = ground_state.ts2(b.ov)
```

`ts2` amplitudes are now implemented on the ground state

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +231,576 @@ def apply(ampl):
...
+
+
+#
+# 2nd order coupling
+#
+def block_ph_pphh_2(ground_state, op):
+
+    d_vo, d_ov, d_oo, d_vv = op.vo, op.ov, op.oo, op.vv
+
+    t1_2 = ground_state.diffdm(level=2).ov
```
jonasleitner:
```suggestion
    t1_2 = ground_state.ts2(b.ov)
```

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +231,576 @@ def apply(ampl):
...
+
+
+#
+# 3rd order main
+#
+def block_ph_ph_3(ground_state, op):
+
+    d_vo, d_ov, d_oo, d_vv = op.vo, op.ov, op.oo, op.vv
+
+    t1_2 = ground_state.diffdm(level=2).ov
```
jonasleitner:
```suggestion
    t1_2 = ground_state.ts2(b.ov)
```

adcc/adc_pp/bmatrix.py
```diff
@@ -222,3 +231,576 @@ def apply(ampl):
...
+    t2_1 = ground_state.t2(b.oovv)
+    t2_2 = ground_state.td2(b.oovv)
+    t2sq = einsum(
+        "ikac,jkbc->iajb",
+        ground_state.t2oo,
+        ground_state.t2oo).evaluate()
```
jonasleitner: `t2sq` is now defined as intermediate. Maybe we can incorporating an `Intermediate` instance into `IsrMatrix`? Alternatively, we could also move `t2sq` to the `GroundState` class (in analogy to the t2eri intermediates)

adcc/adc_pp/matrix.py
```diff
@@ -965,7 +965,7 @@ def adc3_i1(hf, mp, intermediates):
 def adc3_i2(hf, mp, intermediates):
     # Used only for general
     td2 = mp.td2(b.oovv)
-    p0 = mp.mp2_diffdm
+    p0 = mp.second_order_dm_correction()
```
jonasleitner, line 968: The correct replacement for `mp2_diffdm` should be `diffdm(level=2)`


```suggestion
    p0 = mp.diffdm(level=2)
```

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -61,29 +61,119 @@ def mtm_isr2(mp, op, intermediates):
...
     )
-    f2 = (1 / sqrt(2)) * einsum("kI,kjab->jIab", op.oc, mp.t2(b.oovv))
+    f2 = (1 / sqrt(2)) * einsum("kI,kjab->jIab", op.oc, ground_state.t2(b.oovv))
     return ampl + AmplitudeVector(ph=f1, pphh=f2)
 
 
+def mtm_isr3(ground_state, op, intermediates):
+
+    f1 = mtm_isr2(ground_state, op, intermediates).ph
+    f2 = mtm_isr2(ground_state, op, intermediates).pphh
```
jonasleitner: This calculates the ISR(2) contributions twice.

```suggestion
    res = mtm_isr2(ground_state, op, intermediates)
```
Below you can then just add the ISR(3) contributions to `res` directly instead of constructing a new `AmplitudeVector`

adcc/adc_pp/modified_transition_moments.py
```diff
@@ -61,29 +61,119 @@ def mtm_isr2(mp, op, intermediates):
...
+    t1_2 = ground_state.diffdm(level=2).ov
+    t2_1 = ground_state.t2(b.oovv)
+    t2_3 = ground_state.td3(b.oovv)
+    t2_2 = ground_state.td2(b.oovv)
+    t3_2 = ground_state.tt2(b.ooovvv)
+    t2sq = einsum(
+        "ikac,jkbc->iajb",
+        ground_state.t2oo,
+        ground_state.t2oo,
+    ).evaluate()
```
jonasleitner:
```suggestion
    t2sq = intermediates.t2sq
```

adcc/adc_pp/state2state_transition_dm.py
```diff
@@ -116,18 +120,369 @@ def s2s_tdm_isr2(mp, amplitude_l, amplitude_r, intermediates):
...
+def s2s_tdm_isr3d(ground_state, amplitude_l, amplitude_r, intermediates):
+    check_doubles_amplitudes([b.o, b.o, b.v, b.v], amplitude_l, amplitude_r)
+    dm = s2s_tdm_isr2(ground_state, amplitude_l, amplitude_r, intermediates)
+
+    # ADC amplitudes
+    ul1, ul2 = amplitude_l.ph, amplitude_l.pphh
+    ur1, ur2 = amplitude_r.ph, amplitude_r.pphh
+
+    t2_1 = ground_state.t2(b.oovv)
+    t1_2 = ground_state.diffdm(level=2).ov
```
jonasleitner:
```suggestion
    t1_2 = ground_state.ts2(b.ov)
```

adcc/adc_pp/state_diffdm.py
```diff
@@ -159,15 +164,194 @@ def diffdm_cvs_isr2(mp, amplitude, intermediates):
...
 
 
+def diffdm_isr3d(ground_state, amplitude, intermediates):
+    dm = diffdm_isr2(ground_state,
+                     amplitude, intermediates)  # starts from ADC2 values
+    check_doubles_amplitudes([b.o, b.o, b.v, b.v], amplitude)
+    ur1, ur2 = amplitude.ph, amplitude.pphh  # ADC amplitudes
+
+    t2_1 = ground_state.t2(b.oovv)  # first order doubles
+    t1_2 = ground_state.diffdm(level=2).ov  # second order singles
```
jonasleitner:
```suggestion
    t1_2 = ground_state.ts2(b.ov)  # second order singles
```

adcc/adc_pp/transition_dm.py
```diff
@@ -100,12 +104,92 @@ def tdm_isr2(mp, amplitude, intermediates):
...
+def tdm_isr3(ground_state, amplitude, intermediates):
+    dm = tdm_isr3d(ground_state, amplitude, intermediates)
+    try:
+        check_triples_amplitudes([b.o, b.o, b.o, b.v, b.v, b.v], amplitude)
+    except ValueError:
+        return dm
+
+    raise NotImplementedError(
+        "Consistent ISR(3) including triples is not implemented yet."
+    )
```
jonasleitner: There are no ISR(3) triples contributions for the ground-to-excited transition density matrix. Please remove this function and rename the `isr3d` function to `isr3` and call the function for both `isr3` and `isr3d` in the `DISPATCH` below

adcc/tests/adc_pp/state_diffdm_test.py
```diff
@@ -40,28 +36,24 @@
 test_cases = testcases.get_by_filename("h2o_sto3g", "cn_sto3g")
 cases = [(case.file_name, c, kind)
          for case in test_cases for c in ["gen"] for kind in case.kinds.pp]
-methods = ["adc0", "adc1", "adc2", "adc3"]
+methods = [
+    ("adc0", None),
+    ("adc1", None),
+    ("adc2", None),
+    ("adc3", 3),
+]
```
jonasleitner: This test should be independent of the `isr_order`. There should be no need to add `isr_order` to the parametrization. You can just load the data for `isr_order=None` in `test_adcn`

adcc/tests/generators/dump_adcc.py
```diff
@@ -49,6 +49,13 @@ def dump_groundstate(ground_state: LazyMp, hdf5_file: h5py.Group,
     # MP3 data
     if not ground_state.has_core_occupied_space:
         gs_data[f"{gs}3/energy"] = ground_state.energy_correction(3)
+        gs_data[f"{gs}3/dipole"] = ground_state.dipole_moment(3)
+        gs_data[f"{gs}3/sigma_inf_ov"] = ground_state.sigma_inf_ov(3).to_ndarray()
+        gs_data[f"{gs}3/m_3_plus"] = ground_state.m_3_plus.to_ndarray()
+        gs_data[f"{gs}3/m_3_minus"] = ground_state.m_3_minus.to_ndarray()
```
jonasleitner, line 55: Probably we want to raise an exception if `GroundState.has_core_occupied_space` is set for `m_3_plus` `m_3_minus` and `sigma_inf_ov`?

adcc/tests/generators/dump_adcc.py
```diff
@@ -69,6 +76,19 @@ def dump_groundstate(ground_state: LazyMp, hdf5_file: h5py.Group,
...
+
+        for block in dm_blocks:
+            blk = block.split("_")[-1]
+            gs_data[f"{gs}3/{block}"] = ground_state.diffdm(3)[blk].to_ndarray()
+        # MP3 density: AO basis
+        dm_bb_a, dm_bb_b = ground_state.diffdm(3).to_ao_basis(
+            ground_state.reference_state
+        )
+        gs_data[f"{gs}3/dm_bb_a"] = dm_bb_a.to_ndarray()
+        gs_data[f"{gs}3/dm_bb_b"] = dm_bb_b.to_ndarray()
```
jonasleitner, line 88:
```suggestion
    if not ground_state.has_core_occupied_space:
        diffdm = ground_state.diffdm(3, apply_cvs=False)
        for block in diffdm.blocks_nonzero:
            gs_data[f"{gs}3/dm_{block}"] = diffdm[block].to_ndarray()
        # MP3 density: AO basis
        dm_bb_a, dm_bb_b = diffdm.to_ao_basis(
            ground_state.reference_state
        )
        gs_data[f"{gs}3/dm_bb_a"] = dm_bb_a.to_ndarray()
        gs_data[f"{gs}3/dm_bb_b"] = dm_bb_b.to_ndarray()
```

adcc/tests/generators/dump_adcc.py
```diff
@@ -171,12 +198,16 @@ def dump_excited_states(states: ExcitedStates, hdf5_file: h5py.Group,
...
     # ssq for unrestriced calculation
     if (
         not states.reference_state.restricted
         and not states.ground_state.has_core_occupied_space
     ):
-        kind_data["state_ssq"] = states.state_ssq
+        try:
+            kind_data["state_ssq"] = states.state_ssq
+        except NotImplementedError:
+            pass
```
jonasleitner: Why is the `try` `except` needed? With the if statement above there should be no problem

adcc/tests/generators/generate_adcc_data.py
```diff
@@ -29,11 +29,14 @@ def generate_adc(test_case: testcases.TestCase, method: AdcMethod, case: str,
...
                  n_triplets: int | None = None, n_spin_flip: int | None = None,
-                 dump_nstates: int | None = None, **kwargs) -> None:
+                 dump_nstates: int | None = None,  isr_order: int | None = None,
+                 **kwargs) -> None:
     """
     Generate and dump the excited states reference data for the given reference case
     of the given test case if the data is not available already.
     """
+    if isr_order == 3 and "cvs" in case:
+        return
```
jonasleitner: Can you please move this check to the other ones below? Should be handled right before/after `if "cvs" in case and method.name == "adc4"` etc

adcc/tests/generators/generate_adcc_data.py
```diff
@@ -47,7 +50,7 @@ def generate_adc(test_case: testcases.TestCase, method: AdcMethod, case: str,
         n_spin_flip=n_spin_flip
     )
     key = f"{case}/{gs_density_order}"
-    if f"{key}/{kind}" in hdf5_file:
+    if f"{key}/{isr_order}/{kind}" in hdf5_file:
```
jonasleitner, line 59: Please add `isr_order` directly to `key`

adcc/tests/generators/generate_adcc_data.py
```diff
@@ -145,8 +150,15 @@ def generate_h2o_sto3g():
...
-            )
+                **n_states, **kwargs)
+            if method.level in [MethodLevel.THREE]:
+                generate_adc_all(test_case, method=method, dump_nstates=2,
+                                 states_per_case=per_case, isr_order=3,
+                                 **n_states, **kwargs)
+            elif method.level in [MethodLevel.FOUR]:
+                generate_adc_all(test_case, method=method, dump_nstates=2,
+                                 states_per_case=per_case, isr_order="3d",
+                                 **n_states, **kwargs)
```
jonasleitner: This is mostly the same code. I think we rather want to create some dictionary (maybe globally?), where we store the `isr_orders` for each `MethodLevel`. Something like?
```
_isr_orders = {
    MethodLevel.THREE: (None, 3),
    MethodLevel.FOUR: (None, "3d")
}
```
Same for the adcman generator

adcc/tests/generators/import_qchem_data.py
```diff
@@ -432,6 +432,8 @@ def import_data(context: h5py.File, dims_pref: str = "dims/",
 _mp_data_large = {
     # MP2 triples amplitudes
     "mp2/tt2_o1o1o1v1v1v1": "mp2/tt_o1o1o1v1v1v1",
+    # MP3 doubles amplitude at third order
+    "mp3/td3_o1o1v1v1": "mp3/td3_o1o1v1v1",
```
jonasleitner: In `adcc` we don't add the order to the key, since it is already included in the `mp3` component
```suggestion
    "mp3/td3_o1o1v1v1": "mp3/td_o1o1v1v1",
```

adcc/tests/LazyMp_test.py
```diff
@@ -222,6 +222,18 @@ def test_td2(self, system: str, case: str, generator: str,
                         refmp["mp2"]["td_o1o1v1v1"], atol=1e-12)
         assert "td2/o1o1v1v1" in instances.get(system, case).timer.tasks
 
+    @pytest.mark.parametrize("system,case",
+                             [(s, c) for s, c in small_cases if "cvs" not in c])
```
jonasleitner: There is a `non_cvs_small_cases` list defined at the top of the file.
Same for `sigma_inf_ov`, `m_3_plus`, `m_3_minus` and the `mp3_diffdm` tests

adcc/tests/properties_test.py
```diff
@@ -39,9 +39,13 @@
 # -> independent of method, case (gen/cvs/fc/fv) and kind (singlet/triplet)
 # Actually, the tests should also be independent of the systems, because
 # we only load some already tested density and contract it with some operator.
-methods = ["adc2"]
+
+methods = [
+    ("adc2", None),
+]
```
jonasleitner: Please remove the `isr_order` from the parametrization and instead just hardcode it as `None` in the tests

adcc/tests/state_densities_test.py
```diff
@@ -48,23 +57,34 @@
...
+    def test_state_diffdm(self, system: str, case: str, kind: str, adc_method: str,
+                          isr_order, generator: str):
+
+        if (
+            "cvs" in case and AdcMethod(adc_method).level.to_int() == 0
+            and generator == "adcman"
+        ):
             pytest.skip("No CVS-ADC(0) adcman reference data available.")
+        if adc_method == "adc4" and system in large_cases:
+            pytest.ski("ADC(4) reference data not generated for large cases.")
```
jonasleitner:
```suggestion
        if adc_method == "adc4" and system in large_cases:
            pytest.skip("ADC(4) reference data not available for large cases.")
```
Same for the other tests :D

adcc/tests/state_densities_test.py
```diff
@@ -135,7 +166,8 @@ def test_state_to_state_tdm(self, system: str, case: str, kind: str,
             fromi_ref_a = s2s_data[f"from_{i}"]["state_to_excited_tdm_bb_a"]
             fromi_ref_b = s2s_data[f"from_{i}"]["state_to_excited_tdm_bb_b"]
 
-            state_to_state = State2States(state, initial=i)
+            state_to_state = State2States(state, initial=i,
```
jonasleitner: Shouldn't `State2States` get the property method automatically from `state`?

adcc/tests/testdata_cache.py
```diff
@@ -213,11 +217,15 @@ def _make_mock_adc_state(self, system: Union[str, testcases.TestCase],
...
-            raise ValueError(f"No data available for kind {kind} in {case} "
-                             f"{method} {system}.")
-        # load the reference state and build a matrix on top
+        isr_data = data.get(str(isr_order), None)
+        if isr_data is None:
+            raise ValueError(
+                f"No data available for isr_order {isr_order} in case"
+                f"{method} {system}"
+            )
+        adc_data = isr_data.get(kind, None)
```
jonasleitner, line 226: `adc_data` might be None. Please add the original check for this case again.

adcc/tests/testdata_cache.py
```diff
@@ -213,11 +217,15 @@ def _make_mock_adc_state(self, system: Union[str, testcases.TestCase],
...
+        if isr_order is None:
+            isr_order = min(AdcMethod(method).level.to_int(), 2)
```
jonasleitner: It feels wrong that the default `isr_order` is determined here, again. Especially, since it is not 100% clear what the default for ADC(4) should be.

adcc/AdcMatrix.py
```diff
@@ -74,6 +74,7 @@ class AdcMatrixlike:
     _special_block_orders = {
         "adc2x": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 1},
         "isr1s": {"ph_ph": 1, "ph_pphh": None, "pphh_ph": None, "pphh_pphh": None},
+        "isr3d": {"ph_ph": 3, "ph_pphh": 2, "pphh_ph": 2, "pphh_pphh": 1},
         "isr2d": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 0},
```
jonasleitner:
```suggestion
        "isr2d": {"ph_ph": 2, "ph_pphh": 1, "pphh_ph": 1, "pphh_pphh": 0},
        "isr3d": {"ph_ph": 3, "ph_pphh": 2, "pphh_ph": 2, "pphh_pphh": 1},
```

adcc/AdcMethod.py
```diff
@@ -306,15 +306,17 @@ class IsrMethod(Method):
...
         LevelKey(
             adc_type=AdcType.PP,
             gs_type=GroundStateType.MP,
             cvs=True
         ): LevelSpec(
-            max_level=2,
-            special_levels=(MethodLevel.ONE_S, MethodLevel.TWO_D)
+            max_level=3,
+            special_levels=(MethodLevel.ONE_S, MethodLevel.TWO_D,
+                            MethodLevel.THREE_D)
```
jonasleitner: You did not implement CVS-ISR(3) and CVS-ISR(3)-d, right? So the spec for CVS-ISR should remain unchanged [max_level=2 and special levels=(MethodLevel.ONE_S, MethodLevel.TWO_D)]

adcc/ElectronicStates.py
```diff
@@ -82,6 +83,8 @@ def __init__(self, data, method: str = None,
         if not isinstance(self.method, AdcMethod):
             self.method: AdcMethod = AdcMethod(self.method)
 
+        if property_method is None and hasattr(data, '_property_method'):
+            property_method = data._property_method
```
jonasleitner, line 87:
```suggestion
        if property_method is None and isinstance(data, ElectronicStates):
            property_method = data._property_method
```

adcc/workflow.py
```diff
@@ -205,11 +209,19 @@ def run_adc(data_or_matrix, n_states=None, kind="any", conv_tol=None,
     if env_matrix_term:
         matrix += env_matrix_term
 
+    property_method = None
+    if isr_order is not None:
+        if isinstance(isr_order, str):
+            cvs_prefix = "cvs-" if matrix.method.is_core_valence_separated else ""
+            property_method = IsrMethod(f"{cvs_prefix}isr{isr_order}")
+        else:
+            property_method = matrix.method.at_level(isr_order).as_method(IsrMethod)
```
jonasleitner: I feel like there should be something like a `as_method_at_level` method on `AdcMethod` that generates the property method for us

adcc/tests/generators/generate_adcc_data.py
whyshna: the reason i did this is cause there are properties which doesnt need `isr_order`, for eg: `matrix`, so i was a bit doubtful
if `isr_order` is included in key, then we will need a different key for `matrix`, right?

adcc/tests/generators/generate_adcc_data.py
```diff
@@ -146,8 +156,11 @@ def generate_h2o_sto3g():
                 kwargs = {"n_guesses": 4}
             generate_adc_all(
                 test_case, method=method, dump_nstates=2, states_per_case=per_case,
-                **n_states, **kwargs
-            )
+                **n_states, **kwargs)
+            for isr_order in _isr_orders.get(method.level, (None,)):
```
jonasleitner: Can you please add all `method.level` to `_isr_orders`? Then please use here and for the other `generate_...` functions.

```suggestion
            for isr_order in _isr_orders[method.level]:
```
(Same for the adcman generator)

adcc/tests/generators/generate_adcman_data.py
```diff
@@ -15,6 +15,10 @@
 _methods = {
     "pp": ("adc0", "adc1", "adc2", "adc2x", "adc3")
 }
+_isr_orders = {
+    MethodLevel.THREE: (None, MethodLevel.THREE),
+    MethodLevel.FOUR: (None, MethodLevel.THREE_D),
+}
```
jonasleitner, line 25: The `_isr_order` dicts for the adcman and adcc generators should either both use `(None, 3, "3d")` or the explicit MethodLevels `None, MethodLevel.THREE, ...)`. I am fine with both :)

adcc/tests/properties_test.py
```diff
@@ -56,21 +58,26 @@
...
-                                       method: str, generator: str):
-        if "cvs" in case and AdcMethod(method).level == 0 and generator == "adcman":
+                                       adc_method: str, generator: str,
+                                       isr_order=None):
+
+        if (
+            "cvs" in case
+            and AdcMethod(adc_method).level == 0
+            and generator == "adcman"
+        ):
```
jonasleitner, line 73:
```suggestion
        if (
            "cvs" in case
            and AdcMethod(adc_method).level is MethodLevel.ZERO
            and generator == "adcman"
        ):
```
Same for `test_oscillator_strengths` and `test_state_dipole_moments`

adcc/tests/properties_test.py
```diff
@@ -125,31 +136,37 @@ def test_oscillator_strengths(self, system: str, case: str, kind: str,
...
-            system=system, method=method, case=case, source="adcc"
-        )[kind]
+            system=system, method=adc_method, case=case, source="adcc"
+        )[str(isr_order)][kind]
         state = testdata_cache._make_mock_adc_state(
-            system=system, method=method, case=case, kind=kind, source="adcc"
+            system=system, method=adc_method, case=case, kind=kind, source="adcc"
         )
+        if "state_ssq" not in refdata:
+            pytest.skip("state_ssq not available for this method yet")
```
jonasleitner: Is this skip needed? We are only testing adc2 with isr2. Thus this should never trigger I think?

adcc/tests/testdata_cache.py
```diff
@@ -213,11 +217,18 @@ def _make_mock_adc_state(self, system: Union[str, testcases.TestCase],
...
+            )
+        adc_data = isr_data.get(kind, None)
         if adc_data is None:
-            raise ValueError(f"No data available for kind {kind} in {case} "
-                             f"{method} {system}.")
-        # load the reference state and build a matrix on top
+            raise ValueError(
+                f"No data available for kind {kind} in case "
+                f"{method} {system}"
+            )
```
jonasleitner, line 231: Please add `isr_order` to the error message :)

adcc/tests/testdata_cache.py
```diff
@@ -233,6 +244,11 @@ def _make_mock_adc_state(self, system: Union[str, testcases.TestCase],
         states.kind = kind
         states.eigenvalues = adc_data["eigenvalues"]
 
+        if isr_order is not None:
+            states._property_method = IsrMethod(
+                f"{'cvs-' if 'cvs' in method else ''}isr{isr_order}"
+            )
```
jonasleitner:
```suggestion
        if isr_order is not None:
            states._property_method = method.as_method_at_level(IsrMethod, isr_order)
```

adcc/ElectronicStates.py
```diff
@@ -91,6 +94,7 @@ def __init__(self, data, method: str = None,
                 property_method = self.method.at_level(2).as_method(IsrMethod)
```
jonasleitner:
```suggestion
                property_method = self.method.as_method_at_level(IsrMethod, 2)
```

adcc/adc_pp/matrix.py
```diff
@@ -965,7 +965,7 @@ def adc3_i1(hf, mp, intermediates):
 def adc3_i2(hf, mp, intermediates):
     # Used only for general
     td2 = mp.td2(b.oovv)
-    p0 = mp.mp2_diffdm
+    p0 = mp.diffdm(level=2)
```
frieschneider:
```suggestion
    p0 = mp.second_order_dm_correction()
```

adcc/adc_pp/transition_dm.py
```diff
@@ -100,12 +103,80 @@ def tdm_isr2(mp, amplitude, intermediates):
...
+    t2_2 = ground_state.td2(b.oovv)
+    t3_2 = ground_state.tt2(b.ooovvv)
+    t2_3 = ground_state.td3(b.oovv)
+
+    p0_3_ov = ground_state.third_order_dm_correction().ov
+    p0_3_oo = ground_state.third_order_dm_correction().oo
+    p0_3_vv = ground_state.third_order_dm_correction().vv
+
+    p0_2_oo = ground_state.second_order_dm_correction().oo
+    p0_2_vv = ground_state.second_order_dm_correction().vv
```
frieschneider:
```suggestion
    p0_2 = ground_state.second_order_dm_correction()
    p0_2_vv, p0_2_oo = p0_2.vv, p0_2.oo

    p0_3 = ground_state.third_order_dm_correction()
    p0_3_vv, p0_3_ov, p0_3_oo = p0_3.vv, p0_3.ov, p0_3.oo
```

jonasleitner (approved):

PR #211, opened 2026-05-13
Full 2p-ISR(2) difference density

jonasleitner:
Add triples ISR(2) contributions to the `state_diffdm` and add a new `2d` special level to explicitly exclude them. Also the generation of the default block orders of the `AdcMatrix` and the `IsrMatrix` has been generalized for n-particle operators. Moreover, `AdcType` and `GroundStateType` enums have been introduced on the `AdcMethod` class to handle IP/EA- and RE/REMP-ADC in the future

### DIFF
--- a/adcc/adc_pp/state_diffdm_2p.py
+++ b/adcc/adc_pp/state_diffdm_2p.py
@@ -20,16 +20,17 @@
 ## along with adcc. If not, see <http://www.gnu.org/licenses/>.
 ##
 ## ---------------------------------------------------------------------
-from adcc import block as b
-from adcc.LazyMp import LazyMp
-from adcc.AdcMethod import IsrMethod
-from adcc.functions import einsum, zeros_like
-from adcc.Intermediates import Intermediates
-from adcc.AmplitudeVector import AmplitudeVector
-from adcc.TwoParticleDensity import TwoParticleDensity
-from adcc.NParticleOperator import OperatorSymmetry
-
-from .util import check_doubles_amplitudes, check_singles_amplitudes
+from .. import block as b
+from ..LazyMp import LazyMp
+from ..AdcMethod import IsrMethod
+from ..functions import einsum, zeros_like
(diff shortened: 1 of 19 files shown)


adcc/adc_pp/state_diffdm_2p.py
```diff
@@ -364,11 +365,27 @@ def diffdm_isr2_2p(mp, amplitude, intermediates):
     return dm
 
 
+def diffdm_isr2_2p(mp: LazyMp, amplitude: AmplitudeVector,
+                   intermediates: Intermediates) -> TwoParticleDensity:
```
frieschneider, line 375: Please add type hints to the other functions as well, so that we are at least consistent within the file.

adcc/adc_pp/util.py
```diff
@@ -32,6 +33,11 @@ def check_doubles_amplitudes(spaces, *amplitudes):
     check_doubles_subspaces(spaces, *amplitudes)
 
 
+def check_triples_amplitudes(spaces: list[str], *amplitudes: AmplitudeVector):
```
frieschneider, line 36: Same here. Please be consistent in a file with the type hints. :)

adcc/tests/AdcMethod_test.py
```diff
@@ -27,7 +27,8 @@
 adc_methods = [("adc1", None), ("adc2x", None), ("cvs-adc3", None),
                ("adc", ValueError), ("cvs_adc2", ValueError),
                ("xyz-adc2", ValueError), ("adc5", NotImplementedError),
-               ("isr2", ValueError)]
+               ("isr2", ValueError), ("mp-adc2", None), ("cvs-mp-adc2", None),
+               ("cvs-cvs-adc2", ValueError), ("adcc", ValueError)]
```
frieschneider: Should'nt we test the `adc_type` assignment for `"pp-adc2"` for example?

frieschneider (approved):

PR #215, opened 2026-06-11
Reduce PySCF gradient contraction memory

### LINKED ISSUE #214: Reduce PySCF gradient memory by streaming AO-pair 2PDM / ERI-derivative contractions
### Context

The nuclear-gradient implementation is currently experimental and lives on the `gradients` feature branch. Before merging this functionality, the PySCF backend should avoid the current `O(N_AO^4)` memory bottlenecks in the two-electron gradient contraction.

### Problem

The current PySCF nuclear-gradient implementation materializes two large `O(N_AO^4)` objects:

```python
g2_ao_1, g2_ao_2 = g2_total.to_ao_basis()
```

and

```python
ERIx = mol.intor("int2e_ip1", aosym="s1")
```

This creates two major memory bottlenecks:

1. full AO-transformed two-particle density matrices;
2. full derivative ERI tensors.

For moderate basis sizes this becomes impractical.

### Goal

Implement a memory-bounded PySCF contraction path for the two-electron gradient term that:

- avoids full `(nao, nao, nao, nao)` AO 2PDM arrays;
- avoids full `(3, nao, nao, nao, nao)` derivative ERI arrays;
- uses adcc's existing tensor contraction/block interface for staged MO-to-AO transformations;
- stores or streams packed AO-pair density blocks;
- contracts these blocks with shell-batched PySCF derivative ERIs.

The existing full-AO implementation should remain available as a validation/debug fallback.

For prototyping, it is acceptable to export individual adcc tensor blocks to NumPy arrays and implement the staged AO-pair transformation with dense batched operations. The production goal should be to keep density construction, block symmetry, and block extraction close to the existing libtensor-backed adcc representation, while using memory-bounded dense kernels for the actual MO-to-AO transformation chunks.

The key requirement is not that the entire transformation is one giant tensor expression. The implementation should avoid forming full AO rank-4 tensors. It may use libtensor/adcc tensors for block algebra and extraction, then transform dense block chunks into packed AO-pair buffers.

### Current reference algebra

The current implementation computes:

```python
Gradient["TEI"][ia] += np.einsum(
    "pqrs,xprqs->x", g2_ao_1, ERIx_a, optimize=True
)
Gradient["TEI"][ia] -= np.einsum(
    "pqrs,xpsqr->x", g2_ao_2, ERIx_a, optimize=True
)
```

where `ERIx_a` is built from the full first-center derivative integral tensor plus explicit four-center symmetrization:

```python
ERIx_a = np.zeros_like(ERIx)
ERIx_a[:, k0:k1] = ERIx[:, k0:k1]
ERIx_a += (
    + ERIx_a.transpose(0, 2, 1, 4, 3)
    + ERIx_a.transpose(0, 3, 4, 1, 2)
    + ERIx_a.transpose(0, 4, 3, 2, 1)
)
```

A useful equivalent view is to define an effective AO density in ERI index order:

```text
D[p,r,q,s] = g2_ao_1[p,q,r,s] - g2_ao_2[p,q,s,r]
```

The new implementation must preserve this contraction exactly, including all prefactors already applied when forming `g2_total`.

### Proposed implementation strategy

Implement a new backend path, for example:

```python
PyScfGradientProvider.correlated_gradient_direct(...)
```

which receives the MO/block-sparse `TwoParticleDensityMatrix`, AO one-particle quantities, and reference-state MO coefficients.

Conceptually:

```text
MO block T[i,j,k,l]
C_mu_i, C_nu_k, C_lambda_j, C_sigma_l
  -> D[mu,nu,lambda,sigma]
  -> packed AO-pair density D_pair[munu, lamsig]
```

but this must be done in memory-bounded stages, not by forming a dense rank-4 AO tensor.

A reasonable division of responsibilities is:

1. use adcc/libtensor-backed tensors for density algebra and block access;
2. extract individual MO 2PDM blocks or subblocks;
3. use dense batched transformations for memory-bounded chunks;
4. pack AO pairs into a compact AO-pair representation;
5. stream PySCF derivative ERIs shell-batch-wise and contract immediately.

### AO-pair density representation

Introduce a packed AO-pair density buffer. Conceptually:

```text
D_pair[munu, lamsig]
```

with, for a simple initial convention:

```text
nao_pair = nao * (nao + 1) // 2
D_pair.shape == (nao_pair, nao_pair)
```

Use HDF5/tempfile-backed storage if needed.

The exact convention must be documented carefully:

- whether off-diagonal AO pairs are summed, averaged, or stored with later factors;
- whether both bra and ket AO pairs are packed, or only the ket pair is packed to match PySCF `aosym="s2kl"`;
- how this convention maps onto the derivative integral contraction.

#### 1. Shell-batched derivative-ERI contraction using current AO tensors

First keep the current `g2_ao_1/g2_ao_2` tensors, but replace the full

```python
mol.intor("int2e_ip1", aosym="s1")
```

allocation by shell-batched PySCF calls such as:

```python
mol.intor(
    "int2e_ip1",
    comp=3,
    aosym="s2kl",
    shls_slice=(b0, b1, 0, mol.nbas, 0, mol.nbas, 0, mol.nbas),
)
```

This validates the new shell-batched contraction against the current implementation before changing the 2PDM transform.

This milestone isolates the derivative-integral batching problem from the AO-pair-density transformation problem.

#### 2. Introduce packed AO-pair density

Build a packed AO-pair representation of the effective density:

```text
D_pair[munu, lamsig]
```

from

```text
D[p,r,q,s] = g2_ao_1[p,q,r,s] - g2_ao_2[p,q,s,r]
```

using the current full AO tensors temporarily.

Validate the packed contraction against milestone 1. This isolates AO-pair packing and packed contraction conventions before changing the MO-to-AO transformation.

#### 3. Replace full AO 2PDM formation

Use staged adcc tensor contractions / dense block transformations to transform MO blocks directly into packed AO-pair density batches.

For each nonzero `TwoParticleDensityMatrix` block:

1. determine the orbital spaces;
2. apply the same spin expansion currently used by `to_ao_basis`;
3. insert contributions directly into the effective ERI-ordered density:

```text
D[p,r,q,s] += g2_ao_1[p,q,r,s]
D[p,r,q,s] -= g2_ao_2[p,q,s,r]
```

without forming `g2_ao_1` or `g2_ao_2`.

Start with the simplest useful case, e.g. MP2 / `oovv`, then generalize to all supported gradient blocks.

A practical staged transform for one block/spin case is:

1. choose a chunk of the second AO/MO pair, e.g. `(q,s)` or `(lambda,sigma)`;
2. transform the first MO pair `(p,r)` to an AO pair `(mu,nu)`;
3. pack `(mu,nu)`;
4. transform the second MO pair `(q,s)` to `(lambda,sigma)`;
5. pack `(lambda,sigma)`;
6. accumulate into the AO-pair density buffer.

#### 4. Add memory controls and optional out-of-core storage

Add configurable block sizes based on available memory.

Use an HDF5/tempfile-backed packed AO-pair density buffer if the packed density cannot fit comfortably in memory.

#### 5. Production integration

Add a new backend method, e.g.

```python
PyScfGradientProvider.correlated_gradient_direct(...)
```

or a similarly named method.

Keep the existing full-AO implementation as a debug/validation fallback until the direct path is fully validated.

### Spin expansion to preserve

The current `TwoParticleDensityMatrix.to_ao_basis` expands spin cases as follows.

For `g2_ao_1`:

```text
aaaa
bbbb
abab
baba
```

More explicitly:

```text
aaaa: C_a(s1), C_a(s2), C_a(s3), C_a(s4)
bbbb: C_b(s1), C_b(s2), C_b(s3), C_b(s4)
abab: C_a(s1), C_b(s2), C_a(s3), C_b(s4)
baba: C_b(s1), C_a(s2), C_b(s3), C_a(s4)
```

For `g2_ao_2`:

```text
aaaa
bbbb
abba
baab
```

More explicitly:

```text
aaaa: C_a(s1), C_a(s2), C_a(s3), C_a(s4)
bbbb: C_b(s1), C_b(s2), C_b(s3), C_b(s4)
abba: C_a(s1), C_b(s2), C_b(s3), C_a(s4)
baab: C_b(s1), C_a(s2), C_a(s3), C_b(s4)
```

The new path must reproduce these contributions exactly, including the exchange-like index permutation in the effective density:

```text
D[p,r,q,s] += g2_ao_1[p,q,r,s]
D[p,r,q,s] -= g2_ao_2[p,q,s,r]
```

This is one of the most important places where sign/index mistakes can occur.

### Existing block prefactors must be preserved

Do not change existing block prefactors applied during `g2_total` construction. The direct path should consume the same already-scaled `g2_total` as the full-AO path.

For non-CVS/non-core cases, current examples include:

```python
g2a.oovv *= 0.5
g2a.oooo *= 0.25
g2a.vvvv *= 0.25
```

For CVS/core cases, current examples include:

```python
g2a.oovv *= 0.5
g2a.ccvv *= 0.5
g2a.occv *= 2.0
g2a.vvvv *= 0.25
```

### PySCF derivative-ERI contraction details

The full derivative integral allocation should be replaced by shell-batched calls. The implementation must carefully match the density packing convention to PySCF's integral layout.

In particular:

- `aosym="s2kl"` packs/symmetrizes the ket AO pair;
- if both AO pairs are packed in `D_pair`, the contraction code must extract/unpack the density slice appropriately;
- alternatively, the packed density format can be chosen to match the PySCF integral layout more directly.

There are two possible accumulation styles:

#### Option A: first-center derivative calls plus density/index permutations

Use PySCF's `int2e_ip1`, which differentiates the first ERI center, and call/contract it for shell batches corresponding to each ERI slot moved into the first position.

This may be the easiest Python implementation, but the density permutations and signs must be checked carefully against the current full-AO reference.

#### Option B: shell-quartet-style accumulation

Loop over shell quartets and accumulate derivative contributions to all four centers directly:

```text
center a: derivative components 0..2
center b: derivative components 3..5
center c: derivative components 6..8
center d: derivative components 9..11
```

PySCF may not expose the derivative tensor in exactly this form, so option A may be simpler initially.

### Testing strategy

Add tests that compare:

1. current full-AO path;
2. shell-batched ERI path with full AO density;
3. packed AO-pair path;
4. final MO-block direct path.

Use tiny molecules/bases initially, because the goal is exact algebraic agreement before performance optimization.

Tests should cover at least:

- MP2 / `oovv`-dominated case;
- at least one ADC case using additional 2PDM blocks;
- RHF and, if supported in gradients, UHF/spin-resolved cases;
- CVS/core blocks if the direct path is extended to CVS gradients.

### Things to be careful about

- Do not change existing block prefactors applied during `g2_total` construction.
- Be explicit about pair-packing convention: whether off-diagonal AO-pair permutations are summed, averaged, or stored separately.
- Match the PySCF integral symmetry used in the shell-batched contraction.
- Validate against the old implementation on tiny systems before optimizing.
- Keep the full-AO path as a fallback until the direct path is fully tested.
- Do not mix up `g2_ao_1` and `g2_ao_2` spin/index conventions.
- PySCF `aosym="s2kl"` does not by itself imply a fully packed bra+ket AO-pair density layout.

### Acceptance criteria

- New direct PySCF path agrees with the existing full-AO path within current gradient tolerances.
- The new path does not allocate full AO 2PDM tensors.
- The new path does not allocate full derivative ERI tensors.
- Existing MP2/ADC gradient tests continue to pass.
- The implementation documents tensor index order, spin expansion, pair-packing convention, derivative-integral layout, and prefactors clearly.

maxscheurer:
Closes #214

Draft PR for reducing PySCF gradient memory by adding a staged, memory-bounded two-electron gradient contraction path.

Implementation plan will be posted as a PR comment.


maxscheurer:
## Implementation Plan

### Problem analysis

The current PySCF nuclear-gradient path has two independent fourth-order memory bottlenecks in the two-electron term:

- `nuclear_gradient()` forms dense AO `g2_ao_1` and `g2_ao_2` via `g2_total.to_ao_basis()`.
- `PyScfGradientProvider.correlated_gradient()` allocates the full derivative ERI tensor with `mol.intor("int2e_ip1", aosym="s1")`.

The existing implementation is the correctness reference. The new path must preserve the current block prefactors, spin expansion, derivative-integral sign, and four-center derivative symmetrization exactly while progressively replacing the dense allocations.

### Deliverables

1. **Staged PySCF TEI contraction methods**
   - Add a PySCF-only shell-batched two-electron gradient contraction that initially accepts the existing dense AO 2PDM inputs.
   - Add packed AO-pair effective-density helpers once the shell-batched integral path is validated.
   - Add a final direct path that consumes the MO/block-sparse `TwoParticleDensityMatrix` and transforms blocks into the packed AO-pair representation without forming full AO rank-4 densities.

2. **Driver integration with fallback**
   - Add a clear selection mechanism in `nuclear_gradient()` so the existing full-AO path remains available as a debug/validation fallback.
   - Route PySCF to the direct path only when supported; leave other backends on the existing `correlated_gradient(...)` interface.
   - Preserve the current `GradientComponents` / `GradientResult` return contract.

3. **Documented conventions**
   - Document the effective density order `D[p,r,q,s] = g2_ao_1[p,q,r,s] - g2_ao_2[p,q,s,r]`.
   - Document the spin cases reproduced from `TwoParticleDensityMatrix.__transform_to_ao()`.
   - Document AO-pair packing, including off-diagonal pair handling and how it maps to PySCF `aosym="s2kl"` ket-pair packing.
   - Document that existing `g2_total` block prefactors are already applied upstream and are not re-applied in the direct path.

4. **Memory controls**
   - Add configurable shell/pair/block chunk sizing for the direct path.
   - Add optional out-of-core storage for the packed AO-pair density if in-memory storage is too large, reusing existing scratch/HDF5 infrastructure where practical.

5. **Focused staged tests**
   - Add tests that compare each new stage against the current full-AO PySCF reference on tiny systems before relying on finite-difference gradient tests.

### Files to modify or add

- `adcc/backends/pyscf.py`
  - Add the shell-batched derivative-ERI contraction.
  - Add packed-density contraction helpers and the final direct PySCF gradient method.
  - Keep `correlated_gradient(...)` unchanged as the full-AO reference/fallback.

- `adcc/gradients/__init__.py`
  - Add the routing/selection point for the PySCF direct path.
  - Avoid `g2_total.to_ao_basis()` when the direct path is selected.
  - Keep existing TPDM formation and block prefactors unchanged.

- `adcc/gradients/TwoParticleDensityMatrix.py` or a new gradients helper module
  - Factor or mirror the block/spin iteration needed by the direct transform.
  - Keep the existing `to_ao_basis()` behavior intact for fallback and validation.

- `adcc/tests/gradient_contraction_test.py` (new)
  - Add fdiff-free staged equality tests for the TEI contraction substeps.

- Existing gradient tests and fixtures
  - Leave `adcc/tests/functionality_gradients_test.py`, `adcc/tests/data/gradient_data.json`, and `adcc/tests/generators/dump_fdiff_gradient.py` unchanged unless end-to-end integration needs a small selector adjustment.

### Testing approach

1. **Stage 1: shell-batched ERI test**
   - Use PySCF with `h2o/sto3g` and build the same dense `g2_ao_1`, `g2_ao_2`, `g1_ao`, and `w_ao` as the current path.
   - Compare the new shell-batched TEI contribution against `PyScfGradientProvider.correlated_gradient(...)`.
   - Verify atom-by-atom Cartesian TEI agreement with tight algebraic tolerance.

2. **Stage 2: packed AO-pair density test**
   - Build the effective density from the dense AO tensors temporarily.
   - Pack it using the documented convention.
   - Compare packed contraction against the stage 1 shell-batched result.
   - Include cases that exercise off-diagonal AO pairs.

3. **Stage 3: direct MO-block transform test**
   - Transform `g2_total.blocks_nonzero` directly into packed AO-pair density.
   - Compare against the dense AO `to_ao_basis()` path.
   - Start with MP2, then add an ADC case that exercises additional blocks, then add CVS coverage once CVS blocks are implemented.

4. **End-to-end gradient regression**
   - Keep existing `test_nuclear_gradient` passing for MP2/ADC and supported backends.
   - Add a small PySCF-only end-to-end direct-vs-fallback test once the selector exists.
   - Keep `h2o/sto3g` in fast mode; place heavier `ccpvdz`, UHF, and CVS expansion cases behind existing full-mode mechanisms or skips until supported.

### Acceptance criteria

- The shell-batched PySCF TEI contraction agrees with the current full-ERI PySCF reference on tiny systems.
- The packed AO-pair contraction agrees with the dense AO-density shell-batched contraction.
- The final direct path agrees with the existing full-AO path within current gradient tolerances.
- The direct path does not allocate full AO 2PDM tensors.
- The direct path does not allocate the full derivative ERI tensor.
- Existing MP2/ADC gradient tests continue to pass.
- Unsupported cases, if any during staging, fall back clearly or raise a documented error rather than silently using incorrect algebra.

### Risks and open questions

- The highest risk is spin/index bookkeeping: `g2_ao_1` uses `aaaa`, `bbbb`, `abab`, `baba`, while `g2_ao_2` uses `aaaa`, `bbbb`, `abba`, `baab`, then contracts with the exchange-like `psqr` order.
- PySCF `aosym="s2kl"` packs only the ket AO pair. The density layout must either match this directly or unpack/slice explicitly.
- Shell slicing uses shell ranges, while current atom slicing uses AO ranges. Mixing these would produce wrong atom contributions.
- The current PySCF gradient provider uses `grad.RHF(self.scfres)`; UHF/spin-resolved gradient behavior should be treated as a pre-existing limitation unless explicitly implemented.
- CVS/core blocks increase the block/index surface area and should be staged after the non-CVS path is validated.
- PySCF derivative-integral conventions need validation in an environment with PySCF available.

---
*Plan created by mach6*

maxscheurer:
## Progress Update

Implemented the staged PySCF gradient memory-contraction path and added a profiling playground:

- Added shell-batched PySCF derivative-ERI contraction as an intermediate validation path.
- Added packed AO-pair effective-density helpers with documented ordering and spin conventions.
- Added direct MO/block-sparse TPDM to packed AO-pair transformation, avoiding full AO rank-4 TPDMs for the PySCF direct path.
- Added routing in `nuclear_gradient()` with `auto`, `direct`, `shell_batched`, and `full_ao` selection.
- Added memory controls for shell chunks, AO-pair chunks, and optional HDF5 packed-density storage.
- Added focused staged tests comparing shell-batched, packed-density, direct, and full-AO paths.
- Added `scripts/profile_pyscf_gradient_memory.py`, defaulting to benzene/STO-3G, with in-process memray tracking around the gradient call only.

Verification run before commit:

- `pytest -q adcc/tests/gradient_contraction_test.py --skip-update` passed.
- `pytest -q adcc/tests/functionality_gradients_test.py --skip-update` passed.
- Direct HDF5 packed-density smoke check matched full-AO to numerical precision.
- Benzene/STO-3G profiling script smoke checks completed for `direct` and `full_ao`.

**Commit:** `a42ccef`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

Removed the unused intermediate shell-batched public code path so there are only two maintained TEI contraction modes:

- `direct` for the new PySCF memory-reduced path.
- `full_ao` for the existing/reference fallback path.

Changes in this batch:

- Removed `PyScfGradientProvider.correlated_gradient_shell_batched(...)`.
- Removed `shell_batched` and `auto` from explicit `tei_contraction` choices; default selection is now internal via `tei_contraction=None`.
- Updated staged tests to compare packed-density/direct helpers directly against the full-AO reference.
- Updated the memray profiling playground to expose only `direct` and `full_ao`.
- Confirmed no stale references to `shell_batched` or `correlated_gradient_shell_batched` remain.

Verification run before commit:

- `pytest -q adcc/tests/gradient_contraction_test.py --skip-update` passed.
- `pytest -q adcc/tests/functionality_gradients_test.py --skip-update` passed.
- Profiling script water direct-vs-full smoke check matched full-AO to numerical precision.

**Commit:** `ee3fb8c`

---
*Progress tracked by mach6*

maxscheurer:
## Plan: move the libtensor → dense handover *later* in the direct gradient path

### Where the switch is today

The `direct` path (`TwoParticleDensityMatrix.to_ao_pair_density`) hands over to NumPy **very early** — right after extracting each raw MO 2PDM block via `tensor.to_ndarray()`. Everything downstream runs in NumPy `einsum`/BLAS:

- the spin expansion (`aaaa`/`bbbb`/`abab`/`baba` for `g2_ao_1`, `abba`/`baab` for `g2_ao_2`),
- the bra-pair and ket-pair MO→AO transforms,
- the exchange-index permutation that builds `D[p,r,q,s] = g2_ao_1[p,q,r,s] − g2_ao_2[p,q,s,r]`,
- the `aosym="s2kl"` packing.

libtensor only does the upstream `g2_total` assembly. So all the spin/sign bookkeeping currently lives in the dense layer as hand-written NumPy index gymnastics.

### Where I want the switch

Push the spin expansion and (at least) the **bra-pair half-transform** `(p,r) → (μ,ν)` back behind the libtensor boundary, exporting dense arrays only at a **half-transformed intermediate** or later. The existing `__transform_to_ao` already proves the *full* MO→AO transform is expressible as a single libtensor `einsum` (`ip,jq,ijkl,kr,ls->pqrs`) including the full spin expansion — so the algebra is already there; we just stop it short of forming the full AO rank-4 and avoid re-deriving the spin/sign logic by hand.

### Honest note on what this buys

All BLAS (libtensor included) goes through the same OpenBLAS here — moving work from NumPy `einsum` to libtensor `einsum` does **not** change the matmul kernel. The wins are:

1. **block-sparse symmetry exploitation** — zero spin blocks and pair antisymmetry handled automatically instead of looped over in Python;
2. **consolidation** — the `g2_ao_1`/`g2_ao_2` distinction and the exchange permutation become symbolic einsum terms rather than NumPy transpose/sign flips that are easy to get subtly wrong.

### Don't give in to NumPy too early

The right boundary should be driven by *what libtensor can express memory-bounded*, not by what is convenient in NumPy. If pushing the transform later needs **C++/libtensor support** (e.g. slicing/chunking the AO axis of the coefficient tensors so we can produce the AO output in batches without ever forming the full rank-4, or a dedicated half-transform primitive), that is on the table and preferable to falling back to dense NumPy prematurely. The only step that is *intrinsically* a scatter/gather and not an einsum is the `s2kl` packing — that one legitimately stays in NumPy. Everything up to it should be a candidate for staying symbolic.

### Milestones

1. **Guardrail tests first.** Pin today's numbers at *intermediate* granularity (half-transformed density and fully packed density), not just the final gradient — the refactor moves the boundary, so we need fixed reference values *at* the boundary.
2. **Spike libtensor expressivity.** Determine whether a memory-bounded MO→AO transform can stay symbolic: can we slice the AO axis of the coefficient tensors to drive chunked AO output, or does libtensor only support the full transform at once? **This is the decision gate** — and the place where a C++-side extension may be warranted rather than retreating to NumPy.
3. **Lift spin + bra-pair transform into libtensor.** Reuse the `__transform_to_ao` spin machinery to produce a half-transformed `D[μ,ν,q,s]` (AO bra pair, MO ket pair) in libtensor; export dense only there.
4. **Push the ket-pair transform `(q,s) → (λ,σ)` into libtensor where feasible.** Keep only the `s2kl` gather/pack in NumPy.
5. **Leave the ERI side untouched.** `_contract_tei_with_packed_density` and the PySCF `int2e_ip1` shell-batched streaming don't change; only the producer feeding `pair_density` changes.
6. **Validate + profile.** Exact agreement (`atol=1e-10`) vs the `full_ao` fallback across RHF/UHF/CVS, then measure peak memory and time.

### Key risk

Moving the boundary later risks materializing a large half-transformed intermediate. The half-transformed `D[μ,ν,q,s]` is `O(nao²·nmo²)`; the whole point of the direct path is avoiding `O(nao⁴)`. The plan must keep that intermediate chunked along the ket pair — which is exactly what milestone 2 must prove is expressible (and, if not expressible today, is the concrete motivation for a C++-side chunked-transform primitive).

maxscheurer:
## Progress Update

Moved the block-tensor → dense boundary later in the direct PySCF gradient path, so only the spin sub-blocks that actually contribute are densified, instead of materialising full zero-padded TPDM blocks and full rank-4 AO tensors.

### Changes
- **New C++ primitive `Tensor.export_block(start, end)`** — exports a dense rank-N sub-block over a half-open per-axis window without materialising the full tensor. It mirrors the `btod_export` scatter (applying the inverse orbit permutation), so it stays correct even on antisymmetric same-space blocks where a naive per-element gather would flip a sign.
- **`to_ao_pair_density` rewritten** — for each spin case it extracts only the relevant rank-4 spin sub-block via `export_block` and contracts it with the compacted (non-zero-row) coefficients. The alpha/beta row split is structural (`[0, n_alpha)` / `[n_alpha, n_orbs)`), so coefficient rows and TPDM sub-block share the same range by construction. Sub-blocks are pre-extracted once and reused across pair chunks; the packed density is written once per chunk to keep the out-of-core path I/O-efficient.
- The derivative-ERI consumer (`int2e_ip1` streaming contraction) is unchanged — only the density producer was rewired.

### Validation
- Machine-precision agreement (≤ 2e-14) with the full-AO path end-to-end: RHF / UHF / CVS, cc-pVDZ and cc-pVTZ, both in-memory and HDF5 storage.
- 39 tests pass (19 contraction guardrails including new `export_block` and spin-subblock tests + 20 functionality gradient tests).
- Real peak RSS at cc-pVTZ (nao=58): full-AO 1697 MB → in-memory direct 1268 MB (−25%) → out-of-core 1064 MB (−37%). The saving grows with basis size, since the full path allocates a `3·nao⁴` derivative-integral tensor.

### Build note
The extension must currently be built with `-fvisibility-inlines-hidden` removed from `CXXFLAGS`; with that flag the block-tensor symmetry product-table singleton fails to resolve at runtime. This is independent of the changes here but needs handling in the build/CI configuration.

**Commit:** `7268528`

---
*Progress tracked by mach6*

maxscheurer:
## Code Review

Five review agents were run (code-reviewer, error-auditor, test-reviewer, completeness-checker, simplifier). The core two-electron contraction algebra was independently verified by multiple agents and is **correct**: the spin cases (`aaaa`/`bbbb`/`abab`/`baba` for `g2_ao_1`; `aaaa`/`bbbb`/`abba`/`baab` for `g2_ao_2`), the packed `s2kl` off-diagonal symmetrization, and the C++ `export_block` inverse-permutation/scalar-transform scatter all match the reference `__transform_to_ao` and `full_ao` paths. **No correctness findings.** All findings below are about memory-bound effectiveness, error handling, test coverage, and maintainability.

### Critical

_None._

### Important

**Item 1 — HDF5 out-of-core storage does not bound peak memory with the default chunk size.**
`adcc/backends/pyscf.py` (`correlated_gradient_direct`) + `adcc/gradients/TwoParticleDensityMatrix.py` (`to_ao_pair_density`). When `pair_chunk_size is None` it defaults to the full `npair`, so the in-memory working buffers `chunk` (≈ `nao⁴/2`) and `swapped` (≈ half that) are allocated at full size even when `pair_density_storage="hdf5"`. A user requesting the documented out-of-core path (with the default `tei_pair_chunk_size=None` everywhere, including `nuclear_gradient` and the profiling script) gets essentially the same peak RAM as `"memory"` mode, with no warning — the feature silently fails to deliver its stated purpose. Fix: pick a finite default `pair_chunk_size` for hdf5/outcore, or warn.

**Item 2 — HDF5 / out-of-core storage path is entirely untested.**
`tei_pair_density_storage` is never set to anything but `"memory"` in any test and is never passed through `nuclear_gradient`. The whole HDF5 branch is untested: temp-file creation/cleanup, the chunked `h5py` dataset, and — critically — the in-place `out[...] = 0` / `out[:, :, start:stop] += chunk` read-modify-write and the `chunk[:, :, offdiag] += swapped` fancy-index assignment, which behave differently for an h5py dataset than a numpy array. A correctness or API bug here would ship undetected.

### Suggestions

**Item 3 — `finally` cleanup can mask the original exception.** `adcc/backends/pyscf.py` (~lines 237-244). `h5file.close()` is unguarded; on disk exhaustion a write raises `OSError(ENOSPC)`, then `close()` flushes to the full disk and raises a second exception that supersedes/discards the root cause. `os.unlink` only suppresses `FileNotFoundError`. Guard cleanup (e.g. `contextlib.suppress`) so it doesn't shadow the primary exception.

**Item 4 — Dead numpy fallback in `to_ao_pair_density`.** `adcc/gradients/TwoParticleDensityMatrix.py` (~lines 408-422). `export_block` is always present on a compiled `Tensor`, so `has_export_block` is always `True` and the `dense_block = np.asarray(ten_obj)` branch is dead. If ever reached it would densify the full zero-padded block — exactly the O(nao⁴) materialization the method exists to avoid — silently, with no error. Drop it or make it raise.

**Item 5 — Redundant `ao_pair_density_from_dense` wrapper on the provider.** `adcc/backends/pyscf.py` (lines 181-185). Pure pass-through to the static method, not used in the production path or tests, and re-imports `TwoParticleDensityMatrix` locally though it's already imported at module top. Remove to avoid a duplicate public API surface.

**Item 6 — `pair_density_storage="outcore"` accepted but undocumented and absent from the error message.** `adcc/backends/pyscf.py` (~lines 163-179). `"outcore"` is silently accepted as an alias for `"hdf5"`, but the docstring and the `ValueError` message only mention `"memory"`/`"hdf5"`. Mirror the explicit `valid_*` set pattern used in `gradients/__init__.py` and make the message match the accepted set.

**Item 7 — Error/validation paths untested.** Invalid `tei_contraction` (`ValueError`), `NotImplementedError` for non-PySCF `direct`, invalid `pair_density_storage`, `shell_chunk_size <= 0` / `pair_chunk_size <= 0`, empty-`blocks_nonzero` and "Invalid output shape" guards — all new user-facing API contracts, none tested. A few `pytest.raises(..., match=...)` would pin them.

**Item 8 — `shell_chunk_size` chunk-boundary independence not tested.** `pair_chunk_size` independence is parametrized over `[1, 2, 3, 1000]`, but the analogous guard for the contraction side (`shell_chunk_size` driving `int2e_ip1` shell slicing and atom-local `ao_loc` ranges) is missing — every contraction test uses `shell_chunk_size=2`. A boundary bug there would only appear for particular chunk sizes.

**Item 9 — No direct-vs-`full_ao` equivalence test for a multi-block (ADC) density through `nuclear_gradient`.** The only end-to-end direct-vs-full_ao assertion uses MP2 (essentially `oovv`). Multi-block spin expansion is tested only on synthetic `_random_tpdm` data via the helpers directly, not through a real ADC excited-state density routed through `nuclear_gradient(..., tei_contraction="direct")` and compared tightly (≤1e-10). The finite-difference functionality test validates the absolute gradient at a loose tolerance that can mask small block-handling discrepancies.

**Item 10 — Core/CVS 2PDM blocks not exercised at the unit level.** `_random_tpdm` and the `export_block` spin-subblock test use only `oooo/oovv/ovov` — no core (`c`) subspace. CVS-ADC gradients do route through the direct path in `functionality_gradients_test.py` vs fdiff (loose 2e-7), but core-block handling in `to_ao_pair_density`/`export_block` is not directly pinned. Adding a core-containing block would give targeted coverage.

**Item 11 — `export_block` validation test uses an overly broad `pytest.raises(Exception)`.** `adcc/tests/gradient_contraction_test.py` (`test_export_block_validates_arguments`). Should assert the specific `invalid_argument` type and `match=` the distinct messages so the wrong-ndim and end-exceeds-shape branches are actually distinguished.

**Item 12 — Dead local `cblk_size` in `export_block`.** `libadcc/TensorImpl.cc` (~lines 1238, 1290). Computed then explicitly discarded with `(void)cblk_size;`. Leftover; drop both lines.

**Item 13 — PEP8 / formatting.** Three blank lines between `to_ao_pair_density` and `__iadd__` in `TwoParticleDensityMatrix.py` (~lines 479-481) and a similar double-blank gap after `export_block` in `TensorImpl.cc`. Project linter will flag.

**Item 14 — `_symmetrized_density_slice` perf (low priority).** `adcc/backends/pyscf.py` (lines 122-147). Recomputes `ao_pair_indices(nao)`/`offdiag` on every call (once per shell chunk per atom) and runs a nested Python loop of `(a1-a0)*nao` iterations; for HDF5-backed density it issues a scattered read per `(a,b)`. Hoist the `nao`-only quantities and vectorize where possible. Memory, not speed, is the PR's stated goal, so low priority.

### Strengths

- Core spin/index algebra is correct and matches the reference across all four functions and the C++ scatter — the highest-risk part of the issue.
- Deliberate, good coverage choices: open-shell (CN doublet, UHF, `n_α≠n_β`) reference exercises mixed-spin cases that cancel under RHF; synthetic `_random_tpdm` gives non-degenerate data in every block; explicit two-stage staging-contract reference pins the boundary for the planned libtensor refactor; `export_block` tested vs `to_ndarray` for random ranges and all 16 spin sub-blocks; deterministic seeding throughout.
- Both acceptance criteria on allocations are met: the direct path forms neither full AO 2PDM tensors nor the full derivative ERI tensor; `full_ao` remains a selectable fallback.
- Conventions (index order, spin expansion, `s2kl` packing, prefactors) are thoroughly documented in docstrings.
- The build-flag concern raised in the progress comment (`-fvisibility-inlines-hidden`) is already handled in CI (`.github/workflows/ci.yaml`), independent of this PR.

---
*Reviewed by mach6*

maxscheurer:
## Code Review — Simplifier (addendum)

Behavior-preserving simplifications (confidence ≥ 80), beyond the maintainability items already listed above:

**Item 15 — Fold the duplicated direct/exchange spin-case loop bodies in `to_ao_pair_density`.** `adcc/gradients/TwoParticleDensityMatrix.py` (~lines 440-464). The two inner loops are byte-for-byte identical except the `(spin_cases, sign, exchange)` triple. Iterate over `((direct_spin_cases, +1.0, False), (exchange_spin_cases, -1.0, True))` to remove ~12 lines. Order (direct then exchange) and all args preserved.

**Item 16 — Drop the redundant `.copy()` in `_symmetrized_density_slice`.** `adcc/backends/pyscf.py` (~line 139). `ket_as_bra[qidx, sidx]` is advanced indexing and already returns a fresh array, so the subsequent in-place ops can't alias `ket_as_bra`; the `.copy()` is unnecessary. (The `density_slice = ...copy()` above does need its copy — leave it.)

**Item 17 — Hoist loop-invariant `np.any(offdiag)` in `_symmetrized_density_slice`.** `adcc/backends/pyscf.py` (~lines 135-141). `offdiag` is fixed before the double loop but `np.any(offdiag)` is recomputed for every `(a,b)`. Precompute `any_offdiag` once, matching the pattern already used in `to_ao_pair_density`. (Overlaps with item 14.)

**Item 18 — Drop dead keys in `_empty_gradient`.** `adcc/backends/pyscf.py` (~lines 48-57). `_final_gradient_components` only uses `grad_nuc()`, `"S"`, `"T+V"`, `"TEI"`. The `"N"`, `"OEI"`, `"Total"` entries are never written or read on the pyscf path (psi4 backend uses them, pyscf does not). Removing them is behavior-preserving.

**Note:** `_ao_coefficient_map` vs `__transform_to_ao` coeff-building looks duplicated but is **not** safely unifiable — the latter must keep `Tensor` objects for libtensor `einsum`, the former needs ndarrays. Left as-is.

---
*Reviewed by mach6*

maxscheurer:
## Review Assessment

Independent assessment of the review findings, each verified against the actual source. Review comments: [findings](https://github.com/adc-connect/adcc/pull/215#issuecomment-4695318943) and [simplifier addendum](https://github.com/adc-connect/adcc/pull/215#issuecomment-4695359058).

**Verified highlights:** Finding 13 is a confirmed **CI blocker** — `flake8` (run with no args → uses `setup.cfg`) reports `E303` at `TwoParticleDensityMatrix.py:474` and `E303`/`E302` at `gradient_contraction_test.py:275,285`; the C++ triple blank at `TensorImpl.cc:1328-1330` violates `libadcc/.clang-format` (`MaxEmptyLinesToKeep: 1`). Finding 1 confirmed: `to_ao_pair_density` sets `pair_chunk_size = npair` when `None`, so the in-RAM `chunk` buffer is the full `(nao,nao,npair) ≈ nao⁴/2` even with `hdf5` storage (which only offloads the *output*). Findings 9/10 partially mitigated: `functionality_gradients_test.py` auto-routes pyscf to `direct` and includes `cvs-adc*`, so CVS/multi-block ADC *is* exercised vs finite-difference at the loose `atol=2e-7`.

### Classifications

|---|---|---|
| 1: HDF5 default chunk doesn't bound memory | **genuine** | `pair_chunk_size=None` → full `npair`; in-RAM working buffers are full nao⁴/2 while HDF5 only offloads output. Memory bound not achieved at default. |
| 2: HDF5/out-of-core path untested | **genuine** | New testable code; storage never set to `hdf5` in tests; in-place dataset writes unexercised. |
| 3: `finally` close masks exception | **genuine (minor)** | Unguarded `h5file.close()` flush error would supersede the root cause. Edge-case robustness. |
| 4: Dead numpy fallback in to_ao_pair_density | **genuine (minor)** | `export_block` always present → branch unreachable. Cleanup only. |
| 5: Redundant provider wrapper | **genuine (minor)** | Pure pass-through, never called in production or tests. |
| 6: `"outcore"` undocumented alias | **genuine (minor)** | Accepted in code, absent from docstring and `ValueError`. |
| 7: Error/validation paths untested | **genuine (test gap)** | New guards have no tests; should ship with the PR. |
| 8: shell_chunk_size boundary untested | **genuine (minor test gap)** | Shell side always 2; boundary bug would go uncaught. |
| 9: No direct-vs-full_ao multi-block test | **genuine (minor test gap)** | Tight comparison only for MP2; multi-block only synthetic/loose. |
| 10: Core/CVS blocks not at unit level | **genuine (minor test gap)** | Only oooo/oovv/ovov at unit level; CVS covered end-to-end (loose) but no core-subspace unit check. |
| 11: Broad `pytest.raises(Exception)` | **nitpick** | Test-quality preference, not correctness. |
| 12: Dead `cblk_size` + `(void)` cast | **nitpick** | Confirmed unused; trivial cleanup. |
| 13: PEP8 + C++ blank lines | **genuine (CI blocker)** | `flake8` and clang-format both fail. Must fix before merge. |
| 14: `_symmetrized_density_slice` perf | **deferred** | Per-call recompute + scattered HDF5 reads; performance, not correctness. |
| 15: Fold duplicated spin-case loops | **nitpick** | Stylistic refactor, no behavioral effect. |
| 16: Redundant `.copy()` | **nitpick** | Confirmed redundant (advanced indexing already copies); harmless. |
| 17: Hoist `np.any(offdiag)` | **nitpick** | Micro-opt, overlaps with 14. |
| 18: Dead `"N"`/`"OEI"`/`"Total"` keys | **nitpick** | Confirmed set but never read; harmless. |

**Counts:** 10 genuine (1 CI blocker), 7 nitpicks, 0 false positives, 1 deferred.

### Action Plan

1. **Fix finding 13 (CI blocker, do first).** Remove extra blank lines at `TwoParticleDensityMatrix.py:474`, `gradient_contraction_test.py:275` & `:285`, and `TensorImpl.cc:1328-1330`. Without this, `flake8` and clang-format CI fail and the PR cannot merge.
2. **Address finding 1 (headline-feature intent).** Choose a bounded default `pair_chunk_size` when storage is `hdf5`/`outcore`, or document clearly that `tei_pair_chunk_size` must be set for the bound to take effect (and fix the docstring).
3. **Add minimal tests for the out-of-core path and validation guards (findings 2, 7).** At least one `hdf5`-vs-`memory` equivalence smoke test plus a few `pytest.raises` checks. New testable code — ships with the PR.
4. **Resolve finding 6.** Document `"outcore"` (and add to the `ValueError` message) or remove the alias.
5. **Cleanup dead code (findings 4, 5, 12, 18).** Remove unreachable fallback, unused wrapper, discarded `cblk_size`, unused dict keys. Fold nitpicks 15/16/17 in opportunistically.
6. **Improve robustness of finding 3.** Guard `h5file.close()` so a `finally` flush error can't mask the original exception.

### Deferred (track separately, not a blocker)
One follow-up issue covering: finding 14 (`_symmetrized_density_slice` index/offdiag hoisting + batched HDF5 reads), finding 8 (shell-chunk-boundary test), and findings 9/10 (tight direct-vs-full_ao multi-block + core-subspace unit coverage). Correctness is already covered by existing equivalence and fdiff tests, so these strengthen rather than gate.

---
*Assessment by mach6*

maxscheurer:
## Progress Update

Addressed the genuine review findings from the assessment (findings 13, 1, 2, 7, 6, 4, 5, 12, 18, 3), plus the requested `outcore`-alias removal and dead-code cleanup.

### Fixes
- **CI blocker (finding 13):** removed the `flake8` blank-line violations (`TwoParticleDensityMatrix.py`, `gradient_contraction_test.py`) and the clang-format triple-blank in `TensorImpl.cc`; also fixed pre-existing E402 in the new profiling script via `# noqa`. `flake8` is now clean on all files changed by this PR.
- **HDF5 memory bound (finding 1):** the `hdf5` path now picks a bounded default `pair_chunk_size = min(npair, 256)` when unset, so the in-memory working buffer no longer silently grows to the full `(nao, nao, npair)`. The docstring documents the working-buffer vs output-density distinction.
- **Explicit storage validation (finding 6):** dropped the undocumented `"outcore"` alias entirely. Storage is validated against an explicit `{"memory", "hdf5"}` set with a matching error message, checked up front.
- **Guarded cleanup (finding 3):** the `finally` block now uses `contextlib.suppress` so a flush/unlink error cannot mask an exception raised inside the `try`.
- **Dead code (findings 4, 5, 12, 18):** removed the unreachable numpy `export_block` fallback, the redundant provider `ao_pair_density_from_dense` wrapper, the unused `cblk_size` in `Tensor::export_block`, and the unused `"N"`/`"OEI"`/`"Total"` keys in `_empty_gradient`.
- **Simplifications (items 15, 16, 17):** folded the duplicated direct/exchange spin-case loops, dropped a redundant `.copy()`, and hoisted a loop-invariant `np.any(offdiag)`.

### Tests added (findings 2, 7)
Eight new tests: `hdf5`-vs-`memory` gradient equivalence, scratch-file cleanup, a bounded-default-chunk test (mutation-verified — it fails if the bound is removed), and validation guards (invalid `tei_contraction`, invalid `pair_density_storage`, non-positive `shell_chunk_size`/`pair_chunk_size`, bad output shape).

### Verification
- `pytest -q adcc/tests/gradient_contraction_test.py --skip-update` — 27 passed.
- `pytest -q adcc/tests/functionality_gradients_test.py --skip-update` — 20 passed.
- C++ rebuilt cleanly under `-Werror`. The `-fvisibility-inlines-hidden` runtime issue surfaced again during the rebuild; stripping it (exactly as CI does at `ci.yaml:183`) produced a working extension.

### Deferred (follow-up, not in this batch)
Finding 14 (`_symmetrized_density_slice` per-call index/offdiag hoisting + batched HDF5 reads) and the test-strengthening of findings 8, 9, 10 (shell-chunk-boundary test; tight direct-vs-`full_ao` multi-block/core-subspace coverage). Correctness is already covered by existing equivalence and finite-difference tests.

**Commit:** `9ff4566`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

Reframed the AO-pair packing docstrings in `TwoParticleDensityMatrix.py` to be backend-neutral, addressing the design concern that the class read as if it assumed PySCF.

The class never had a functional PySCF dependency (no PySCF imports, types, or calls; method names were already backend-neutral). The only coupling was docstring wording that framed a generic convention as PySCF-specific. The lower-triangular symmetric AO-pair packing (`max*(max+1)//2 + min`) is a standard quantum-chemistry convention; PySCF `aosym="s2kl"` is just one concrete consumer.

### Changes (docstrings only, no logic touched)
- `ao_pair_index` / `ao_pair_indices`: describe the lower-triangular symmetric-pair packing as the general convention; PySCF demoted to a parenthetical "e.g." example.
- `ao_pair_density_from_dense`: "for a derivative-ERI contraction"; off-diagonal summation explained via the generic symmetric-ket-pair layout.
- `to_ao_pair_density`: "for the gradient two-electron contraction"; cross-references `ao_pair_index` for the packing.

The PySCF-specific `aosym="s2kl"` calls remain where they belong, in `adcc/backends/pyscf.py`.

`flake8` clean; 27 contraction tests still pass.

**Commit:** `8c3551f`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

Renamed the two-electron-integral terminology to **ERI** across the nuclear
gradient code to match adcc's established convention, and added a documentation
page for nuclear gradients.

**Terminology rename (TEI → ERI):**
- `nuclear_gradient` kwargs: `eri_contraction`, `eri_shell_chunk_size`,
  `eri_pair_chunk_size`, `eri_pair_density_storage`
- PySCF backend: `_contract_eri_with_packed_density` method and the
  `Gradient["ERI"]` component key
- Psi4 backend: `Gradient["ERI"]` component key (the Psi4 `ao_tei_deriv1`
  API call is left untouched)
- Tests and the profiling script updated to the new names

**New documentation page (`docs/gradients.rst`):**
- Usage-first structure: how to request a gradient, then a supported-methods
  table, then a "Theory and algorithm" section with the working equations
- Supported-methods table derived from the actual dispatch table (MP2,
  ADC(0/1/2/2x/3), CVS-ADC(0/1/2/2x); CVS-ADC(3) marked unavailable)
- Corrected attribution of the Lagrangian formalism (credited broadly, with
  the working equations attributed to Rehn and Dreuw)
- Added citations for Rehn and Dreuw (2019) and the CVS-ADC gradients paper
  by Brumboiu et al. (2021)

flake8 clean, 27/27 gradient contraction tests pass, full Sphinx + Doxygen
docs build succeeds with no warnings on the new page.

**Commit:** `c001cfd`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

### CI test-data download fix

The CI test jobs were failing during pytest **collection** (an `INTERNALERROR`
before any test ran), not because of the gradient changes. Root cause: the
upstream test-data server no longer serves the `0.7.2` path that this branch
pointed at, so `update_testdata.sh` failed its download/checksum step.

`master` already fixed this by bumping the test data to `0.7.4`. I brought just
those two files over (`update_testdata.sh` SOURCE bump + regenerated
`SHA256SUMS`) rather than merging master wholesale. Verified the download now
succeeds and all 47 gradient tests pass against the refreshed data.

**Commit:** `ab57394`

### Deferred: full merge of master

A full `origin/master` merge pulls in a large, unrelated refactor (the
`libadcc/` -> `libadcc_src/` move, `setup.py` -> `pyproject.toml`, and a
restructuring of `Excitation` / `ExcitedStates` / `StateView` that removes the
`mark_excitation_property` machinery our gradient code builds on). Resolving
those conflicts is its own piece of work and will be planned separately so it
gets proper review rather than being rushed into this PR.

### Still red: code style (pre-existing)

The `flake8` "Code style" check fails on 6 lint errors in files this branch
never touched (`adcc/backends/__init__.py`,
`adcc/tests/functionality_gradients_test.py`,
`adcc/tests/generators/dump_fdiff_gradient.py`). They are present on the base
branch as well. Can be cleaned up here as a courtesy or left out of scope.

---
*Progress tracked by mach6*

maxscheurer:
## Code Review

Five review agents were run on the current state of PR 215.

### Critical

**Finding 1 — UHF nuclear gradients are silently computed with the RHF PySCF gradient object.**
`adcc/backends/pyscf.py` (`correlated_gradient` and `correlated_gradient_direct`). Both paths unconditionally construct `gradient = grad.RHF(self.scfres)`. For unrestricted PySCF references this produces incorrect gradients without raising a documented error. A TODO comment at line 89 already flags that RHF/UHF handling is unresolved. The open-shell tests only validate the packed-density algebra in isolation, not the end-to-end gradient.
*Confidence: 100 (error-auditor), 95 (completeness-checker).*

**Finding 2 — No dedicated CVS/core-orbital unit tests for the direct transform.**
`adcc/tests/gradient_contraction_test.py`, `adcc/gradients/TwoParticleDensityMatrix.py`. Unit tests only populate `oooo`, `oovv`, and `ovov` blocks on non-CVS references. Core-containing blocks (`cccc`, `ococ`, `cvcv`, etc.) are exercised only via loose finite-difference end-to-end tests in `functionality_gradients_test.py`, leaving chunk-boundary and spin-subblock bugs in core handling uncovered.
*Confidence: 95 (test-reviewer).*

### Important

**Finding 3 — C++ `Tensor::export_block` dereferences an unevaluated expression pointer.**
`libadcc/TensorImpl.cc`, `libadcc/pyiface/export_Tensor.cc`. `export_block` calls `libtensor_ptr()` without first checking `needs_evaluation()` or calling `evaluate()`. Calling `tensor.export_block(...)` on an expression-backed tensor dereferences a null pointer / causes UB. The gradient path evaluates `g2_total` beforehand, so it is safe in practice, but the public API is unguarded.
*Confidence: 90 (error-auditor).*

**Finding 4 — No direct-contraction test for excited-state gradients.**
`adcc/tests/gradient_contraction_test.py`, `adcc/gradients/__init__.py`. All explicit `eri_contraction="direct"` tests pass a `LazyMp` ground-state target. `nuclear_gradient` also accepts `Excitation` objects and routes them through the same direct contraction, but no test compares `direct` vs `full_ao` for an excited state.
*Confidence: 95 (test-reviewer).*

**Finding 5 — `shell_chunk_size` boundary correctness is only tested at one value.**
`adcc/tests/gradient_contraction_test.py`, `adcc/backends/pyscf.py`. Contraction correctness tests use `shell_chunk_size=2` only. A bug in the shell-loop boundaries or in `_symmetrized_density_slice` could depend on the chunk size and would not be caught.
*Confidence: 90 (test-reviewer).*

### Suggestions

**Finding 6 — `GradientComponents.total` returns `None` when `custom` components are present.**
`adcc/gradients/__init__.py` (lines 50–56). The `total` property returns only inside the `if self.custom is None:` branch. When `custom` is a non-empty dict, the loop accumulates into `ret` but never returns it, so the property silently returns `None`. This is currently latent because no caller passes `custom`.
*Confidence: 100 (code-reviewer, error-auditor).*

**Finding 7 — HDF5 scratch-file cleanup silently swallows close/unlink failures.**
`adcc/backends/pyscf.py` (`correlated_gradient_direct` `finally` block). `contextlib.suppress(Exception)` is used for `h5file.close()` and `suppress(OSError)` for `os.unlink()`. After a successful gradient, a disk-full flush during `close()` or a permission error is silently ignored and the function returns normally.
*Confidence: 90 (error-auditor).*

**Finding 8 — `to_ao_pair_density` assumes an unvalidated spin-block structure in coefficient matrices.**
`adcc/gradients/TwoParticleDensityMatrix.py`. The function compacts each coefficient matrix by locating the first and last non-zero row and assumes those rows are contiguous and align exactly with the alpha/beta subspace. It does not validate contiguity or alignment with `mospaces.n_orbs_alpha`. A custom coefficient dict or noisy coefficients could silently extract the wrong spin sub-block.
*Confidence: 85 (error-auditor).*

**Finding 9 — Default `eri_contraction` auto-selects the new direct path, and the default is not pinned by tests.**
`adcc/gradients/__init__.py`, `adcc/tests/functionality_gradients_test.py`. `nuclear_gradient` defaults to `eri_contraction="direct"` for PySCF. Existing functionality tests call `nuclear_gradient` without specifying the strategy but only assert the final gradient against finite-difference references; they do not assert that the direct path was used. A regression that changed the default back to `full_ao` would pass those tests silently.
*Confidence: 80 (error-auditor), 90 (test-reviewer).*

**Finding 10 — Pre-allocated ndarray `out` argument for `to_ao_pair_density` is not positively tested.**
`adcc/tests/gradient_contraction_test.py`. Only the wrong-shape rejection path is tested. No test passes a pre-allocated `np.ndarray` and verifies the result matches the no-`out` path.
*Confidence: 90 (test-reviewer).*

**Finding 11 — HDF5 scratch-file cleanup on the error path is untested.**
`adcc/tests/gradient_contraction_test.py`. The success-path cleanup test does not verify that the temporary HDF5 file is removed when the gradient computation raises an exception.
*Confidence: 85 (test-reviewer).*

**Finding 12 — `export_block` empty/zero-size ranges are not tested.**
`adcc/tests/gradient_contraction_test.py`. Tests cover full-range, random sub-ranges, spin sub-blocks, and shape validation, but not `start == end` or `end < start` edge cases.
*Confidence: 85 (test-reviewer).*

**Finding 13 — Non-PySCF backend rejection of `eri_contraction="direct"` is untested.**
`adcc/tests/gradient_contraction_test.py`. `nuclear_gradient` raises `NotImplementedError` when `direct` is requested for a provider without `correlated_gradient_direct`, but this guard has no test.
*Confidence: 85 (test-reviewer).*

**Finding 14 — `to_ao_pair_density` with a coefficient-dict input is untested.**
`adcc/tests/gradient_contraction_test.py`. The method accepts either a `ReferenceState` or a `dict` of coefficient matrices, but all existing tests pass a `ReferenceState`.
*Confidence: 85 (test-reviewer).*

**Finding 15 — `ao_pair_density_from_dense` pre-allocated `out` path is untested.**
`adcc/tests/gradient_contraction_test.py`. The static helper supports an `out` argument, but only the no-`out` path is exercised.
*Confidence: 80 (test-reviewer).*

**Finding 16 — Psi4 gradient dict still contains dead keys.**
`adcc/backends/psi4.py`. `"OEI"` and `"Total"` entries are computed but never consumed by `GradientComponents`. Removing them would fully resolve the prior dead-key finding.
*Confidence: 100 (code-reviewer).*

### Simplifier findings (behavior-preserving)

17. Redundant `import adcc.backends` in `adcc/tests/gradient_contraction_test.py` (confidence 90).
18. Redundant `g2.reference_state = refstate` in `_random_tpdm` (confidence 95).
19. Redundant `g2[blk] = tensor` reassignment after `set_random()` (confidence 95).
20. Unnecessary `int()` cast around `np.count_nonzero` (confidence 90).
21. Python loop in `ao_pair_density_from_dense` can be vectorized with advanced indexing (confidence 90).
22. Dead `pair_density = None` initialization in `correlated_gradient_direct` (confidence 95).
23. Duplicate `hasattr(provider, "correlated_gradient_direct")` check in `nuclear_gradient` (confidence 90).
24. Intersection strides recomputed inside the flat loop in `TensorImpl::export_block` (confidence 85).
25. Duplicate local import of `split_spaces` in `test_export_block_spin_subblocks_match` (confidence 85).

### Strengths

- Core direct-contraction algebra is correct and well tested against the `full_ao` reference across RHF and UHF spin expansions.
- Good staged test coverage: packed density, open-shell reference, two-stage staging contract, pair-chunk-size independence, `export_block` spin sub-blocks, and HDF5 equivalence/cleanup.
- The direct path meets its memory goal: it avoids full AO rank-4 two-particle density matrices and the full derivative ERI tensor.
- New documentation page covers usage, supported methods, theory, and the memory-bounded algorithm.

---
*Reviewed by mach6*

maxscheurer:
## Review Assessment

Review comment: https://github.com/adc-connect/adcc/pull/215#issuecomment-4701610529

The independent assessor read the actual source files for every finding.

### Classifications

|---|---|---|
| 1: UHF nuc grad uses RHF gradient object | false-positive | `grad.RHF(scfres)` is only used for spin-independent `hcore_generator()`/`grad_nuc()`; UHF densities are supplied separately. Numerically correct. |
| 2: No dedicated CVS unit tests for direct transform | nitpick | `cvs-adc*` methods in `functionality_gradients_test.py` run with the PySCF default (`direct`), so core blocks are covered end-to-end vs finite-difference. |
| 3: `export_block` dereferences unevaluated expr pointer | false-positive | `export_block` accesses data only through `libtensor_ptr()`, which calls `evaluate()` first (same contract as `export_to`). |
| 4: No direct test for excited-state gradients | false-positive | `functionality_gradients_test.py` runs `adc0/1/2/2x/3` excitations with the default `direct` path against finite-difference references. |
| 5: `shell_chunk_size` boundary tested at one value | genuine (minor) | New shell chunking loop is only tested at `shell_chunk_size=2`; should mirror the `pair_chunk_size` parametrization. |
| 6: `total` returns `None` with `custom` | genuine | Confirmed: the `for c in self.custom` loop falls through without `return ret`. Real bug, currently latent. |
| 7: HDF5 cleanup swallows close/unlink failures | nitpick | Intentional cleanup-during-exception handling; broad `suppress(Exception)` on close could be narrowed but is defensible. |
| 8: `to_ao_pair_density` assumes spin-block structure | false-positive | Compaction slices the bounding box of nonzero rows and the TPDM sub-block over the same range; self-consistent. |
| 9: Default `direct` not pinned by tests | nitpick | Default is exercised implicitly by the functionality suite; an explicit pin test is worthwhile but minor. |
| 10: Pre-allocated `out` for `to_ao_pair_density` untested | nitpick | The ndarray `out=` path is exercised indirectly by the memory-storage gradient test. |
| 11: HDF5 cleanup on error path untested | nitpick | Happy-path cleanup is tested; injecting a failure to test `finally` is low-value coverage. |
| 12: `export_block` empty/zero-size ranges untested | genuine (minor) | The `total == 0` early-return is new code never hit by tests. Trivial to add a case. |
| 13: Non-PySCF rejection of `direct` untested | nitpick | Straightforward `NotImplementedError` branch; testing needs a second backend. |
| 14: Coefficient-dict input untested | nitpick | Convenience/testing API not used in production paths. |
| 15: `ao_pair_density_from_dense` `out=` path untested | nitpick | Thin, low-risk addition. |
| 16: Psi4 gradient dict has dead keys | nitpick / out-of-scope | Pre-existing code (since 2021), only touched by the TEI→ERI rename. |
| 17–25: Simplifier behavior-preserving items | nitpick | By definition stylistic; no correctness impact. |

**Counts:** 3 genuine (all minor), 15 nitpicks, 4 false positives, 0 deferred.

### Action Plan

1. **Finding 6 (genuine — fix before merge):** Add the missing `return ret` to `GradientComponents.total` in `adcc/gradients/__init__.py` so it returns the accumulated total when `custom` components are present.
2. **Finding 5 (genuine, minor):** Parametrize `shell_chunk_size` over `[1, 2, 1000]` in the packed-contraction tests, mirroring the existing `pair_chunk_size` parametrization.
3. **Finding 12 (genuine, minor):** Add a zero-size (`start == end`) case to the `export_block` validation tests.

Optional quick win:
- **Finding 1:** Replace `grad.RHF(self.scfres)` with `self.scfres.nuc_grad_method()` in `adcc/backends/pyscf.py` for robustness/clarity (resolves the existing TODO). Not a correctness fix.

Out of scope / track separately:
- **Finding 16:** Remove dead `"OEI"`/`"Total"` keys in the Psi4 gradient dict. This predates the PR.

---
*Assessment by mach6*

maxscheurer:
## Progress Update

Implemented fixes for the genuine findings from the latest mach6 review assessment:

- **Finding 6:** Added the missing `return ret` to `GradientComponents.total` in `adcc/gradients/__init__.py` so the property returns the accumulated gradient when `custom` components are present.
- **Finding 5:** Parametrized `shell_chunk_size` over `[1, 2, 1000]` in both packed-contraction correctness tests (`h2o_sto3g` closed-shell and `cn_sto3g` open-shell), mirroring the existing `pair_chunk_size` boundary coverage.
- **Finding 12:** Added zero-size range cases to `test_export_block_validates_arguments` for the new `Tensor.export_block` primitive.

Verification:

- `pytest -q adcc/tests/gradient_contraction_test.py --skip-update` — 31 passed.
- `pytest -q adcc/tests/functionality_gradients_test.py --skip-update` — 20 passed.
- `flake8 adcc/gradients/__init__.py adcc/tests/gradient_contraction_test.py` — clean.

**Commit:** `8ab2040`

---
*Progress tracked by mach6*

PR #217, opened 2026-06-12
ADC(4)

frieschneider:
Addition of ADC(4):
- matrix vector product for ADC(4) matrix
- guess zero triples (`guess_symmetry_triples`)
- testing
- `LazyMp.py`: taken from ISR(3) PR (PR #210), so that `td3` and `mp3_dm_correction` are available

ToDo:
- [x] adcman reference data for ground state (MP4 energy)
- [x] update adcman reference for ISR(2) properties 
- [x] ADC(4) `TestPolarizableEmbedding` test fail -> updated adcman reference needed


adcc/adc_pp/matrix.py
```diff
@@ -135,6 +135,30 @@ def apply(ampl):
     return AdcBlock(apply, diagonal_pphh_pphh_0(hf))
 
 
+def diagonal_ppphhh_ppphhh_0(hf):
+    res = direct_sum("-i-j-k+a+b+c->ijkabc",
+                     hf.foo.diagonal(), hf.foo.diagonal(), hf.foo.diagonal(),
+                     hf.fvv.diagonal(), hf.fvv.diagonal(), hf.fvv.diagonal())
+    return AmplitudeVector(ppphhh=res.symmetrise(3, 4, 5))
```
jonasleitner: Since we build the diagonal from orbital energy differences, both the occupied and virtual indices have to be symmetrised.

```suggestion
    return AmplitudeVector(ppphhh=res.symmetrise(0, 1, 2).symmetrise(3, 4, 5))
```

adcc/adc_pp/matrix.py
```diff
@@ -135,6 +135,30 @@ def apply(ampl):
...
+        return AmplitudeVector(ppphhh=(
+            # Prefactor of 3: the antisymmetrisation generates 36 terms with
+            # an overall factor of 1/36, so each term appears with weight 1/12,
+            # yielding effectively 3 distinct terms.
+            3 * (
+                # N^7: O^3V^4 / N^6: O^3V^3
+                + 1 * einsum("ijkabd,cd->ijkabc", ur3, hf.fvv)
+                # N^7: O^4V^3 / N^6: O^3V^3
+                + 1 * einsum("iklabc,jl->ijkabc", ur3, hf.foo)
+            ).antisymmetrise(0, 1, 2).antisymmetrise(3, 4, 5)
```
jonasleitner: This can also be implemented using `libadcc.Tensor.__mul__` (element wise multiplication) if we can somehow get access to the `ppphhh_ppphhh_0` diagonal in this function: `d3 * ur3`. This would reduce the scaling to $$N^6$$

adcc/adc_pp/matrix.py
```diff
@@ -310,6 +393,115 @@ def apply(ampl):
...
+    return AmplitudeVector(pphh=(
+        # 0th + 1st order
+        + direct_sum("ia+Jb->iJab", dinterm_ov, dinterm_Cv).symmetrise(2, 3)
+        + direct_sum("iJ+ab->iJab", diag_oC, diag_vv)
+        # 2nd order
+        + 2 * direct_sum("i+jab->ijab", dinterm_o, dinterm_ovv).symmetrise(0, 1)
+        + 2 * direct_sum("ija+b->ijab", dinterm_oov, dinterm_v).symmetrise(2, 3)
+        + direct_sum("ij+ab->ijab", dinterm_oo, dinterm_vv)
+        + 2 * direct_sum("ia+jb->ijab", dinterm_ov_2, dinterm_ov_2).symmetrise(2, 3)
+    ))
```
jonasleitner, line 415: I think we can just drop CVS support for now to simplify the function. Then we don't have to define any intermediates, but instead can simply use nested `direct_sum("...", einsum(...), einsum(...))` calls - as far as I see none of the intermediate quantities is used more than once anyway.
If I am not mistaken, it should also be possible to alternatively compute some of the contributions using the `ph_ph_2` diagonal.

adcc/adc_pp/matrix.py
```diff
@@ -554,3 +1109,478 @@ def adc3_pib(hf, mp, intermediates):
...
+    p0_2 = mp.mp2_dm_correction
+    p0_2_oo, p0_2_vv, t1_2 = p0_2.oo, p0_2.vv, p0_2.ov
+
+    t2_1 = mp.t2(b.oovv)
+    t2_2 = mp.td2(b.oovv)
+    t2_3 = mp.td3(b.oovv)
+    t3_2 = mp.tt2(b.ooovvv)
+
+    p0_3 = mp.mp3_dm_correction
+    p0_3_vv, p0_3_ov, p0_3_oo = p0_3.vv, p0_3.ov, p0_3.oo
```
jonasleitner: For all terms with any of the `p0_3` blocks, there should be a completely analogous term with the corresponding `p0_2` block. In principle we can delete the term with `p0_2` and replace `p0_3` by `p0` (the mp3 diffdm). (`p0_2` is still needed as intermediate for other terms)

adcc/adc_pp/matrix.py
```diff
@@ -554,3 +1109,478 @@ def adc3_pib(hf, mp, intermediates):
...
+            # N^4: O^2V^2 / N^4: O^2V^2
+            + 1 * einsum("ab,ij->iajb",
+                         einsum("kbla,kl->ab", hf.ovov, p0_3_oo), d_oo)
+        )
+    )
+
+
+@register_as_intermediate
+def t2sq(hf, mp, intermediates):
+    return einsum("ikac,jkbc->iajb", mp.t2oo, mp.t2oo).evaluate()
```
jonasleitner: I think t2sq is still computed on the fly in `adc3_m11`, `cvs_adc3_m11` and `adc4_m11`

adcc/tests/generators/dump_adcc.py
```diff
@@ -26,7 +26,10 @@ def dump_groundstate(ground_state: LazyMp, hdf5_file: h5py.Group,
     # MP1 data
     gs_data[f"{gs}1/df_o1v1"] = ground_state.df("o1v1").to_ndarray()
     gs_data[f"{gs}1/t_o1o1v1v1"] = ground_state.t2("o1o1v1v1").to_ndarray()
-    if not ground_state.reference_state.restricted:
+    if (
+        not ground_state.reference_state.restricted
+        and set(ground_state.reference_state.mospaces.subspaces) == {"o1", "v1"}
+    ):
```
jonasleitner:
```suggestion
    if (
        not ground_state.reference_state.restricted
        and not ground_state.has_core_occupied_space
    ):
```
same below for MP2 and the `state_ssq`

adcc/tests/generators/dump_adcc.py
```diff
@@ -47,6 +53,9 @@ def dump_groundstate(ground_state: LazyMp, hdf5_file: h5py.Group,
     # MP3 data
     if not ground_state.has_core_occupied_space:
         gs_data[f"{gs}3/energy"] = ground_state.energy_correction(3)
+    # MP4 data
+    if not ground_state.has_core_occupied_space and not only_full_mode:
+        gs_data[f"{gs}4/energy"] = ground_state.energy_correction(4)
```
jonasleitner, line 54: How do we want to deal with data that itself is not large (in this case just a number), but the calculation requires "large" data - the `tt2` amplitudes? Generate and dump the data for all test cases or only for the small ones? (Also relevant for e.g. `td3`)

adcc/tests/generators/generate_adcc_data.py
```diff
@@ -46,6 +46,12 @@ def generate_adc(test_case: testcases.TestCase, method: AdcMethod, case: str,
     key = f"{case}/{gs_density_order}"
     if f"{key}/{kind}" in hdf5_file:
         return None
+    # CVS-ADC(4) not available
+    if "cvs" in case and method.name == "adc4":
+        return None
+    # Generate adc4 reference only for small testcases
+    if test_case.only_full_mode and method.name == "adc4":
+        return None
```
jonasleitner: This should be handled before calling the `generate_adc` function, since an (unnecessary) hdf5 file might have been already created at this point. In `generate_adcman_data.py` I defined an additional `_small_cases_methods` to handle this. We can also use something different, but then please also adjust the adcman generator such that the same strategy is used in both generators.

adcc/tests/functionality_test.py
```diff
@@ -88,35 +90,38 @@ def base_test(self, system: testcases.TestCase, case: str, method: str,
...
-            # in such cases and skip the test for the exact values.
-            if system.name != "cn":
-                assert_allclose_signfix(res_tdm, ref_tdm, atol=1e-5)
-
-        # Computing the dipole moment implies a lot of cancelling in the
-        # contraction, which has quite an impact on the accuracy.
-        assert_allclose(res.state_dipole_moment[:n_ref],
-                        ref["state_dipole_moments"], atol=1e-4)
+            assert_allclose(res.state_dipole_moment[:n_ref],
+                            ref["state_dipole_moments"], atol=1e-4)
```
jonasleitner: Can't this be reverted once we fixed the adcman ADC(4) properties?

adcc/AdcMatrix.py
```diff
@@ -448,10 +448,23 @@ def construct_symmetrisation_for_blocks(self):
         else:
             def symmetrise_generic_adc_doubles(invec):
                 # doubles part is antisymmetric wrt. (i,j,a,b) <-> (i,j,b,a)
-                scratch = invec.antisymmetrise([(2, 3)])
+                # doubles part is antisymmetric wrt. (i,j,a,b) <-> (j,i,a,b)
+                scratch = invec.antisymmetrise([(2, 3)]).antisymmetrise([(0, 1)])
                 # doubles part is symmetric wrt. (i,j,a,b) <-> (j,i,b,a)
                 return scratch.symmetrise([(0, 1), (2, 3)])
```
jonasleitner, line 454: Is the `symmetrise` still needed? I think this might be unnecessary if we apply both antisymmetrisations.
-> We should check the symmetry object

Some random thought also related to the preconditioning: What about the projection of quintuples in adcc?

adcc/LazyMp.py
```diff
@@ -200,6 +309,41 @@ def mp2_diffdm(self):
         ret.reference_state = self.reference_state
         return evaluate(ret)
 
+    @cached_property
+    def mp2_diffdm(self):
```
jonasleitner:
```suggestion
    @property
    def mp2_diffdm(self):
```

adcc/tests/functionality_test.py
frieschneider: Yes, just wanted make sure that the energies are correct :)

adcc/tests/generators/dump_adcc.py
frieschneider: I would only store it for the small test cases. Since we don't intend to compute these properties for the larger test cases in the CI, I don't see much benefit in storing the corresponding reference data.

jonasleitner: Yes, I think you are right. I am not 100% sure whether we want to keep the large test cases at all in the long run, since their data takes up quite a lot of disk space...

adcc/AdcMatrix.py
frieschneider: `symmetrise` seems to be needed here. I couldn't spot any obvious difference when comparing it with `describe_symmetry`, but if I remove it, the lanczos solver no longer converges.

For triples, symmetrisation involving more than one index triple is not implemented yet (I haven't looked into that part in detail).

frieschneider:
New ToDo:
- [x] Symmetrisation of doubles/triples in `AdcMatrix.construct_symmetrisation_for_blocks`
- [x] Projection of quintuples? -> it is handled in `amplitude_vector_enforce_spin_kind`
- [x] `third_order_densitity_correction` and `m_3_plus`, `m_3_minus` implementation specific for MP ?

adcc/adc_pp/matrix.py
```diff
@@ -555,3 +1082,468 @@ def adc3_pib(hf, mp, intermediates):
...
+
+
+@register_as_intermediate
+def df3(hf, mp, intermediates):
+    return direct_sum(
+        "-i-j-k+a+b+c->ijkabc",
+        hf.foo.diagonal(), hf.foo.diagonal(), hf.foo.diagonal(),
+        hf.fvv.diagonal(), hf.fvv.diagonal(), hf.fvv.diagonal()
+    ).symmetrise(0, 1, 2).symmetrise(3, 4, 5).evaluate()
```
jonasleitner:
```suggestion
    ).symmetrise(0, 1, 2).symmetrise(3, 4, 5)
```

adcc/adc_pp/matrix.py
```diff
@@ -555,3 +1082,468 @@ def adc3_pib(hf, mp, intermediates):
...
+            # N^4: O^2V^2 / N^4: O^2V^2
+            + 1 * einsum("ab,ij->iajb",
+                         einsum("kbla,kl->ab", hf.ovov, p0_oo), d_oo)
+        )
+    )
+
+
+@register_as_intermediate
+def t2sq(hf, mp, intermediates):
+    return einsum("ikac,jkbc->iajb", mp.t2oo, mp.t2oo).evaluate()
```
jonasleitner:
```suggestion
    return einsum("ikac,jkbc->iajb", mp.t2oo, mp.t2oo)
```

adcc/tests/functionality_test.py
```diff
@@ -87,6 +89,12 @@ def base_test(self, system: testcases.TestCase, case: str, method: str,
                 #      are implemented
                 assert res.ground_state.energy_correction(3) == \
                     approx(refmp["mp3"]["energy"])
+        if res.method.level.to_int() >= 4:
+            if not res.method.is_core_valence_separated:
+                # TODO The latter check can be removed once CVS-MP3 energies
+                #      are implemented
```
jonasleitner: maybe add a `with pytest.raises(...)` so we don't forget?

adcc/tests/LazyMp_test.py
```diff
@@ -145,6 +145,20 @@ def test_mp3_energy(self, system: str, case: str, generator: str,
             == approx(refmp["mp3"]["energy"])
         )
 
+    # MP4 energy not implemented for CVS
+    @pytest.mark.parametrize("system,case", non_cvs_small_cases)
+    # no reference data yet for adcman
+    @pytest.mark.parametrize("generator", ["adcc"])
```
jonasleitner: The adcman data should be available now :)

adcc/GroundState.py
```diff
@@ -235,6 +240,43 @@ def t2eri(self, space: str, contraction: str) -> libadcc.Tensor:
         contraction_str, eri_block = expressions[key]
         return einsum(contraction_str, self.t2oo, hf.eri(eri_block))
 
+    @property
+    def m_3_plus(self) -> libadcc.Tensor:
+        """
+        Third order contribution to the ov block of the N+1 part of the
+        dynamic self-energy.
+        """
+        raise NotImplementedError(
```
jonasleitner: The `m_3_plus` and `m_3_minus` intermediates should also be implemented on the `GroundState` class for now I think. Although I am not 100% sure about that.

adcc/LazyMp.py
```diff
@@ -234,3 +319,42 @@ def mp2_density(self):
...
+    def third_order_dm_correction(self, apply_cvs: bool = False
+                                  ) -> OneParticleDensity:
+        """
+        Return the third-order MP contribution to the ground state
+        difference density in the MO basis.
+        """
+        if apply_cvs:
+            raise NotImplementedError(
+                "CVS-MP3 difference density not implemented yet"
+            )
```
jonasleitner:
```suggestion
        if self.has_core_occupied_space:
            raise NotImplementedError(
                "CVS-MP3 difference density not implemented yet"
            )
        assert not apply_cvs
```

adcc/LazyMp.py
```diff
@@ -234,3 +319,42 @@ def mp2_density(self):
     @property
     def mp2_dipole_moment(self):
         return self.dipole_moment(level=2)
+
+    @property
+    def mp3_diffdm(self):
+        """
+        Return the MP3 difference density in the MO basis.
+        """
+        return self.diffdm(3)
```
jonasleitner: I would prefer to not add more `mpn_diffdm` (or `mpn_dipole_moment`) methods, but instead the `diffdm` and `dipole_moment` methods should be used. I would only keep the `mp2` methods for backwards compatibility


```suggestion
```

libadcc_src/TensorImpl.cc
```diff
@@ -1235,6 +1241,12 @@ std::shared_ptr<Tensor> TensorImpl<N>::antisymmetrise(
...
                                std::get<2>(parsed), lthis);
+      } else if (permutations.size() == 2) {
+        const auto parsed =
+              parse_permutation<3, 2>(m_axes, strip_safe<N>(label), permutations);
+        return 1. / 6. *
+               lt::expr::asymm(std::get<0>(parsed), std::get<1>(parsed),
+                               std::get<2>(parsed), lthis);
       } else {
         throw runtime_error(
               "Antisymmetrisation not implemented for more than one index triple.");
```
jonasleitner:
```suggestion
              "Antisymmetrisation not implemented for more than two index triple.");
```

adcc/tests/data/update_testdata.sh
```diff
@@ -4,7 +4,7 @@
 trap "exit 1" TERM
 export SCRIPT_PID=$$
 
-SOURCE="https://wwwagdreuw.iwr.uni-heidelberg.de/adcc_test_data/0.7.5/"
+SOURCE="https://wwwagdreuw.iwr.uni-heidelberg.de/adcc_test_data/adc4/"
```
jonasleitner: Just so we don't forget: we have to pick an actual version number before merging

adcc/GroundState.py
frieschneider: Are the expressions identical, or are the equations specific for an MP partitioning? Maybe I missunderstood earlier :D

jonasleitner (approved):

PR #223, opened 2026-06-15
LazyMp refactoring for different Hamiltonian partitionings

jonasleitner:
This PR introduces a new `GroundState` base class for ground states to be able to incorporate different Hamiltonian partitionings in the future (RE, REMP, BWs). `GroundState` holds implementations for densities and properties, since the partitionings I have in mind share common algebraic expressions for the densities. Only t-amplitudes and energies differ and thus have to be implemented on the child classes.
Additionally, I removed the `cvs_p0` intermediate and instead introduced an optional argument `apply_cvs` to the ground state densities and properties.
Moreover, I added tests for the CVS-MP2 density and the 2-particle MP densities.

TODO:
- [x] Change the test data version to an appropriate number before merging


frieschneider (commented):
Looks good 👍 I'll approve it once the testdata version number is updated.

frieschneider (approved):

PR #225, opened 2026-06-16
Speed up restricted AO-pair TPDM transform

### LINKED ISSUE #220: Speed up restricted (RHF) gradient: build only unique TPDM spin-block transforms with prefactors
## Summary

For a **restricted (RHF)** reference, the direct AO-pair TPDM transform in `TwoParticleDensityMatrix.to_ao_pair_density` computes redundant spin cases. We can build only the unique transformations and apply spin-multiplicity prefactors, halving the dominant einsum work, while leaving the unrestricted path untouched.

## Background

The hot path `to_ao_pair_density` calls `_add_direct_pair_transform` **8 times per nonzero block per pair-chunk**:

- direct ordering: `aaaa`, `bbbb`, `abab`, `baba`
- exchange ordering: `aaaa`, `bbbb`, `abba`, `baab`

In the gradient profile (benzene 6-31g\*, adc2, direct), `_add_direct_pair_transform` was ~66 s of ~140 s wall, with the einsum core ~50 s — so this is the single largest reducible cost.

## What is redundant for RHF

For a restricted reference the alpha and beta MO coefficients coincide and spin symmetry makes the spin sub-blocks collapse pairwise. Verified **numerically** on real amplitude-relaxed densities (`export_block` sub-blocks, H2O/STO-3G, both MP2 and every nonzero ADC(2) block):

| pairing | equivalence | verified |
|---|---|---|
| same-spin | `aaaa == bbbb` | EQUAL (all nonzero blocks) |
| opposite-spin direct | `abab == baba` | EQUAL |
| opposite-spin exchange | `abba == baab` | EQUAL |

So for RHF we only need **4 contractions instead of 8**:

- direct: `aaaa` (×2), `abab` (×2)
- exchange: `aaaa` (×2), `abba` (×2)

each scaled by a prefactor of 2 (the spin multiplicity of the merged pair). This reproduces the identical packed AO-pair density at ~half the einsum cost.

## Why the guard is required

The equivalence genuinely breaks for unrestricted references — for UHF the `aaaa` and `bbbb` sub-blocks can even have **different shapes** (e.g. `(7,7,3,3)` vs `(6,6,4,4)`) because the alpha/beta occupations differ. Detection is trivial and already available: `ReferenceState.restricted` is `True` for RHF and `False` for UHF.

- `restricted` -> 4-contraction fast path with ×2 prefactors
- unrestricted -> existing 8-case path, unchanged

- [ ] Add a restricted fast path in `to_ao_pair_density` gated on `reference_state.restricted` that builds only the unique spin cases (`aaaa`, `abab` direct; `aaaa`, `abba` exchange) with ×2 prefactors.
- [ ] Keep the existing 8-case path for unrestricted references.
- [ ] The prefactor folds into spin multiplicity only — the off-diagonal ket-pair swap (`offdiag`/`swapped`) accumulation and the per-ordering `sign` must still be applied exactly as today.
- [ ] Add a correctness test asserting the restricted fast path matches the unrestricted-style (8-case) path bit-for-bit on an RHF system.
- [ ] All existing gradient tests stay green (especially the UHF/open-shell contraction tests).
- [ ] (Nice to have) Re-profile to quantify the wall/CPU saving on a representative case.

## Notes / dependencies

- This touches the same `to_ao_pair_density` hot path that the master-refactor integration (#219) does **not** modify, so it is independent of that work.
- **To be addressed once `gradient_contraction` is merged into `gradients`** (i.e. after PR #215 lands), to avoid churn on an in-flight branch.
- This is preferable to the previously rejected tensordot rewrite: it removes work rather than reshuffling the same work, with no executor/precision trade-offs.

## Out of scope

- Changes to the unrestricted contraction algorithm or the integral-side contraction.

maxscheurer:
Closes #220

Adds a restricted-reference fast path for the direct AO-pair TPDM transform by planning to contract only unique RHF spin cases with spin-multiplicity prefactors while preserving the unrestricted path.

### DIFF
--- a/adcc/gradients/TwoParticleDensityMatrix.py
+++ b/adcc/gradients/TwoParticleDensityMatrix.py
@@ -337,6 +337,12 @@ class TwoParticleDensityMatrix:
         - ``g2_ao_1``: ``aaaa``, ``bbbb``, ``abab``, ``baba``
         - ``g2_ao_2``: ``aaaa``, ``bbbb``, ``abba``, ``baab``
 
+        For restricted references, equivalent spin cases are merged and
+        accumulated with their spin multiplicity (``aaaa``/``abab`` for
+        ``g2_ao_1`` and ``aaaa``/``abba`` for ``g2_ao_2``, each with a factor
+        of two).  Raw coefficient dictionaries do not carry a trustworthy
+        restricted/unrestricted flag and therefore use the full spin expansion.
+
         The returned/filled array has shape ``(nao, nao, nao * (nao + 1) // 2)``
         and contains ``D[p,r,q,s] = g2_ao_1[p,q,r,s] - g2_ao_2[p,q,s,r]`` with
         the ``q,s`` ket pair packed in lower-triangular order (see
@@ -356,6 +362,16 @@ class TwoParticleDensityMatrix:
         if not len(self.blocks_nonzero):
             raise ValueError("At least one non-zero block is needed to "
                              "transform the TwoParticleDensityMatrix.")
+        restricted_refstate = None
+        if isinstance(refstate_or_coefficients, libadcc.ReferenceState):
+            restricted_refstate = refstate_or_coefficients
+        elif (refstate_or_coefficients is None
+              and hasattr(self, "reference_state")):
+            restricted_refstate = self.reference_state
+        use_restricted_fast_path = bool(
+            getattr(restricted_refstate, "restricted", False)
+        )
+
         coeff_map = self._ao_coefficient_map(refstate_or_coefficients)
         nao = next(iter(coeff_map.values())).shape[1]
         npair = nao * (nao + 1) // 2
@@ -387,18 +403,32 @@ class TwoParticleDensityMatrix:
                     lo, hi = int(nz[0]), int(nz[-1]) + 1
                     compact[(sp, spin)] = (c[lo:hi], lo, hi)
 
-        direct_spin_cases = [
-            ("a", "a", "a", "a"),
-            ("b", "b", "b", "b"),
-            ("a", "b", "a", "b"),
-            ("b", "a", "b", "a"),
-        ]
-        exchange_spin_cases = [
-            ("a", "a", "a", "a"),
-            ("b", "b", "b", "b"),
-            ("a", "b", "b", "a"),
-            ("b", "a", "a", "b"),
-        ]
+        if use_restricted_fast_path:
+            direct_spin_cases = [
+                (("a", "a", "a", "a"), 2.0),
+                (("a", "b", "a", "b"), 2.0),
+            ]
+            exchange_spin_cases = [
+                (("a", "a", "a", "a"), 2.0),
+                (("a", "b", "b", "a"), 2.0),
+            ]
+        else:
+            direct_spin_cases = [
+                (("a", "a", "a", "a"), 1.0),
+                (("b", "b", "b", "b"), 1.0),
+                (("a", "b", "a", "b"), 1.0),
+                (("b", "a", "b", "a"), 1.0),
+            ]
+            exchange_spin_cases = [
+                (("a", "a", "a", "a"), 1.0),
+                (("b", "b", "b", "b"), 1.0),
+                (("a", "b", "b", "a"), 1.0),
+                (("b", "a", "a", "b"), 1.0),
+            ]
+        required_spin_cases = {
+            spins for spin_cases in (direct_spin_cases, exchange_spin_cases)
+            for spins, _ in spin_cases
+        }
 
         # Pre-extract every required rank-4 spin sub-block exactly once (each is
         # only the non-zero-row portion of a block, i.e. a small fraction of the
@@ -410,7 +440,7 @@ class TwoParticleDensityMatrix:
         for block in self.blocks_nonzero:
             spaces = split_spaces(block)
             ten_obj = self.block(block)
-            for spins in set(direct_spin_cases) | set(exchange_spin_cases):
+            for spins in required_spin_cases:
                 ranges = [compact[(sp, spin)] for sp, spin in zip(spaces, spins)]
                 coeff_sets[(block, spins)] = [r[0] for r in ranges]
                 if any(r[0] is None for r in ranges):
@@ -440,18 +470,20 @@ class TwoParticleDensityMatrix:
                 for spin_cases, sign, exchange in (
                         (direct_spin_cases, +1.0, False),
                         (exchange_spin_cases, -1.0, True)):
-                    for spins in spin_cases:
+                    for spins, prefactor in spin_cases:
                         sub = subblocks[(block, spins)]
                         if sub is None:
                             continue
                         coeffs = coeff_sets[(block, spins)]
+                        scaled_sign = sign * prefactor
                         self._add_direct_pair_transform(
-                            chunk, sub, *coeffs, qidx, sidx, sign, exchange
+                            chunk, sub, *coeffs, qidx, sidx, scaled_sign,
+                            exchange
                         )
                         if any_offdiag:
                             self._add_direct_pair_transform(
-                                swapped, sub, *coeffs, qswap, sswap, sign,
-                                exchange
+                                swapped, sub, *coeffs, qswap, sswap,
+                                scaled_sign, exchange
                             )
 
             if any_offdiag:
--- a/adcc/tests/gradient_contraction_test.py
+++ b/adcc/tests/gradient_contraction_test.py
@@ -195,6 +195,38 @@ def test_direct_mp2_gradient_matches_full_ao_fallback():
                     full.components.two_electron, atol=1e-10)
 
 
+def test_restricted_fast_path_matches_eight_case_transform():
+    """RHF fast path must reproduce the legacy eight-spin-case transform."""
+    hf = cached_backend_hf("pyscf", "h2o_sto3g", conv_tol=1e-11)
+    refstate = adcc.ReferenceState(hf)
+    assert refstate.restricted
+    mp = adcc.LazyMp(refstate)
+    gradient = adcc.nuclear_gradient(mp, eri_contraction="full_ao")
+    g2 = gradient.g2
+
+    # A raw coefficient dict has no restricted flag and therefore exercises the
+    # legacy eight-case spin expansion with the same fused contraction order.
+    coeffs = g2._ao_coefficient_map(refstate)
+    eight_case = g2.to_ao_pair_density(coeffs, pair_chunk_size=5)
+    fast_path = g2.to_ao_pair_density(refstate, pair_chunk_size=5)
+
+    assert_allclose(fast_path, eight_case, atol=1e-12)
+
+
+def test_restricted_coefficient_dict_uses_full_spin_expansion():
+    """Coefficient dictionaries must not enable the RHF-only shortcut."""
+    hf = cached_backend_hf("pyscf", "h2o_sto3g", conv_tol=1e-11)
+    refstate = adcc.ReferenceState(hf)
+    assert refstate.restricted
+
+    g2 = _random_tpdm(refstate)
+    coeffs = g2._ao_coefficient_map(refstate)
+    reference = _two_stage_packed_density(g2, refstate)
+    produced = g2.to_ao_pair_density(coeffs, pair_chunk_size=3)
+
+    assert_allclose(produced, reference, atol=1e-10)
+
+
 # ---------------------------------------------------------------------------
 # Guardrail tests (milestone m0)
 #
@@ -241,7 +273,10 @@ def test_packed_density_matches_two_stage_staging_contract(system):
 
     g2 = _random_tpdm(refstate)
     reference = _two_stage_packed_density(g2, refstate)
-    produced = g2.to_ao_pair_density(refstate, pair_chunk_size=3)
+    coeffs_or_refstate = (
+        g2._ao_coefficient_map(refstate) if refstate.restricted else refstate
+    )
+    produced = g2.to_ao_pair_density(coeffs_or_refstate, pair_chunk_size=3)
 
     assert_allclose(produced, reference, atol=1e-10)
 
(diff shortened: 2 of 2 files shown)


maxscheurer:
## Implementation Plan

### Analysis

`TwoParticleDensityMatrix.to_ao_pair_density` is the hot path for direct gradient ERI contraction. It currently loops over four direct spin cases and four exchange spin cases for every nonzero TPDM block and AO-pair chunk:

- direct: `aaaa`, `bbbb`, `abab`, `baba` with positive sign
- exchange: `aaaa`, `bbbb`, `abba`, `baab` with negative sign

For a restricted reference, the alpha and beta orbital coefficients coincide and the physical TPDM spin sub-blocks collapse pairwise. The transform can therefore contract only the unique cases and multiply by the spin multiplicity:

- direct: `aaaa` times 2 and `abab` times 2
- exchange: `aaaa` times 2 and `abba` times 2

The unrestricted path must remain the existing eight-case behavior because alpha and beta spaces can have different shapes and values. The downstream PySCF gradient code already passes a real `ReferenceState`, so the fast path can trigger in the production direct-gradient path without backend changes.

No project convention files were found (`CONTRIBUTING.md`, `DEVELOPMENT.md`, `CLAUDE.md`, `AGENTS.md`, `.dreb/CONTEXT.md`). The PR target is `gradients`; the old `gradient_contraction` worktree risk is intentionally ignored per request.

### Deliverables

1. Add a restricted-reference schedule in `to_ao_pair_density` that uses only the unique RHF spin cases with a prefactor of 2.
2. Keep the unrestricted schedule byte-for-byte equivalent in behavior to the current eight-case loop.
3. Gate the fast path only when a real `ReferenceState` is available and reports `restricted`:
   - use the passed `ReferenceState` when supplied,
   - use `self.reference_state` when the argument is `None`,
   - fall back to the eight-case path for raw coefficient dictionaries because they carry no trustworthy restricted/unrestricted flag.
4. Apply the multiplicity by multiplying the existing per-ordering sign (`+1` for direct, `-1` for exchange), or by an equivalent explicit prefactor, without changing `_add_direct_pair_transform` semantics unnecessarily.
5. Preserve the current off-diagonal ket-pair swap accumulation exactly: the `offdiag`/`swapped` handling must receive the same spin multiplicity and ordering sign as the primary chunk.
6. Optionally narrow the `export_block` pre-extraction set to the selected schedule so the restricted path saves extraction and cache memory as well as einsum work.

### Files to modify

- `adcc/gradients/TwoParticleDensityMatrix.py`
  - Resolve whether the current call has a trustworthy restricted `ReferenceState`.
  - Replace the fixed direct/exchange spin-case lists with schedules that include a multiplicity/prefactor.
  - Use the restricted schedule only for RHF; use the current eight-case schedule otherwise.
  - Pre-extract only spin blocks referenced by the selected schedules.
  - Keep output shape, pair chunking, HDF5 `out` behavior, and error validation unchanged.

- `adcc/tests/gradient_contraction_test.py`
  - Add focused RHF coverage for the restricted fast path.
  - Adjust any artificial random-RHF TPDM tests if needed, because a fully random TPDM is not guaranteed to satisfy the physical RHF spin-block identities that the optimization relies on.

### Testing approach

Add or update tests in `adcc/tests/gradient_contraction_test.py`:

1. **Restricted fast path versus eight-case oracle**
   - Use `h2o_sto3g` with `adcc.ReferenceState(hf)` and assert `refstate.restricted`.
   - Use a real restricted-gradient TPDM, for example from `adcc.nuclear_gradient(mp, eri_contraction="full_ao").g2`, or construct a test TPDM whose relevant spin sub-blocks are explicitly symmetric.
   - Compare `g2.to_ao_pair_density(refstate, pair_chunk_size=...)` to an eight-case oracle.
   - Prefer an oracle that preserves the old fused operation order when asserting exact equality; otherwise use the existing independent two-stage oracle with a tight tolerance.
   - Parametrize at least one small chunk size to exercise off-diagonal `swapped` accumulation.

2. **Coefficient-dictionary fallback**
   - Pass the coefficient map/dictionary form for an RHF reference and verify the result follows the full eight-case behavior rather than the restricted fast path.
   - This protects callers that provide coefficients without a `restricted` flag.

3. **Existing UHF guardrails must remain green**
   - `test_open_shell_pair_density_matches_dense_ao_transform`
   - `test_open_shell_packed_contraction_matches_dense_reference`
   - `test_packed_density_independent_of_pair_chunk_size`
   - These cover the `cn_sto3g` unrestricted path where alpha and beta shapes/values can differ.

4. **Existing direct-gradient coverage must remain green**
   - `test_direct_mp2_pair_density_matches_dense_ao_transform`
   - `test_direct_mp2_gradient_matches_full_ao_fallback`
   - `test_direct_gradient_hdf5_matches_memory`
   - These verify the production path and the HDF5 output storage path still behave correctly.

Suggested focused command:

```bash
python -m pytest adcc/tests/gradient_contraction_test.py -v
```

### Acceptance criteria

- RHF direct-gradient AO-pair density uses four contractions per block/chunk instead of eight.
- UHF and raw-coefficient-dictionary calls retain the existing eight-case behavior.
- Direct and exchange signs remain unchanged apart from the intended RHF multiplicity.
- Off-diagonal packed ket-pair accumulation remains correct for restricted and unrestricted references.
- Restricted fast-path output matches an unrestricted-style eight-case oracle on an RHF physical/symmetric TPDM.
- Existing open-shell gradient contraction tests stay green.
- Existing HDF5 pair-density storage tests stay green.

### Risks and open questions

- **Artificial random RHF TPDMs:** existing helper-generated random TPDMs may not obey the physical RHF spin-block identities. Tests should avoid treating arbitrary random RHF tensors as valid inputs for the restricted optimization unless they are explicitly symmetrized.
- **Exact equality wording:** exact bitwise equality is realistic only when the oracle uses the same fused operation order and the paired spin sub-blocks are truly identical. A two-stage dense oracle is still valuable but may require a tight numerical tolerance because it changes contraction order.
- **Restricted detection:** enabling the fast path for coefficient dictionaries would be unsafe; keep the fallback conservative unless the API gains an explicit restricted flag.
- **Performance validation:** profiling is nice to have after correctness lands, but not required for the implementation PR.

---
*Plan created by mach6*

maxscheurer:
## Progress Update

Implemented the restricted-reference AO-pair TPDM fast path:

- Added RHF spin-case schedules that contract only the unique direct and exchange spin blocks with spin-multiplicity prefactors.
- Kept the full eight-case path for unrestricted references and raw coefficient dictionaries.
- Pre-extracts only the spin subblocks required by the selected schedule.
- Added RHF fast-path and coefficient-dictionary fallback tests.
- Verified the gradient contraction test module locally: `python -m pytest adcc/tests/gradient_contraction_test.py -v` passed with 33 tests.

**Commit:** `d89aaea`

---
*Progress tracked by mach6*

maxscheurer:
## Code Review

### Critical
No critical findings.

### Important
1. **No chunking-independence coverage for RHF fast path**  
   File: `adcc/tests/gradient_contraction_test.py` (`test_packed_density_independent_of_pair_chunk_size`, `test_restricted_fast_path_matches_eight_case_transform`)  
   The new RHF fast path applies `scaled_sign = sign * prefactor` in both the regular chunk accumulation and the `swapped` off-diagonal accumulation, but it is only exercised at a single `pair_chunk_size=5`. A chunk-boundary bug that only appears for certain chunk sizes in the RHF branch would not be caught. Add a parametrized RHF test that compares `g2.to_ao_pair_density(refstate, pair_chunk_size=...)` against `pair_chunk_size=None` (or the coefficient-dict legacy result) for several chunk sizes.

2. **Modified staging-contract test no longer exercises RHF fast path with generic TPDM data**  
   File: `adcc/tests/gradient_contraction_test.py` (`test_packed_density_matches_two_stage_staging_contract`)  
   The modified test now passes `g2._ao_coefficient_map(refstate)` for RHF, so the fast-path branch is not compared against the independent `_two_stage_packed_density` reference for random-TPDM data. The new fast-path comparison test only uses a physical MP2 density. Add a test that calls `g2.to_ao_pair_density(refstate, pair_chunk_size=3)` on a `_random_tpdm(refstate)` and compares to `_two_stage_packed_density(g2, refstate)`.

### Suggestions
3. **Coefficient-dict guard test is too weak to detect accidental fast-path use**  
   File: `adcc/tests/gradient_contraction_test.py` (`test_restricted_coefficient_dict_uses_full_spin_expansion`)  
   For an RHF reference the alpha and beta spatial coefficients are identical, so the full 8-case expansion and the restricted fast path produce mathematically identical results. The assertion would still pass if the implementation incorrectly took the fast path for a raw coefficient dict. Strengthen the test by constructing a synthetic coefficient dict with deliberately different alpha/beta spatial matrices and verifying the result matches `_two_stage_packed_density` built from that same dict.

4. **HDF5 path is not exercised with more than one AO-pair chunk**  
   File: `adcc/tests/gradient_contraction_test.py` (`test_direct_gradient_hdf5_matches_memory`)  
   For `h2o_sto3g`, `npair = 28`, so the default HDF5 `pair_chunk_size = min(npair, 256)` equals the full output. The test therefore never executes the `out[:, :, start:stop] += chunk` read-modify-write across chunk boundaries on an HDF5 dataset. Add a test that passes an explicit small `eri_pair_chunk_size` (e.g., 3 or 5) together with `eri_pair_density_storage="hdf5"` and compares to the in-memory result.

5. **Optional profiling deliverable is missing**  
   Source: issue 220 acceptance criterion 6 (nice-to-have)  
   The PR does not include a profiling script or timing data quantifying the wall/CPU saving. This is explicitly labeled as nice-to-have, so it is not required for merge.

### Strengths
- The restricted fast path is correctly gated on `ReferenceState.restricted`; raw coefficient dicts fall back to the full 8-case path.
- Spin-case schedules and prefactors are applied consistently for both direct/exchange orderings and off-diagonal `swapped` accumulation.
- `required_spin_cases` is derived directly from the selected schedule, narrowing sub-block extraction in the restricted path.
- Full `gradient_contraction_test.py` suite passes locally (33 tests) and CI is green.
- The UHF/open-shell guardrail tests are unchanged and continue to exercise the 8-case path.

**Agents run:** code-reviewer, error-auditor, test-reviewer, completeness-checker

---
*Reviewed by mach6*

maxscheurer:
## Review Assessment

Review comment: https://github.com/adc-connect/adcc/pull/225#issuecomment-4720494104

### Classifications

|---|---|---|
| Finding 1: No chunking-independence coverage for RHF fast path | nitpick | The chunk loop is identical for fast and 8-case paths; only the spin-case lists and prefactors differ, which are chunk-independent. The fast path already runs at `pair_chunk_size=5` over h2o (npair=28, 6 chunks incl. an uneven last chunk and off-diagonal swapped handling), and chunking is parametrized `[1,2,3,1000]` on UHF. |
| Finding 2: Staging test no longer exercises RHF fast path with generic TPDM | genuine (minor) | The diff changes `test_packed_density_matches_two_stage_staging_contract` to pass `_ao_coefficient_map(refstate)` for restricted, so the fast path is no longer compared against the independent `_two_stage_packed_density` on random data. The fast path's only independent random-data check is now gone; it's validated against a dense reference only on physical MP2 data. |
| Finding 3: Coefficient-dict guard test too weak | genuine (minor) | For a restricted reference, `set_random()` produces `aaaa == bbbb` sub-blocks and equal alpha/beta compacted coefficients, so the fast path and 8-case path are numerically identical. `test_restricted_coefficient_dict_uses_full_spin_expansion` therefore cannot detect accidental fast-path use; its stated purpose is unverifiable as written. |
| Finding 4: HDF5 path not exercised with more than one AO-pair chunk | false positive | `test_direct_gradient_hdf5_matches_memory` explicitly passes `eri_pair_chunk_size=3`; with npair=28 that is 10 chunks, so the HDF5 read-modify-write across chunk boundaries is exercised and compared to the memory path. |
| Finding 5: Optional profiling deliverable missing | nitpick | Profiling is explicitly marked nice-to-have in issue 220, and a profiling script already exists from PR 215. Not blocking. |

### Action Plan

Two genuine but non-blocking test-quality improvements:

1. **Strengthen the coefficient-dict guard test.** Modify `test_restricted_coefficient_dict_uses_full_spin_expansion` so it feeds `to_ao_pair_density` a coefficient dict with deliberately different alpha/beta spatial matrices together with a TPDM whose `aaaa != bbbb`, and assert the result matches the full 8-case `_two_stage_packed_density` reference. This makes the test able to catch accidental fast-path activation for raw dicts.

2. **Restore independent fast-path coverage on random data.** Parametrize `test_packed_density_matches_two_stage_staging_contract` (or add a sibling test) so the restricted case runs both via the coefficient dict (8-case) and via `refstate` (fast path), each compared against `_two_stage_packed_density`. This re-pins the fast path against an independent reference on generic random data rather than only on physical MP2 data.

Neither finding blocks merge — CI passes and the core transform is correct. No tracking issue is required for the deferred nice-to-have profiling item.

---
*Assessment by mach6*

maxscheurer:
## Progress Update

Addressed review findings 2 and 3 from the mach6 assessment:

- **Finding 2:** Re-parametrized `test_packed_density_matches_two_stage_staging_contract` over `(system, use_coeff_dict)` so the restricted reference is now exercised both via a raw coefficient dict (full eight-spin-case path) and via the `ReferenceState` (restricted fast path), each compared against the independent `_two_stage_packed_density` reference on generic random TPDM data.
- **Finding 3:** Strengthened `test_restricted_coefficient_dict_uses_full_spin_expansion` by feeding it a coefficient dict with deliberately different alpha/beta spatial matrices and a TPDM with distinct `aaaa`/`bbbb` spin sub-blocks. This makes the test capable of detecting accidental fast-path activation for raw dicts.
- Added helpers `_random_tpdm_with_distinct_spin_blocks` and `_perturbed_restricted_coefficient_map`, and extended `_two_stage_packed_density` with an optional `coeff_map` argument.

Local verification: `python -m pytest adcc/tests/gradient_contraction_test.py -v` passed with 34 tests.

**Commit:** `81b6b20`

---
*Progress tracked by mach6*

PR #226, opened 2026-06-17
Add geometry optimization scanner for gradients

### LINKED ISSUE #221: Geometry optimization interface: PySCF-style gradient scanner for use with geomeTRIC
## Summary

Provide a geometry-optimization interface for ADC/MP excited- and ground-state gradients. The goal is a PySCF-style **scanner** object — callable with a new geometry, returning `(energy, gradient)` — that can be handed directly to an external optimizer such as [geomeTRIC](https://github.com/leeping/geomeTRIC) (and ideally PySCF's own `optimize` driver, which speaks the same scanner protocol).

## Motivation

We now have analytic nuclear gradients (#214 / #215). The natural next capability is relaxed-geometry and excited-state-minimum searches. External optimizers (geomeTRIC, berny, PySCF `geomopt`) all drive a simple contract: given updated nuclear coordinates, return the energy and its gradient. We just need to expose adcc gradients through that contract.

## The core difficulty: two libraries, one consistent calculation

Unlike a pure-PySCF gradient scanner, an adcc gradient at each geometry spans **two libraries**:

1. **PySCF** runs the SCF (the `mol` + `scf.RHF/UHF` object).
2. **adcc** runs the correlated part (`LazyMp` / `run_adc`) and the gradient (`nuclear_gradient`).

For the optimization to be well-defined, **both libraries' settings must stay consistent and fixed across every geometry step**, otherwise the energy/gradient surface is discontinuous and the optimizer fails or converges to garbage. Concretely the scanner must re-create, identically at each new geometry:

- **PySCF / SCF side**: basis set, charge, spin/multiplicity, reference type (RHF vs UHF/ROHF), SCF convergence (`conv_tol`, `conv_tol_grad`, `max_cycle`), density fitting on/off, point-group symmetry handling, integral thresholds, and — importantly — **MO orbital continuity / SCF guess** between steps (feed the previous step's converged MOs as the initial guess to avoid root flipping and speed convergence).
- **adcc side**: method (e.g. `adc2`, `cvs-adc2`), `core_orbitals` / frozen-core selection, which **state** is being optimized (state index / `kind` / `n_states` or `n_singlets`), property method, ADC `conv_tol`. For an excited-state optimization the scanner must reliably **track the same physical state** across geometries (state following / overlap-based root tracking), not just "state index 2", which can swap character along the path.

Today this consistency is implicit and manual: the finite-difference generator
(`adcc/tests/generators/dump_fdiff_gradient.py`) re-runs
`adcc.backends.run_hf('pyscf', geom, basis, ...)` then `run_adc` / `LazyMp` then
reads the energy for each displaced geometry. A scanner needs to encapsulate
exactly this loop, with settings frozen and state tracked, behind a clean
callable.

## Proposed design (sketch)

A scanner factory that captures all settings once and is then called per geometry:

```python
scanner = adcc.nuclear_gradient_scanner(
    # PySCF SCF spec
    mol_template=mol, scf_type="RHF", scf_conv_tol=1e-11,
    # adcc spec
    method="adc2", state_index=0, n_singlets=5, core_orbitals=None,
    # gradient spec (existing nuclear_gradient kwargs)
    eri_contraction="direct",
)

# geomeTRIC / PySCF geomopt contract:
energy, gradient = scanner(new_coords)   # gradient shape (natoms, 3)
```

Internally each call would:
1. Rebuild the PySCF `mol` at `new_coords` keeping basis/charge/spin fixed.
2. Run SCF, seeded from the previous step's MOs for continuity.
3. Build `LazyMp` / run ADC with the frozen adcc settings.
4. Select/track the target state (root following).
5. Call `nuclear_gradient(...)` and return `(energy, gradient.total)`.

We should target the interface geomeTRIC expects (its custom-engine API) and,
where it lines up, PySCF's `as_scanner` / `geomopt` protocol so both drivers work.

- [ ] A scanner/engine object exposing a per-geometry `(energy, gradient)` callable.
- [ ] Works as a geomeTRIC custom engine end-to-end (a small optimization, e.g. H2O excited-state or ground-state minimum, converges).
- [ ] All PySCF SCF settings and all adcc method settings are captured once and held fixed across geometries.
- [ ] SCF MO continuity between steps (previous MOs as guess).
- [ ] Excited-state **root tracking** so the same physical state is followed across geometries (not a fixed index).
- [ ] Handles both ground-state (MP2 via `LazyMp`) and excited-state (`Excitation`) optimizations.
- [ ] Example / docs section showing a geomeTRIC-driven optimization.
- [ ] Tests: at minimum a short optimization that reaches a known stationary point; gradient at the optimized geometry ~ 0.

## Notes / dependencies

- **Depends on the gradients branch landing**: to be addressed **after `gradient_contraction` merges into `gradients`** (PR #215) and ideally after the master integration (#219), since the scanner builds directly on `adcc.nuclear_gradient` and the backend/SCF entry points.
- Initial target backend is **PySCF** (consistent with the `direct` gradient path being PySCF-only). Psi4 can follow if there is demand.
- geomeTRIC is an **optional** dependency — must not become a hard requirement (same policy as other optional backends).

## Open questions

- Reuse PySCF's `as_scanner` machinery for the SCF half, or manage the SCF object lifecycle ourselves?
- Best mechanism for robust excited-state following (vector overlap vs. character/property-based)?
- How much of the consistency contract can be enforced programmatically vs. documented as user responsibility?

maxscheurer:
Closes #221

Adds a planned implementation path for a PySCF/adcc nuclear-gradient scanner for geometry optimization, including geomeTRIC integration, state following, and MECP support.

### DIFF
--- a/adcc/__init__.py
+++ b/adcc/__init__.py
@@ -47,7 +47,9 @@ from .OneParticleDensity import OneParticleDensity
 from .TwoParticleOperator import TwoParticleOperator
 from .TwoParticleDensity import TwoParticleDensity
 from .opt_einsum_integration import register_with_opt_einsum
-from .gradients import nuclear_gradient
+from .gradients import (ExcitedStateTarget, GroundStateTarget,
+                        NuclearGradientScanner, nuclear_gradient,
+                        nuclear_gradient_scanner)
 
 # This has to be the last set of import
 from .guess import (guess_symmetries, guess_zero, guesses_any, guesses_singlet,
@@ -69,7 +71,8 @@ __all__ = ["run_adc", "InputError", "AdcMatrix",
            "guess_symmetries", "guesses_spin_flip", "guess_zero", "LazyMp",
            "adc0", "cis", "adc1", "adc2", "adc2x", "adc3",
            "cvs_adc0", "cvs_adc1", "cvs_adc2", "cvs_adc2x", "cvs_adc3",
-           "nuclear_gradient",
+           "nuclear_gradient", "nuclear_gradient_scanner",
+           "NuclearGradientScanner", "GroundStateTarget", "ExcitedStateTarget",
            "banner"]
 
 __version__ = "0.17.1"
--- a/adcc/gradients/__init__.py
+++ b/adcc/gradients/__init__.py
@@ -36,7 +36,7 @@ from .orbital_response import (
     orbital_response, orbital_response_rhs, energy_weighted_density_matrix
 )
 from .amplitude_response import amplitude_relaxed_densities
-from .scanner import (ExcitedStateTarget, GroundStateTarget,
+from .scanner import (ExcitedStateTarget, GroundStateTarget,  # noqa: F401
                       NuclearGradientScanner, density_overlap_score,
                       nuclear_gradient_scanner)
 
--- a/adcc/gradients/__init__.py
+++ b/adcc/gradients/__init__.py
@@ -36,9 +36,14 @@ from .orbital_response import (
     orbital_response, orbital_response_rhs, energy_weighted_density_matrix
 )
 from .amplitude_response import amplitude_relaxed_densities
-from .scanner import (ExcitedStateTarget, GroundStateTarget,  # noqa: F401
+from .scanner import (ExcitedStateTarget, GroundStateTarget,
                       NuclearGradientScanner, density_overlap_score,
                       nuclear_gradient_scanner)
+
+__all__ = [
+    "nuclear_gradient", "nuclear_gradient_scanner", "NuclearGradientScanner",
+    "GroundStateTarget", "ExcitedStateTarget", "density_overlap_score",
+]
 
 
 @dataclass(frozen=True)
--- a/adcc/gradients/__init__.py
+++ b/adcc/gradients/__init__.py
@@ -36,6 +36,9 @@ from .orbital_response import (
     orbital_response, orbital_response_rhs, energy_weighted_density_matrix
 )
 from .amplitude_response import amplitude_relaxed_densities
+from .scanner import (ExcitedStateTarget, GroundStateTarget,
+                      NuclearGradientScanner, density_overlap_score,
+                      nuclear_gradient_scanner)
 
 
 @dataclass(frozen=True)
--- a/adcc/gradients/scanner.py
+++ b/adcc/gradients/scanner.py
@@ -59,7 +59,6 @@ class ExcitedStateTarget:
         """Return keyword arguments forwarded to :func:`adcc.run_adc`."""
         ret = dict(self.run_adc_kwargs)
         ret["method"] = self.method
-        ret.setdefault("output", None)
         return ret
 
 
@@ -263,8 +262,11 @@ class NuclearGradientScanner:
         return excitations[best]
 
     def _descriptor(self, excitation, mol):
+        # PySCF SCF scanner objects mutate their internal Mole in-place between
+        # geometries.  Keep an independent copy for cross-overlap root tracking;
+        # otherwise the "previous" AO basis silently turns into the current one.
         return _TrackingDescriptor(
-            mol=mol,
+            mol=mol.copy(),
             transition_dm=_safe_ao_ndarray(excitation, "transition_dm_ao"),
             state_diffdm=_safe_ao_ndarray(excitation, "state_diffdm_ao"),
         )
--- a/adcc/gradients/scanner.py
+++ b/adcc/gradients/scanner.py
@@ -1,0 +1,503 @@
+#!/usr/bin/env python3
+## vi: tabstop=4 shiftwidth=4 softtabstop=4 expandtab
+## ---------------------------------------------------------------------
+##
+## Copyright (C) 2026 by the adcc authors
+##
+## This file is part of adcc.
+##
+## adcc is free software: you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published
+## by the Free Software Foundation, either version 3 of the License, or
+## (at your option) any later version.
+##
+## adcc is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with adcc. If not, see <http://www.gnu.org/licenses/>.
+##
+## ---------------------------------------------------------------------
+"""Geometry-optimisation scanner for adcc nuclear gradients.
+
+The scanner owns the per-geometry PySCF -> adcc -> gradient loop.  It is
+intentionally PySCF-only for now, because the production direct nuclear-gradient
+path is PySCF-only as well.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, Callable, Optional
+import warnings
+
+import numpy as np
+
+from adcc.Excitation import Excitation
+from adcc.LazyMp import LazyMp
+
+
+@dataclass(frozen=True)
+class GroundStateTarget:
+    """Ground-state MP target for a nuclear-gradient scanner."""
+
+    level: int = 2
+
+
+@dataclass(frozen=True)
+class ExcitedStateTarget:
+    """Excited-state ADC target for a nuclear-gradient scanner."""
+
+    method: str
+    state_index: int = 0
+    n_states: Optional[int] = None
+    kind: str = "any"
+    n_singlets: Optional[int] = None
+    n_triplets: Optional[int] = None
+    n_spin_flip: Optional[int] = None
+    core_orbitals: Any = None
+    frozen_core: Any = None
+    frozen_virtual: Any = None
+    conv_tol: Optional[float] = None
+    environment: Any = None
+    solverargs: dict[str, Any] = field(default_factory=dict)
+
+    def run_adc_kwargs(self) -> dict[str, Any]:
+        """Return keyword arguments forwarded to :func:`adcc.run_adc`."""
+        ret = {
+            "method": self.method,
+            "n_states": self.n_states,
+            "kind": self.kind,
+            "n_singlets": self.n_singlets,
+            "n_triplets": self.n_triplets,
+            "n_spin_flip": self.n_spin_flip,
+            "core_orbitals": self.core_orbitals,
+            "frozen_core": self.frozen_core,
+            "frozen_virtual": self.frozen_virtual,
+            "conv_tol": self.conv_tol,
+            "environment": self.environment,
+            "output": None,
+        }
+        ret.update(self.solverargs)
+        return {key: value for key, value in ret.items() if value is not None}
+
+
+@dataclass
+class TrackingResult:
+    """Diagnostic information for the most recent root-tracking decision."""
+
+    index: int
+    scores: np.ndarray
+
+
+@dataclass
+class _TrackingDescriptor:
+    mol: Any
+    transition_dm: Optional[np.ndarray]
+    state_diffdm: Optional[np.ndarray]
+
+
+@dataclass
+class _ScfSettings:
+    conv_tol: float
+    conv_tol_grad: Optional[float]
+    max_cycle: int
+    verbose: int
+    attributes: dict[str, Any]
+
+
+class NuclearGradientScanner:
+    """Callable scanner returning adcc energies and nuclear gradients.
+
+    Parameters are usually created through :func:`nuclear_gradient_scanner`.
+    The scanner accepts Cartesian coordinates in Bohr and returns an energy in
+    Hartree and a gradient in Hartree/Bohr.
+    """
+
+    def __init__(self, mol_template=None, *, scf_template=None,
+                 scf_factory: Optional[Callable[[Any], Any]] = None,
+                 target: GroundStateTarget | ExcitedStateTarget | str | None = None,
+                 method: Optional[str] = None, state_index: int = 0,
+                 mp_level: int = 2, n_states: Optional[int] = None,
+                 kind: str = "any", n_singlets: Optional[int] = None,
+                 n_triplets: Optional[int] = None,
+                 n_spin_flip: Optional[int] = None, core_orbitals: Any = None,
+                 frozen_core: Any = None, frozen_virtual: Any = None,
+                 adc_conv_tol: Optional[float] = None, environment: Any = None,
+                 scf_type: Optional[str] = None, scf_conv_tol: Optional[float] = None,
+                 scf_conv_tol_grad: Optional[float] = None,
+                 scf_max_cycle: Optional[int] = None, symmetry: Any = None,
+                 verbose: Optional[int] = None, follow: str = "overlap",
+                 tracking_min_score: float = 0.0, tracking_min_gap: float = 0.0,
+                 gradient_kwargs: Optional[dict[str, Any]] = None,
+                 adc_solverargs: Optional[dict[str, Any]] = None):
+        self._pyscf = _import_pyscf()
+        self.scf_factory = scf_factory
+        self.scf_type = scf_type
+        self._requested_symmetry = symmetry
+        self.follow = follow
+        self.tracking_min_score = tracking_min_score
+        self.tracking_min_gap = tracking_min_gap
+        self.gradient_kwargs = dict(gradient_kwargs or {})
+        self.last_tracking: Optional[TrackingResult] = None
+
+        if scf_template is None and _is_pyscf_scf(self._pyscf, mol_template):
+            scf_template = mol_template
+            mol_template = scf_template.mol
+        if scf_template is not None and mol_template is None:
+            mol_template = scf_template.mol
+        if mol_template is None:
+            raise ValueError("mol_template or scf_template needs to be provided.")
+
+        self._init_mol_template(mol_template)
+        self.scf_class = type(scf_template) if scf_template is not None else None
+        self.scf_settings = self._capture_scf_settings(
+            scf_template, scf_conv_tol, scf_conv_tol_grad, scf_max_cycle, verbose
+        )
+        self.target = self._normalise_target(
+            target, method, state_index, mp_level, n_states, kind, n_singlets,
+            n_triplets, n_spin_flip, core_orbitals, frozen_core, frozen_virtual,
+            adc_conv_tol, environment, adc_solverargs or {}
+        )
+
+        self.previous_scf = None
+        self.previous_dm = None
+        self.previous_states = None
+        self.previous_excitation: Optional[Excitation] = None
+        self.previous_descriptor: Optional[_TrackingDescriptor] = None
+        self.last_scf = None
+        self.last_states = None
+        self.last_excitation = None
+        self.last_gradient = None
+
+    @property
+    def natoms(self) -> int:
+        return len(self.atom_symbols)
+
+    def __call__(self, coords):
+        """Return ``(energy, gradient)`` for Cartesian coordinates in Bohr."""
+        coords = self._coords_array(coords)
+        scfres = self._run_scf(coords)
+        target = self._build_target(scfres)
+
+        from adcc.gradients import nuclear_gradient
+        grad = nuclear_gradient(target, **self.gradient_kwargs)
+        if isinstance(target, LazyMp):
+            energy = target.energy(self.target.level)
+        else:
+            energy = target.total_energy
+
+        self.last_scf = scfres
+        self.last_gradient = grad
+        self.previous_scf = scfres
+        self.previous_dm = scfres.make_rdm1()
+        if isinstance(target, Excitation):
+            self.previous_states = self.last_states
+            self.previous_excitation = target
+            self.previous_descriptor = self._descriptor(target, scfres.mol)
+        return float(energy), np.asarray(grad.total)
+
+    def calc_new(self, coords):
+        """geomeTRIC custom-engine entry point.
+
+        ``coords`` is a flattened Cartesian coordinate array in Bohr.  The
+        returned gradient is flattened in Hartree/Bohr.
+        """
+        energy, gradient = self(coords)
+        return {"energy": energy, "gradient": np.asarray(gradient).ravel()}
+
+    def _init_mol_template(self, mol):
+        self.atom_symbols = [mol.atom_symbol(i) for i in range(mol.natm)]
+        self.basis = mol.basis
+        self.ecp = getattr(mol, "ecp", None)
+        self.charge = mol.charge
+        self.spin = mol.spin
+        self.cart = bool(getattr(mol, "cart", False))
+        self.symmetry = (getattr(mol, "symmetry", False)
+                         if self._requested_symmetry is None
+                         else self._requested_symmetry)
+        self.output = getattr(mol, "output", None)
+        self.max_memory = getattr(mol, "max_memory", None)
+        self.initial_coords = np.asarray(mol.atom_coords(unit="Bohr"), dtype=float)
+
+    def _capture_scf_settings(self, scf_template, conv_tol, conv_tol_grad,
+                              max_cycle, verbose):
+        attributes = {}
+        if scf_template is not None:
+            for attr in [
+                "diis_space", "diis_start_cycle", "damp", "level_shift",
+                "direct_scf", "init_guess", "max_memory",
+            ]:
+                if hasattr(scf_template, attr):
+                    attributes[attr] = getattr(scf_template, attr)
+            default_conv_tol = scf_template.conv_tol
+            default_conv_tol_grad = getattr(scf_template, "conv_tol_grad", None)
+            default_max_cycle = scf_template.max_cycle
+            default_verbose = scf_template.verbose
+        else:
+            default_conv_tol = 1e-11
+            default_conv_tol_grad = 1e-9
+            default_max_cycle = 150
+            default_verbose = 0
+        return _ScfSettings(
+            conv_tol=default_conv_tol if conv_tol is None else conv_tol,
+            conv_tol_grad=(default_conv_tol_grad if conv_tol_grad is None
+                           else conv_tol_grad),
+            max_cycle=default_max_cycle if max_cycle is None else max_cycle,
+            verbose=default_verbose if verbose is None else verbose,
+            attributes=attributes,
+        )
+
+    def _normalise_target(self, target, method, state_index, mp_level, n_states,
+                          kind, n_singlets, n_triplets, n_spin_flip,
+                          core_orbitals, frozen_core, frozen_virtual,
+                          adc_conv_tol, environment, solverargs):
+        if isinstance(target, (GroundStateTarget, ExcitedStateTarget)):
+            return target
+        if isinstance(target, str):
+            if target.lower().startswith("mp"):
+                try:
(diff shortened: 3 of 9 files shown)


maxscheurer:
## Implementation Plan

### Problem analysis

Issue 221 asks for a geometry-optimization interface on the `gradients` branch: a scanner that freezes the PySCF SCF setup and adcc MP/ADC setup once, then evaluates a new geometry as `(energy, gradient)`. The scanner should support direct use from Python and a geomeTRIC custom-engine adapter. It should also provide a clean path for MECP/MECI-style optimizations between two states without derivative couplings.

The key implementation constraint is that every geometry step crosses two libraries. PySCF owns the molecular geometry, SCF convergence, orbital guess, and AO basis; adcc owns `LazyMp`/`run_adc`, state objects, and `nuclear_gradient`. The scanner should therefore own the PySCF lifecycle, hand the converged SCF object into adcc, select the target state, call `nuclear_gradient`, and cache enough previous-step state for SCF continuity and root tracking.

User clarification incorporated: adcc has no density-fitting support right now, so density-fitting consistency is not part of this plan. Root tracking should use the previous `ExcitedStates`/`Excitation` object and the root-character information adcc already exposes, especially transition densities and state-difference densities.

### Deliverables

1. **Public scanner API**
   - Add `adcc.nuclear_gradient_scanner(...)` as the user-facing factory.
   - The returned object should expose:
     - `scanner(coords) -> (energy, gradient)` with gradient shape `(natoms, 3)`.
     - `scanner.calc_new(flat_coords) -> {"energy": energy, "gradient": flat_gradient}` for geomeTRIC custom-engine style use.
   - Initial supported backend: PySCF.

2. **PySCF lifecycle management**
   - Build a scanner-owned PySCF `mol`/`mf` at each geometry from a template molecule or SCF object.
   - Freeze basis, charge, spin, reference type, symmetry choice, SCF convergence thresholds, and max-cycle settings at construction.
   - Reuse the previous converged SCF density or MO coefficients as the next `dm0` guess.
   - Hard-fail on unconverged SCF rather than returning a discontinuous bad point.

3. **Ground-state and excited-state target support**
   - Ground-state path: build `ReferenceState(scfres)`, `LazyMp`, evaluate requested MP level, call `nuclear_gradient(mp)`.
   - Excited-state path: run `adcc.run_adc(scfres, **frozen_adc_kwargs)`, select/track one `Excitation`, call `nuclear_gradient(excitation)`.
   - Energy source should be `mp.energy(level)` for ground-state targets and `Excitation.total_energy` for excited-state targets.

4. **Stateful root tracking**
   - Store previous SCF object/molecule, previous `ExcitedStates`, previous selected `Excitation`, and cached tracking descriptors.
   - First call seeds from `state_index`.
   - Later calls choose among current `states.excitations` using density-based state character from the previous selected excitation.
   - Primary descriptors:
     - `transition_dm_ao`
     - `state_diffdm_ao`
   - Compare old and new AO-basis matrices through PySCF AO cross-overlap, e.g. `pyscf.gto.intor_cross("int1e_ovlp", old_mol, new_mol)`, rather than raw MO-vector overlap.
   - Provide an explicit fixed-index mode for debugging or simple cases, but do not make it the robust default for excited-state optimization.
   - Add a tracking-loss guard: if the best match is ambiguous or below a threshold, raise a clear error or warning according to the configured policy.

5. **MECP/MECI-oriented paired-state interface**
   - Add a paired scanner/engine abstraction that evaluates two targets at the same geometry and exposes both energies/gradients to a penalty-function MECP objective or to geomeTRIC's MECI/MECP machinery.
   - Support excited/excited pairs from one SCF and one ADC calculation when both states are in the same `ExcitedStates` object.
   - Support ground/excited pairs from one SCF plus `LazyMp` and ADC.
   - Track both states statefully and ensure they remain distinct.
   - Do not require derivative couplings.

6. **geomeTRIC integration**
   - Keep geomeTRIC optional; no hard dependency for importing adcc or using the plain scanner.
   - Provide a thin adapter or method that returns the geomeTRIC `calc_new` contract: flattened Bohr coordinates in, Hartree energy and flattened Hartree/Bohr gradient out.
   - If a convenience optimizer wrapper is added, guard imports and raise an informative error when geomeTRIC is unavailable.

7. **Docs and examples**
   - Add a geometry-optimization section to `docs/gradients.rst`.
   - Add examples under `examples/optimization/`, replacing the empty/stale prototype area with source files.
   - Include at least one ground-state example and one excited-state/geomeTRIC example. Include MECP usage if the paired-state interface lands in the same PR; otherwise document it as a follow-up limitation.

### Files to create or modify

- `adcc/gradients/scanner.py` — new scanner implementation, target specifications, root-tracking helpers, geomeTRIC adapter.
- `adcc/gradients/__init__.py` — re-export the scanner factory alongside `nuclear_gradient`.
- `adcc/__init__.py` — public import and `__all__` entry for `nuclear_gradient_scanner` and any intentionally public target spec classes.
- `adcc/backends/pyscf.py` — optional small helper(s) for rebuilding PySCF SCF objects or AO cross-overlap if this avoids duplicating backend-specific code in the scanner.
- `pyproject.toml` — optionally add a geomeTRIC extra if the project wants discoverable optional install support.
- `adcc/tests/geomopt_scanner_test.py` or similar — new unit/functional tests for scanner behavior.
- `adcc/tests/functionality_geomopt_test.py` or similar — end-to-end geomeTRIC smoke tests, skipped unless PySCF and geomeTRIC are available.
- `docs/gradients.rst` — scanner and optimizer documentation.
- `examples/optimization/` — new runnable example scripts.

### Testing approach

Testing is mandatory for each new layer.

1. **Scanner unit tests, PySCF required but geomeTRIC not required**
   - A single scanner call returns `(float, ndarray)` with gradient shape `(natoms, 3)`.
   - `calc_new(flat_coords)` returns keys `energy` and `gradient`, with flattened gradient shape `(3 * natoms,)`.
   - The scanner result at a fixed geometry matches the explicit loop: PySCF SCF, `run_adc` or `LazyMp`, `nuclear_gradient`.
   - Ground-state path works for MP2 on a tiny molecule.
   - Excited-state path works for a low singlet state on a tiny molecule.
   - Unconverged SCF is reported as an error.
   - Previous-density/MO guess is actually passed on subsequent calls; if iteration-count assertions are too brittle, test the state update and identical final energies instead.

2. **Root-tracking tests**
   - Fixed-index mode selects the requested positional state.
   - Density-overlap mode stores previous state descriptors and selects the candidate with highest transition/difference-density similarity.
   - Unit-test the overlap/scoring function separately using controlled matrices and mock/cached AO overlap matrices.
   - Integration-test on a small displaced-geometry sequence that the selected state character remains continuous, even if energy ordering changes or nearly changes.
   - Test ambiguous tracking behavior: low/near-tied scores should warn or fail according to policy instead of silently switching state.

3. **MECP/paired-state tests**
   - Excited/excited paired scanner evaluates both target states from one ADC result and returns distinct energies/gradients.
   - Ground/excited paired scanner evaluates MP and ADC targets from one SCF result.
   - Penalty/objective gradient combination is tested algebraically with controlled energies/gradients.
   - If an end-to-end MECP optimization is too expensive for standard CI, add a small skipped/slow smoke test and keep fast unit coverage for the paired engine mechanics.

4. **geomeTRIC tests**
   - Skipped when geomeTRIC is unavailable.
   - Custom-engine adapter accepts flattened Bohr coordinates and returns Hartree/Hartree-per-Bohr values in the expected shape.
   - A short ground-state optimization on a tiny molecule converges to small gradient norm.
   - If feasible within runtime, a short excited-state optimization smoke test converges or at least performs multiple scanner calls with state tracking enabled.

5. **Docs/example validation**
   - Examples should be importable/runnable when optional dependencies are installed.
   - Docs should mention units, optional geomeTRIC dependency, PySCF-only initial scope, SCF continuity, and root-tracking caveats.

### Acceptance criteria

- `adcc.nuclear_gradient_scanner` is importable from the public API.
- The scanner supports ground-state MP and excited-state ADC targets.
- The scanner stores previous-step state and uses it for SCF guess continuity.
- Excited-state optimization has density-based state following using previous `ExcitedStates`/`Excitation` data, not only positional index.
- The geomeTRIC adapter satisfies `calc_new(flat_coords)` with correct units and shapes.
- A small PySCF/adcc scanner test matches the explicit `run_adc`/`LazyMp` plus `nuclear_gradient` loop.
- At least one end-to-end geometry optimization smoke test is present and skipped cleanly when optional dependencies are missing.
- Docs and examples show how to run a geomeTRIC-driven optimization.
- MECP support is either implemented with tests in this PR or clearly separated into a follow-up deliverable if the initial scanner PR is intentionally narrowed.

### Risks and open questions

- **Root tracking correctness:** comparing states across geometries is the main correctness risk. AO-basis density descriptors plus PySCF cross-overlap are the planned robust route; raw excitation-vector overlap across different MO bases should not be the primary method.
- **Dark or weak-transition states:** transition-density similarity may be weak, so state-difference density should be included as a fallback/second channel.
- **Near degeneracy in MECP:** two tracked states may become hard to distinguish. The paired tracker needs distinct assignment and ambiguity handling.
- **Runtime:** end-to-end optimizer tests can be slow. Keep most coverage in unit tests and use tiny/loose smoke tests for full optimizations.
- **API size:** avoid over-coupling the core scanner to geomeTRIC. Prefer an adcc-native scanner plus thin adapters.
- **Symmetry/frame handling:** optimizers expect coordinate frames not to change unexpectedly. The scanner should pin symmetry behavior and document this.

### Suggested implementation stages

1. Core PySCF scanner with ground-state MP support and `calc_new`/`__call__` contracts.
2. Excited-state fixed-index support and basic tests.
3. Density-based root tracking using previous `ExcitedStates`/`Excitation` descriptors.
4. geomeTRIC adapter and small optimization smoke tests.
5. Paired-state MECP interface and tests.
6. Docs and examples.

---
*Plan created by mach6*

maxscheurer:
## Progress Update

Implemented the initial geometry-optimization scanner core:

- Added public `adcc.nuclear_gradient_scanner` / `NuclearGradientScanner` exports.
- Added PySCF-owned per-geometry SCF lifecycle from a template molecule or SCF object, preserving original molecular/SCF settings and reusing the previous converged density as the next guess.
- Added ground-state MP and excited-state ADC target support.
- Added stateful excited-state tracking using previous `ExcitedStates` / `Excitation` descriptors, with AO-basis transition and state-difference density overlaps through PySCF AO cross-overlap integrals.
- Added geomeTRIC-style `calc_new(flat_coords)` support.
- Added scanner unit tests, docs, an optional `geomopt` extra, and an example under `examples/optimization/`.

Verification run:

- `python -m py_compile adcc/gradients/scanner.py adcc/tests/geomopt_scanner_test.py examples/optimization/pyscf_adcc_geoopt.py`
- `python -m pytest adcc/tests/geomopt_scanner_test.py -q` — 4 passed

MECP paired-state orchestration is not fully implemented in this batch; the scanner/root-tracking foundation is in place for that next layer.

**Commit:** `7d274da`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

Refactored the scanner API to use a configured PySCF SCF object as the primary interface:

- `adcc.nuclear_gradient_scanner(mf, ...)` now requires a PySCF SCF object instead of a molecule template.
- Removed the previous scanner-specific SCF keyword surface; SCF settings now live on the PySCF `mf` object.
- Excited-state keyword arguments such as `conv_tol`, `n_singlets`, `core_orbitals`, etc. are forwarded unchanged to `adcc.run_adc`.
- The scanner uses PySCF's `as_scanner` machinery for per-geometry SCF and guess continuity.
- Updated tests, docs, the ground-state optimization example, and added/updated the excited-state optimization example script.

Verification run:

- `python -m py_compile adcc/gradients/scanner.py adcc/tests/geomopt_scanner_test.py examples/optimization/pyscf_adcc_geoopt.py examples/optimization/pyscf_adcc_geoopt_excited.py`
- `python -m pytest adcc/tests/geomopt_scanner_test.py -q` — 6 passed

**Commit:** `81f6273`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

Fixed a real root-tracking bug and reworked the excited-state example.

**Scanner fix**
- The PySCF SCF scanner mutates its internal `Mole` in-place between geometries. The tracking descriptor previously stored that `Mole` by reference, so the "previous" AO basis used for cross-overlap root tracking could silently become the current geometry. The descriptor now stores an independent `mol.copy()`.
- Added a regression test asserting the descriptor's `mol` snapshot stays independent after the source `Mole` geometry is mutated.

**Excited-state example (`examples/optimization/pyscf_adcc_geoopt_excited.py`)**
- Now optimises in two stages: first relaxes the ground state at MP2 level, then runs the ADC(2) excited-state optimisation starting from the relaxed MP2 geometry.
- Switched basis from STO-3G to cc-pVDZ.
- Both stages converge cleanly in a smoke run (MP2 in 3 steps, ADC(2) in 5 steps) and the followed ADC root stays positive throughout.

**Note on negative roots:** the mol-mutation bug was genuine, but it did not fully explain the earlier negative excitation energies. Those came from the STO-3G lowest-singlet water surface being driven toward a near-crossing/unstable region. Starting from the MP2-relaxed geometry in a larger basis avoids that pathological path in the demo.

All scanner tests pass (`python -m pytest adcc/tests/geomopt_scanner_test.py -q`, 7 passed).

**Commit:** `aabe731`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

Switched the excited-state example to the s-trans-butadiene geometry optimisation showcase from Rehn & Dreuw (*J. Chem. Phys.* 150, 174110, 2019).

- Replaced the water geometry with planar all-trans butadiene (C4H6).
- Switched the basis from cc-pVDZ to 6-31G*.
- Updated the docstring to cite the reference.

The two-stage MP2 -> ADC(2) workflow is unchanged; the optimisation runs cleanly on this system.

**Commit:** `f8c1f0a`

---
*Progress tracked by mach6*

maxscheurer:
## Progress Update

Filled the two testing gaps identified in the plan-vs-implementation check.

**Scanner unit tests** (`adcc/tests/geomopt_scanner_test.py`, 7 -> 14 tests)
- Unconverged SCF is reported as a `RuntimeError`.
- Excited-state scanner result matches the explicit `run_adc` + `nuclear_gradient` loop; fixed-index mode selects the requested positional state.
- Density-overlap root tracking re-selects the seeded state (self-overlap ~1) and stays continuous under a small displacement.
- Tracking-loss guards: below-min-score and ambiguous-gap cases both raise clearly.
- SCF guess continuity: previous state is updated and the surface is reproduced after a there-and-back move.

**End-to-end geomeTRIC smoke test** (new `adcc/tests/functionality_geomopt_test.py`)
- Drives a full MP2/STO-3G ground-state optimisation through geomeTRIC and asserts the gradient at the optimised geometry is ~0 (max < 5e-4).
- Checks the `calc_new` custom-engine contract (flattened Bohr in, energy + flattened Hartree/Bohr gradient out).
- Both skip cleanly unless PySCF and geomeTRIC are installed (verified the guard skips when a dependency is missing).

All 16 geomopt tests pass.

**Commit:** `6a17103`

---
*Progress tracked by mach6*

maxscheurer:
## Code Review

Multi-agent review of the geometry-optimization scanner. Findings below; severity reflects the reporting agent's assessment (an independent assessment follows separately).

### Critical

_None._

### Important

**1. Root-tracking guards are disabled by default (error-auditor, conf 92).** `scanner.py` defaults `tracking_min_score=0.0` and `tracking_min_gap=0.0`. With the transition channel using `use_abs=True` (range `[0,1]`), `scores[best] < 0.0` is essentially never true, and because `scores` is sorted descending `best - second >= 0` always, so the gap guard can never fire either. Net effect: in default `follow="overlap"` mode the scanner silently follows the highest-overlap root even on a near-degeneracy / avoided crossing / poor match, producing a discontinuous PES and wrong gradients with no warning. The guards that exist to catch bad tracking are off unless the user sets both thresholds.

**2. MP3 ground-state target pairs an MP3 energy with an MP2 gradient (code-reviewer, conf 88).** `_normalise_target` accepts `target="mp3"` / `GroundStateTarget(level=3)`. In `__call__`, `energy = target.energy(self.target.level)` returns the MP3 energy, but `nuclear_gradient(LazyMp)` always computes the MP2 gradient (no MP3 gradient exists in adcc). The returned `(energy, gradient)` pair no longer satisfies `gradient = dE/dx`, so an optimizer drives to a non-stationary point. The scanner should reject `level != 2`.

**3. `_safe_ao_ndarray` swallows every exception, silently degrading tracking (error-auditor, conf 88).** `except Exception -> return None` (warning only). If `transition_dm_ao` or `state_diffdm_ao` fails consistently (real bug, unsupported variant, memory error), that channel is silently dropped and tracking proceeds on the single remaining channel — materially weaker root following — while returning a normal result. Only when both fail does it raise. The blanket catch also masks genuine programming bugs as "tracking ran on fewer channels."

**4. Overlap-tracking reordering behavior is never tested (test-reviewer, conf 88).** The feature's headline value — following the same physical state when root *ordering* flips (so the selected index differs from `state_index`) — is not exercised. All overlap tests re-run at identical geometry (self-overlap ~1) or use a 0.02 Bohr nudge too small to reorder roots. A regression making tracking silently fall back to positional indexing would pass the whole suite.

**5. No end-to-end excited-state optimization (test-reviewer, conf 85).** Both geomeTRIC smoke tests use `method="mp2"`. The excited-state path through many real steps — where SCF-guess continuity, the reused PySCF scanner, and overlap tracking interact — is never run to convergence, which is the feature's main use case.

### Suggestions

**6. Degraded-tracking warning is deduplicated by Python's warning filter (error-auditor, conf 85).** The one signal of channel failure (finding 3) originates from a fixed line, so the default `"once"`-style filter shows it on the first failing step and never again, even though every subsequent step stays degraded. Surface this via a per-call diagnostic (e.g. on `last_tracking`) instead.

**7. `density_overlap_score` silently substitutes identity overlaps when `s_old`/`s_new` omitted (error-auditor, conf 80).** This module-level (effectively public) helper substitutes `np.eye(n)` for real AO overlaps when they default to `None`, returning a plausible-but-wrong score in any non-orthogonal basis with no warning. Internal callers always pass real overlaps, so production is safe, but it is a footgun.

**8. Invalid `follow` value validated only on the second geometry (code-reviewer, conf 90).** On the first call `previous_descriptor is None`, so an invalid `follow` (e.g. `"closest"`) silently falls into the index branch; the `ValueError` only fires on the second call, after a full SCF+ADC+gradient has been spent. Validate in `__init__`.

**9. `target="mp2"` string path ignores `run_adc` kwargs without the warning (code-reviewer, conf 84).** `nuclear_gradient_scanner(scf, "mp2", n_singlets=3)` silently drops `n_singlets`, whereas `method="mp2"` emits the "Ignoring run_adc keyword arguments" warning. Inconsistent handling of the same request.

**10. Target-resolution / validation / scoring edge cases untested (test-reviewer, conf 82-85).** Untested: `target="mp3"` and direct `GroundStateTarget`/`ExcitedStateTarget` instances; `_mp_level` `ValueError` fallback; `state_index` out-of-range and empty-`excitations` guards; invalid-`follow` raise; `_safe_ao_ndarray` failure + `weight==0` raise; `density_overlap_score` NaN-denominator path and non-identity overlaps; the ground-state kwargs-ignored warning.

**11. `test_overlap_tracking_continuous_under_small_displacement` is a weak assertion (test-reviewer, conf 82).** Tracking `state_index=0` with a tiny displacement and asserting `|domega| < 0.05` would pass identically under plain fixed-index selection, and it never asserts which index was selected. As written it adds little over the self-overlap test.

**Simplifications (simplifier, conf 80-85):**
- `previous_states` / `previous_excitation` are write-only dead code (only `previous_descriptor` / `previous_scf` are read).
- `s_old` / `s_new` overlap matrices are recomputed each loop iteration in `_tracking_score`; hoist out of the loop.
- Redundant local `from adcc import LazyMp` in `_build_target` (already imported at module top).
- `_PyscfModules` dataclass could be a `SimpleNamespace` (stylistic).

### Strengths

- PySCF scanner lifecycle (`as_scanner` / `undo_scanner` reuse for guess continuity) verified correct by code-reviewer; `_mol_at` settings preservation is sound.
- Root-tracking math (AO cross-overlap cosine, NaN filtering via `np.isfinite`) is a sound measure; NaN cannot leak into the final scores.
- The mol-snapshot independence fix (`mol.copy()` in `_descriptor`) is correct and has a regression test.
- Ground-state end-to-end smoke test genuinely asserts a stationary point (`gmax < 5e-4` post-optimization), skips cleanly without PySCF/geomeTRIC.
- Completeness: all eight issue 221 acceptance criteria are met; MECP is explicitly and acceptably deferred per the plan's own criteria.

---
*Reviewed by mach6*

maxscheurer:
## Review Assessment

Independent verification of the [code review](https://github.com/adc-connect/adcc/pull/226#issuecomment-4731541656) — each finding checked against the actual source.

### Classifications

|---|---|---|
| 1: Root-tracking guards disabled by default | **genuine** | `order = argsort(scores)[::-1]` (descending) → `best - second >= 0`, so the gap guard `< 0.0` can never fire; score guard nearly never fires (transition channel `use_abs` ∈[0,1]). Default `follow="overlap"` switches roots silently with no diagnostic. scanner.py:94,251,257. |
| 2: MP3 energy + MP2 gradient | **genuine** | `_mp_level("mp3",2)→3` is reachable; `energy(3)` is MP3 but `amplitude_relaxed_densities` hardcodes `method_name="mp2"` (amplitude_response.py:705), no MP3 DISPATCH entry. `(E,∇E)` non-conservative → optimizer drives to non-stationary point. |
| 3: `_safe_ao_ndarray` swallows all exceptions | **genuine** | Blanket `except Exception`→None silently drops a tracking channel and masks real bugs; only double-failure raises. scanner.py:339. |
| 4: Reordering behavior untested | **genuine** (test gap) | Headline feature (selected index ≠ `state_index` under a root flip) never exercised. Not deferred per policy. |
| 5: No e2e excited-state optimization | **genuine** (test gap) | Both geomeTRIC tests use `mp2`; the main use case never runs to convergence. |
| 6: Degraded-tracking warning deduplicated | **genuine** (minor) | `warnings.warn` from fixed line 340; default filter shows once → persistent degradation reported once. Better surfaced on `last_tracking`. Tied to finding 3. |
| 7: `density_overlap_score` identity substitution | **nitpick** | Real footgun for a public helper, but all internal callers pass real overlaps (lines 288-289); production is safe. |
| 8: `follow` validated only on 2nd geometry | **genuine** | `... or previous_descriptor is None` (line 233) routes invalid values into index branch on call 1; error only after a wasted SCF+ADC+gradient. Validate in `__init__`. |
| 9: `target="mp2"` drops kwargs without warning | **genuine** (minor) | Inconsistent with `method="mp2"` which warns for the same request. scanner.py:165-175. |
| 10: Misc edge cases untested | **genuine** (test gap) | Many validation/scoring branches untested; cheap unit tests. |
| 11: Weak continuous-displacement test | **nitpick** | Would pass under fixed-index, never asserts selected index — additive weakness, subsumed by fixing finding 4. |
| Simpl: dead `previous_states`/`previous_excitation` | **genuine** (cleanup) | Confirmed write-only (lines 124-125, 154-155), never read anywhere. |
| Simpl: `s_old`/`s_new` recomputed in loop | **nitpick** | Minor perf; hoist out of the 2-iteration loop. |
| Simpl: redundant `LazyMp` import | **nitpick** | Line 219 re-imports a module-level symbol; harmless. |
| Simpl: `_PyscfModules` → SimpleNamespace | **nitpick** | Pure style; codebase favors dataclasses. |

**Counts:** 9 genuine (3 of which are test gaps), 5 nitpicks, 0 false positives, 0 deferred.

### Action Plan

1. **Reject `GroundStateTarget(level != 2)` (finding 2)** — validate in `_normalise_target`/`__init__` and raise. Highest correctness impact: MP3 energy + MP2 gradient breaks optimizers.
2. **Fix root-tracking guard defaults/semantics (finding 1)** — the gap guard can never fire at 0.0. Set sensible non-zero defaults, or always emit a diagnostic when the best–second gap is small or the index changes, so silent state-switching is observable.
3. **Narrow `_safe_ao_ndarray` handling + surface degradation (findings 3, 6)** — catch only the expected "operator unavailable" case; let real bugs propagate. Record per-call channel degradation on `last_tracking` instead of a deduplicated warning.
4. **Validate `follow` in `__init__` (finding 8)** — reject anything but `"overlap"`/`"index"` before any SCF/ADC work.
5. **Make `target="mp2"` warn on dropped run_adc kwargs (finding 9)** — route the string-MP path through the same kwargs check as `method=`.
6. **Add missing tests (findings 4, 5, 10, 11) — do not defer.** Real root-reorder case asserting `last_tracking.index != state_index` with continuous character; e2e excited-state geomeTRIC smoke test; unit tests for `level!=2` rejection, invalid `follow`, `state_index` out-of-range, empty `excitations`, `_safe_ao_ndarray` failure + `weight==0`, `density_overlap_score` NaN + non-identity overlaps, and the ground-state kwargs-ignored warning.
7. **Cleanup** — remove write-only `previous_states`/`previous_excitation`; optionally hoist `s_old`/`s_new` and drop the redundant `LazyMp` import. Remaining simplifications are nitpicks.

No deferred items — all genuine findings are in scope for this PR; no separate tracking issue needed if the test gaps are addressed here.

---
*Assessment by mach6*

maxscheurer:
## Progress Update

Addressed the genuine review findings from the assessment. All changes in `adcc/gradients/scanner.py` plus tests; **30 geomopt tests pass, flake8 clean**.

**Correctness fixes**
- **Finding 2** — `_check_ground_state_level` rejects MP level != 2 with `NotImplementedError` at construction, covering all paths (`target="mp3"`, `method="mp3"`, and a directly-passed `GroundStateTarget(level=3)`). No more pairing an MP3 energy with an MP2 gradient.
- **Finding 1** — `TrackingResult` now carries `best_score`, `gap`, `previous_index`, `switched`, and `unavailable_channels`, populated on `last_tracking` after every overlap call. Silent root-switching and small-gap matches are now observable rather than hidden behind ineffective default guards.
- **Findings 3 + 6** — `_safe_ao_ndarray` catches only `(AttributeError, NotImplementedError)` so genuine bugs propagate; channel degradation is recorded on `last_tracking` every call instead of a one-shot deduplicated warning. `_tracking_score` still raises when both channels are unavailable.
- **Finding 8** — invalid `follow` raises `ValueError` eagerly in `__init__`, before any SCF/ADC work.
- **Finding 9** — the `target="mp2"` string path now emits the same "ignoring run_adc keyword arguments" warning as the `method="mp2"` spelling.

**Cleanup** — removed write-only `previous_states`/`previous_excitation`, hoisted the `s_old`/`s_new` overlap integrals out of the channel loop, dropped the redundant local `LazyMp` import.

**New tests (findings 4, 5, 10)**
- Root-reorder test asserting the tracker follows a state whose positional index differs from `state_index`, plus the new diagnostic fields.
- End-to-end ADC(2) excited-state geomeTRIC optimisation smoke test (skips cleanly without geomeTRIC; ran and passed locally).
- Unit tests: level != 2 rejection (all paths), eager invalid-`follow`, out-of-range `state_index`, empty excitations, `_safe_ao_ndarray` failure + `weight==0` raise, `density_overlap_score` NaN and non-identity-overlap cases, and the kwargs-ignored warning for both target spellings.

Remaining nitpicks (identity-overlap default in the public `density_overlap_score` helper, `_PyscfModules` -> `SimpleNamespace`, the weak small-displacement test) were intentionally left as-is.

**Commit:** `0e8f059`

---
*Progress tracked by mach6*

PR #230, opened 2026-06-17
ADC Nuclear Gradients

maxscheurer:
Moving to fresh branch, will update the description shortly...


maxscheurer:
## Progress Update

Dropped the `nuclear_gradient_scanner` factory in `adcc/gradients/scanner.py` — it was a one-liner that just forwarded to `NuclearGradientScanner(*args, **kwargs)`.

Callers in the package, docs, tests, and both optimisation examples now use `adcc.NuclearGradientScanner(...)` directly:

- `adcc/__init__.py`, `adcc/gradients/__init__.py` — dropped the re-export and the `__all__` entry
- `docs/gradients.rst` — switched the `:py:func:` reference to `:py:class:` and updated the example snippet
- `adcc/tests/geomopt_scanner_test.py`, `adcc/tests/functionality_geomopt_test.py` — updated all scanner constructions
- `examples/optimization/pyscf_adcc_geoopt.py`, `examples/optimization/pyscf_adcc_geoopt_excited.py` — updated all scanner constructions and docstring cross-references
- Class docstring and the two internal `TypeError`/`ModuleNotFoundError` messages in `scanner.py` updated to say `NuclearGradientScanner`

Verified: `grep` finds no remaining references to `nuclear_gradient_scanner`; all 27 scanner unit tests and 3 end-to-end geomopt tests pass; both example files still compile.

**Commit:** `3dd1715`

---
*Progress tracked by mach6*

PR #234, opened 2026-07-27
Refactor Arithmetic Operations on AmplitudeVector

jonasleitner:
This PR refactors the arithmetic operations implemented on the `AmplitudeVector` class such that missing blocks are consistently treated as zero blocks.

Previously, this was only the case for `__add__`, while the other operations employed `__forward_to_blocks`, which raised an exception for differing blocks. Moreover, `dot` showed inconsistent behavior, `ADC1.dot(ADC2)` gave the `ph` overlap, while `ADC2.dot(ADC1)` raised an exception due to a missing `pphh` block. The in-place operators, also employed `__forward_to_blocks`, which created a new `AmplitudeVector` instance and thus were not really in-place.

In this PR I explicitly implemented the arithmetic operations ensuring that inputs and result don't share memory (`libadcc.Tensors` are copied if necessary). Moreover, in-place operations mutate the current instance modifying/deleting existing blocks.
Finally, type hints are introduced throughout `AmplitudeVector.py`

A Note on the reflective operations:
`libadcc.Tensor * Amplitude` (same for other operations) is not possible, since the backend raises a `TypeError`. Thus, `__rmul__` etc. are not reachable for this case and the implementation contains in principle dead code.

### DIFF
--- a/adcc/AmplitudeVector.py
+++ b/adcc/AmplitudeVector.py
@@ -168,6 +168,9 @@ class AmplitudeVector(dict[str, libadcc.Tensor]):
     ) -> "AmplitudeVector":
         if isinstance(other, AmplitudeVector):
             return other.__mul__(self)
+        # This path is not reachable currently, because the libadcc backend
+        # raises a TypeError instead of returning NotImplemented.
+        # Thus, the reflective operations are not called in this case.
         elif isinstance(other, libadcc.Tensor):
             ret = {block: other.__mul__(tensor) for block, tensor in self.items()}
         else:  # float
@@ -214,6 +217,9 @@ class AmplitudeVector(dict[str, libadcc.Tensor]):
     ) -> "AmplitudeVector":
         if isinstance(other, AmplitudeVector):
             return other.__add__(self)
+        # This path is not reachable currently, because the libadcc backend
+        # raises a TypeError instead of returning NotImplemented.
+        # Thus, the reflective operations are not called in this case.
         elif isinstance(other, libadcc.Tensor):
             ret = {block: other.__add__(tensor) for block, tensor in self.items()}
         else:  # float
@@ -258,6 +264,9 @@ class AmplitudeVector(dict[str, libadcc.Tensor]):
     ) -> "AmplitudeVector":
         if isinstance(other, AmplitudeVector):
             return other.__sub__(self)
+        # This path is not reachable currently, because the libadcc backend
+        # raises a TypeError instead of returning NotImplemented.
+        # Thus, the reflective operations are not called in this case.
         elif isinstance(other, libadcc.Tensor):
             ret = {block: other.__sub__(tensor) for block, tensor in self.items()}
         else:  # float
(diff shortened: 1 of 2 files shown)


adcc/AmplitudeVector.py
```diff
@@ -20,146 +20,313 @@
...
+        Ensures all blocks of the AmplitudeVector are evaluated and
+        resilient in memory.
+        """
         for t in self.values():
             t.evaluate()
         return self
 
     @property
-    def needs_evaluation(self):
+    def needs_evaluation(self) -> "bool":
```
frieschneider:
```suggestion
    def needs_evaluation(self) -> bool:
```

The return type doesn't need to be quoted. bool is a built-in type, so -> bool is sufficient.

adcc/AmplitudeVector.py
```diff
@@ -20,146 +20,313 @@
...
-    def dot(self, other):
-        """Return the dot product with another AmplitudeVector
+    @overload
+    def dot(self, other: "AmplitudeVector") -> float:
+        ...
+
+    @overload
+    def dot(
+        self, other: Sequence["AmplitudeVector"]
+    ) -> np.ndarray[tuple[int], np.dtype[np.float64]]:
```
frieschneider, line 105: We would currently mix plain `np.ndarray` and parameterized `np.ndarray[...]`. I  would prefer the parameterized form where possible. For example, `ElectronicTransition.py` uses the unparameterized form.

What do you think @jonasleitner ?

adcc/AmplitudeVector.py
```diff
@@ -20,146 +20,313 @@
...
+        # here all blocks of other have to be in self present, too!
+        # block missing in other: x / 0 -> division by zero
+        # block missing in self: 0 / x = 0 -> fine but not in result
         if isinstance(other, AmplitudeVector):
-            allblocks = sorted(set(self.blocks).union(other.blocks))
-            ret = {k: self.get(k, 0) + other.get(k, 0) for k in allblocks}
-            ret = {k: v for k, v in ret.items() if v != 0}
+            if any(block not in other for block in self.keys()):
+                raise ZeroDivisionError(
+                    "Divisian by zero, since missing blocks are treated as "
```
frieschneider:
```suggestion
                    "Division by zero, since missing blocks are treated as "
```
typo

adcc/AmplitudeVector.py
```diff
@@ -20,146 +20,313 @@
...
+        # missing block in other: x / 0 -> division by zero
+        # missing block in self: 0 / x = 0 -> fine but not in result
         if isinstance(other, AmplitudeVector):
-            return other.__add__(self)
-        else:
-            ret = {k: other + tensor for k, tensor in self.items()}
-            return AmplitudeVector(**ret)
+            if any(block not in other for block in self.keys()):
+                raise ZeroDivisionError(
+                    "Divisian by zero, since missing blocks are treated as "
```
frieschneider:
```suggestion
                    "Division by zero, since missing blocks are treated as "
```

typo

adcc/AmplitudeVector.py
jonasleitner: I also prefer the parametrized form including shape and optionally the data type. However, I would suggest to resolve this at a later stage once we revisit the corresponding files regarding type hints?

frieschneider:
About the reflective operations: In my opinion, we should add a comment explaining that expressions such as `libadcc.Tensor * AmplitudeVector` (and the analogous operations) are currently unreachable because the backend raises a `TypeError` first.
Otherwise, this will likely be forgotten in the future.

frieschneider (approved):

PR #235, opened 2026-08-07
Fix MP(0) Energy

jonasleitner:
Add nuclear repulsion energy to the HF interface. Consequently, all HF interfaces from now on need to provide this! Here, I updated the psi4 and pyscf interfaces.
The nuclear repulsion energy is needed for the calculation of the MP(0) energy, which was recently added, but the nuclear contributions were missing.

I also updated the HF test data to include the nuclear repulsion energy

TODO:
- [x] rename testdata version


adcc/backends/psi4.py
```diff
@@ -241,6 +241,9 @@ def get_restricted(self):
     def get_energy_scf(self):
         return self.wfn.energy()
 
+    def get_nuclear_repulsion_energy(self) -> float:
```
frieschneider: I’m not sure I’m a fan of adding type hints only to this one method of the class. I get it  that adding them across the whole class would be a lot of work. However, if we want to introduce type hints step by step, I feel like this PR would be a good point to do it in this file (or at least this class).

adcc/tests/backends/pyscf_test.py
```diff
@@ -45,6 +45,9 @@ def base_test(self, scfres):
         hfdata = adcc.backends.import_scf_results(scfres)
         assert hfdata.backend == "pyscf"
 
+        assert hfdata.energy_scf == scfres.e_tot
+        assert hfdata.nuclear_repulsion_energy == scfres.energy_nuc()
```
frieschneider, line 49:
```suggestion
        assert hfdata.nuclear_repulsion_energy == scfres.energy_nuc()
        assert hfdata.spin_multiplicity == int(scfres.mol.spin) + 1
```
Is also missing, right?

adcc/backends/psi4.py
frieschneider: same for the `pyscf.py` file

frieschneider (approved):

PR #236, opened 2026-08-13
Introduce type hints in the HF backend files

jonasleitner:
Improves static typing of the Python-facing API. Mostly annotations and stub
regeneration, plus a few interface changes needed to make the generated stub
useful.

### C++ / pybind11 interface
- Named all binding arguments so the stub exposes keyword arguments instead of `arg0`, `arg1`, ...
- New `libadcc_src/pyiface/ndarray.hh`: `NDArray<T, Ndim>`, a thin `py::array_t<T>`
  wrapper that carries the dimensionality into the stub
  (`numpy.ndarray[tuple[int, int], numpy.dtype[numpy.float64]]`) and rejects
  arrays of the wrong `ndim` at the boundary (data type casts unchanged).
- `py::array` / `py::tuple` / `py::list` parameters replaced by more specific types that generate
    more specific signatures for the stub file. The runtime behavior should not be affected.
- pybind11 minimal version updated to 3.0 for the support of variadic arguments.

### Python backends
- Type hints for `pyscf`, `psi4` and `EriBuilder`. Note that a breaking change in the signature of
   `EriBuilder.compute_mo_eri` is introduced.
- `veloxchem`: added `get_nuclear_repulsion_energy` so the backend remains functional.

Additionally, the temporary `libxc-c=7.0.0` pin for psi4 is removed again.


adcc/backends/EriBuilder.py
```diff
@@ -107,20 +140,30 @@ def split_1d_slice(self, sl):
                 toslice = (bslice.start - sl.start, sl.stop - sl.start)
             if fromslice is None or toslice is None:
                 continue   # Not found
-            ret.append(SpinBlockSlice(block[0].upper(), block[1],
-                                      slice(*fromslice), slice(*toslice)))
+            ret.append(SpinBlockSlice(
+                block[0], block[1], slice(*fromslice), slice(*toslice)
```
frieschneider:
```suggestion
                block[0], block[1], slice(*fromslice, 1), slice(*toslice, 1)
```
Otherwise, the type checker should report an error, since `slice(*fromslice)` and `slice(*toslice)` have step=None and therefore does not satisfy `IntSlice` (`slice[int, int, int]`).

adcc/backends/psi4.py
```diff
@@ -20,16 +20,32 @@
...
+DipoleLike = tuple[Array2D, Array2D, Array2D]
+# Once we drop python 3.10 we can write
+# QuadrupoleLike = tuple[*DipoleLike, *DipoleLike, *DipoleLike]
+QuadrupoleLike = tuple[
+    Array2D, Array2D, Array2D,
+    Array2D, Array2D, Array2D,
+    Array2D, Array2D, Array2D,
+]
+Coordinate = tuple[float, float, float]
+Environment = Literal["pe", "pcm"]
```
frieschneider, line 47: These type definitions are present in both files (`pyscf.py` and `psi4.py`). I think it makes sense to define them in a shared location. Only `EnvironmentImplementation` is different

adcc/backends/pyscf.py
```diff
@@ -201,29 +229,30 @@ def compute_mo_eri(self, blocks, spins):
...
         super().__init__()
-        self.scfres = scfres
+        self.scfres: scf.hf.SCF = scfres
         n_alpha, n_beta = scfres.mol.nelec
-        self.eri_builder = PyScfEriBuilder(self.scfres, self.n_orbs,
-                                           self.n_orbs_alpha, n_alpha,
-                                           n_beta, self.restricted)
-        self.operator_integral_provider = PyScfOperatorIntegralProvider(
-            self.scfres
+        self.eri_builder: PyScfEriBuilder = PyScfEriBuilder(
```
frieschneider, line 245: Could we omit the explicit type annotation here? `PyScfEriBuilder(...)` already gives the type checker enough information to know the type of `self.eri_builder`, so the annotation seems redundant here.

frieschneider: It occurs in several places, so perhaps a more general question: What style do we want?

adcc/backends/psi4.py
jonasleitner: Yes I agree. I don't know yet where to put those definitions. Maybe we want to have a `typing.py` file for the backends and/or globally in adcc. I wanted to postpone that decision to a later point when we find that we need to repeat such type aliases more often.

adcc/backends/pyscf.py
jonasleitner: I think we want to have the explicit type hints to prevent an assignment like `self.eri_builder = 5`

adcc/backends/psi4.py
```diff
@@ -20,16 +20,32 @@
...
 import numpy as np
 
-from libadcc import HartreeFockProvider
-
 import psi4
 
-from .EriBuilder import EriBuilder
+import libadcc
+
+from .EriBuilder import EriBuilder, Spin4D, Block4D
```
frieschneider:
```suggestion
from .EriBuilder import EriBuilder, Block4D,  Spin4D
```
To be consistent with `pyscf.py` :)

adcc/backends/EriBuilder.py
```diff
@@ -48,49 +73,57 @@ class EriBuilder:
...
+        return [SpinBlockSlice4D(
+            (sl1.block, sl2.block, sl3.block, sl4.block),
+            (sl1.spin, sl2.spin, sl3.spin, sl4.spin),
+            (sl1.fromslice, sl2.fromslice, sl3.fromslice, sl4.fromslice),
+            (sl1.toslice, sl2.toslice, sl3.toslice, sl4.toslice)
+        ) for sl1, sl2, sl3, sl4 in product(*splitted)]
 
-    def split_1d_slice(self, sl):
+    def split_1d_slice(
+        self, sl: "slice[int | None, int, int | None]"
```
frieschneider:
```suggestion
        self, sl: IntSlice
````

If I'm not mistaken, it's only called from `split_4d_slice` which already guarantees `IntSlice` inputs. So the `int | None` parameter type seems a bit broader than it needs to be, since `None` values should never actually occur here.
That would also make the `None`  to `int` normalization at the start of the function unneccesary.

jonasleitner: Yes you are absolutely right. I didn't want to change the runtime behavior and thus widened the type hint for this method. For our code this shouldn't be a problem. But potentially for the code of other people. But I agree: we already adjusted the signature for `compute_mo_eri`, so we can also adjust `split_1d_slice` 👍🏻

frieschneider (approved):

PR #238, opened 2026-08-17
Fix <S^2> for frozen_core References

jonasleitner:
<S^2> was wrong for any `ReferenceState` with `frozen_core`.
"Normal" and `frozen_virtual` references were not affected.
This was a consequence of missing blocks in `ReferenceState.density_2p` and `OperatorIntegrals.ssq_2p`.

**Changes:**
* Generalized the construction of `density_2p` and `ssq_2p`. Note: the operator should most likely be also valid for CVS, but could not test it without CVS 2p densities and thus kept the `NotImplementedError` in place.
* `ReferenceState` and `LazyMp` <S^2> tests were updated to cover all cases to catch this in the future.
* The test data of the open shell systems was updated with the new `ssq` data.


adcc/ReferenceState.py
```diff
@@ -224,16 +224,27 @@ def density_2p(self):
         """
         Return the two-particle Hartree-Fock density in the MO basis
         """
+        # NOTE: running over canonical blocks here, to avoid incorporating
+        # implicitly knowledge about the definition of a canonical block in
+        # the implementation
         density = TwoParticleDensity(self.mospaces,
                                      symmetry=OperatorSymmetry.HERMITIAN)
+        # gamma_pqrs = delta_pr delta_qs - delta_ps delta_qr for occupied p,q,r,s
```
frieschneider:
```suggestion
        # gamma_ijkl = delta_ik delta_jl - delta_il delta_jk for i,j,k,l
```
occupied indices are usually i, j, k, l

adcc/ReferenceState.py
```diff
@@ -224,16 +224,27 @@ def density_2p(self):
...
+            p, q, r, s = splitted
+            if p == r and q == s:
+                density[block].set_mask("ijij", 1)
+            if p == s and q == r:
+                density[block].set_mask("ijji", -1)
+            # only set elements to zero if one of the checks were true and
+            # __getitem__ initialized a tensor. There should be no need
+            # to initialize all canonical blocks here.
+            if p == q and r == s and block in density.blocks_nonzero:
+                density[block].set_mask("iijj", 0)
```
frieschneider:
```suggestion
            i, j, k, l = splitted
            if i == k and j == l:
                density[block].set_mask("ijij", 1)
            if i == l and j == k:
                density[block].set_mask("ijji", -1)
            # only set elements to zero if one of the checks were true and
            # __getitem__ initialized a tensor. There should be no need
            # to initialize all canonical blocks here.
            if i == j and k == l and block in density.blocks_nonzero:
                density[block].set_mask("iijj", 0)
```

frieschneider (approved):

